Issue #11, opened 2024-02-06
Patterns with Placeholders


jpgough-ms:
### Description of Problem:

Defining more detailed patterns is a unique opportunity in the CALM model.
The level of detail allows us to create re-usable patterns with different configuration depending on the business case.
The API Gateway pattern is a good example of this, where an API Consumer is routing via an API Gateway to an API Producer.
In each instantiation of the pattern unique values will be required for:

* The `host`, `port`, and `path` of the API hosted on the API Gateway
* The OAuth2 well known endpoint to validate a JWT token
* The audience of a given API
* The `host`, `port`, and `path` of the downstream service

![gateway-pattern](https://github.com/finos-labs/architecture-as-code/assets/152306432/de9e480f-7a1f-4709-8ab8-59d953d40f23)

#### Possible Stories

* As an architect, I would like to provide patterns with mandatory placeholders that need to be supplied by the implementer, so that I can ensure pattern conformance within an opinionated architecture.
* As a security reviewer, I can validate that constraints are followed within an approved pattern, so that I can approve new deployments quicker.
*As a developer, I can choose a pattern that best fits my solution and provide the values from a clear to understand and check list, so that I can deliver my functionality quicker.

### Potential Solutions:

It is my opinion that we need to consider patterns with placeholders.
There are several approaches that we could take:
1. Patterns could be modelled as JSON Schemas that extend the CALM model 
   * Pros
      * It is what needs to be supplied and has a mechanism of type checking/completion
   * Cons
      * Difficult for pattern author
      * Limitations within JSON schema for variables and inheritance (though I am far from an expert here)
1. Patterns are JSON/YAML documents that use the CALM schema.
Architects design their pattern and provide placeholder values.
This approach becomes a capability where CALM models can be incomplete.
   * Pros
      * Easy to work with for architects and developers
   * Cons
      * Tooling would need to check that all placeholder values are supplied

It would be good to hear from the community if they agree with this concept and whether there are additional approaches that we could consider.

rocketstack-matt:
So following our attempts earlier in the week, I managed to get this working, you can see a sample of this here - https://github.com/rocketstack-matt/architecture-as-code/blob/patterns/calm/pattern/api-platform.json

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "API Platform",
  "type": "object",
  "properties": {
    "client": {
      "allOf": [
        {
[...]

PR #15, opened 2024-02-21
Major refactor - migrating CALM to a specific JSON vocabulary

rocketstack-matt:
Adding a new JSON Schema Vocab and switching to being meta schema based to provide better autocompletion and validation capabilities.


jpgough-ms:
Will resolve #11

rocketstack-matt:
Good point, I wasn't thinking I've just been so deep in understanding how to write json vocabs and meta-schemas :D 

Will re-add.

rocketstack-matt:
@Budlee  see https://github.com/finos-labs/architecture-as-code/pull/22

Budlee:
@jpgough-ms  and @rocketstack-matt , again this goes back to my question in issue#10 (https://github.com/finos-labs/architecture-as-code/issues/10)

In the api-gateway-implementation you are using Interacts as the relationship type instead of connects. Yet in TraderX we use interacts for Actors to Nodes and not for Services connecting. This goes to the point I made in the issue where we  have equivalent types of relationship types that not not strongly defined.

I think these need to be defined or at least documented which each one is

jpgough-ms:
@jpgough-ms this is my error, whilst looking at the schema for the first time.

In addition this is something that we can look to add into the ruleset for #20

rocketstack-matt:
Have merged this

Issue #20, opened 2024-02-21
Spectral Rules for Advanced Checking on Models


jpgough-ms:
### Description of Problem:

JSON schema provides a significant amount of checking, but when we need referential constraints e.g. `uniqueId` defined in relationships it would be good to use an advanced linting tool. 

### Potential Solutions:

- Use Spectral to create a limited set of core rules that will allow us to identify any relationships that do not fall within constraints.
- Use Spectral to create rules that will check for where instantiated patterns have not replaced placeholder/default values

jpgough-ms:
Another feature we might want to add is to constrain the relationship type based on the nodes in the relationship. For example, and `actor` should be able to `interact` with a `system` or service`. A system should not `interact` with another system, but instead `connect`

willosborne:
This has been done for instantiations, but not patterns. Basic validations are run as part of the CI build, see issue #35 
The full feature will be implemented as part of the CLI, see issue #73

PR #28, opened 2024-02-22
Spectral validation PoC

willosborne:
Adds spectral validation rules to inspect implementation documents.

Rules:
- check that documents have nodes and relationships as top level elements (error, this is a very simple example rule)
- check that every relationship references a node that exists in the document (error)
- check that every node referenced by a composition relationship exists (error)
- check that every node is referenced by at least one relationship (warning)


willosborne:
Addresses issue #20

PR #44, opened 2024-03-07
Issue#29 New Specification for 2024-03

Budlee:
A new specification for strongly typed relationships

### DIFF
--- a/calm/samples/traderx/traderx.json
+++ b/calm/samples/traderx/traderx.json
@@ -16,7 +16,7 @@
     {
       "uniqueId": "web-client",
       "node-type": "webclient",
-      "name": "Web  Client",
+      "name": "Web Client",
       "description": "Browser based web interface for TraderX",
       "data-classification": "Confidential",
       "run-as": "user"
--- a/calm/samples/traderx/traderx.yaml
+++ b/calm/samples/traderx/traderx.yaml
@@ -10,7 +10,7 @@ nodes:
     description: Person who manages accounts and executes trades
   - uniqueId: web-client
     node-type: webclient
-    name: Web  Client
+    name: Web Client
     description: Browser based web interface for TraderX
     data-classification: Confidential
     run-as: user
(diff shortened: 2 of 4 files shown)


calm/draft/2024-03/meta/core.json
```diff
@@ -0,0 +1,229 @@
...
+    "required": [
+      "nodes",
+      "relationships"
+    ]
+  },
+  "defs": {
+    "node": {
+      "type": "object",
+      "properties": {
+        "uniqueId": {
```
aidanm3341, line 27: I noticed that we have a mix of camel case and kebab case in the spec. Most are kebab case but fields like `uniqueId` and `dataClassification` are camel. If there isn't a reason behind this, maybe we could take this opportunity to make them consistent.

rocketstack-matt: Let's raise that as a separate issue given the amount of changes we've already asked for, but agree with making this happen.

jpgough-ms: @aidanm3341 agree this should be a separate issue and would be good to do now we're into a next version of the schema.

calm/draft/2024-03/meta/core.json
```diff
@@ -0,0 +1,229 @@
...
+          "$ref": "#/defs/authentication"
+        }
+      },
+      "required": [
+        "uniqueId",
+        "relationship-type"
+      ],
+      "additionalProperties": true
+    },
+    "dataClassification": {
```
jpgough-ms, line 115: @aidanm3341 when you create the issue for `unqiue-id` it's worth including `data-classification too`

cc/ @rocketstack-matt

jpgough-ms (approved):

calm/draft/2024-03/meta/core.json
rocketstack-matt: Already have #45

aidanm3341 (approved):

Issue #35, opened 2024-03-01
Add a GitHub action to run Spectral validation on CALM specification changes against existing patterns


rocketstack-matt:
### Description of Problem:
We now have a Spectral validator thanks to @willosborne in #28 - ideally this would be setup to run when the CALM specification changes to confirm patterns and samples are still valid, or conversely if the patterns or samples change they should run to confirm they're still valid per the existing spec.

rocketstack-matt:
@willosborne are we happy to close this issue now?

willosborne:
@rocketstack-matt No, this is not working - PR was not ready when it was merged. I thought I had marked the PR as a draft, my bad.

The validation works, but it can't mark itself as having failed due to permissions issues and thus always succeeds.

willosborne:
Going to try a new PR that does this manually, without using the Stoplight action.

willosborne:
Currently we have spectral rules to check instantiations, nothing on the patterns themselves.
Proper JSON schema validation will be a much more valuable inspection to run.

jpgough-ms:
@willosborne is this working now?

rocketstack-matt:
We will change approach to integrate actions validation with the new CLI - @willosborne please link to a new issue detailing that implementation then close this issue.

willosborne:
Currently we have a basic version of this running. This work will be extended in issue #97 - so closing issue for now.

PR #54, opened 2024-03-12
Add simple validation against samples

willosborne:
Addresses issue #35 - simple rule on all samples. 
Currently doesn't do pattern files as the rules don't understand the json metaschema syntax yet.


willosborne:
Note - branch is passing even though validation fails because it doesn't seem to have permission. 
https://github.com/stoplightio/spectral-action/issues/624 Seems others have had this problem. 

Concern that this may just not work for PRs from forks: https://github.com/italia/developers-italia-api/issues/67

.github/workflows/spectral-validation.yml
```diff
@@ -0,0 +1,23 @@
+name: Run Spectral Validation on all samples
+
+on:
+  - pull_request
```
matthewgardner, line 4:
```suggestion
  - pull_request
   - push:
      branches:
        - main
```

matthewgardner: Should we also not run this again main when merged? (perhaps over kill but also nice to know the main branch is correct:) )

willosborne: Yeah I think so.
We should run this 
 - when a PR is raised
 - when any new commits are pushed to said PR
 - when a PR is merged

rocketstack-matt (approved):

PR #60, opened 2024-03-14
Add GitHub action for the translator service

rocketstack-matt:


.github/workflows/translator.yml
```diff
@@ -0,0 +1,27 @@
+name: translator build with maven
+
+on:
+  push:
+    branches: [ "main" ]
+    paths: [ "translator/**" ]
+  pull_request:
+    branches: [ "main" ]
```
matthewgardner, line 8: think you need to remove this line - it wont run on branches otherwise

.github/workflows/translator.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          cache: maven
+      - run: mvn -B install --file ./translator/pom.xml
```
Budlee, line 23: So we want to do the docker build as well or shall we leave that for when we look at publishing?

rocketstack-matt: I would suggest we add that as a separate build . . . I'll need to look into that one a bit more, but this should at least prevent code causing test failures / missing dependencies / etc.

PR #61, opened 2024-03-14
Fix spectral tests

willosborne:
Spectral PR was merged before it was ready - currently it is unable to mark itself as having failed due to permissions issues. This PR will fix that - note that this is experimental as I need to test on the GH actions.


jpgough-ms (dismissed):
This looks sensible, hard to know if will work until run with GitHub actions.

PR #62, opened 2024-03-14
Intentionally break sample to test spectral rule

willosborne:
This PR is to test whether the actions trigger correctly from PRs that **don't** come from forks.
This PR should fail.


jpgough-ms (approved):

jpgough-ms (changes requested):
PR to see build fail

Issue #74, opened 2024-03-22
Pattern api-gateway.json does not conform to JSON Schema draft 2020-12 


lbulanti-ms:
## Bug Report

The [api-gateway.json ](https://github.com/finos-labs/architecture-as-code/blob/main/calm/pattern/api-gateway.json) does not conform to the JSON Schema Draft 2020-12. As mentioned in the [release notes](https://json-schema.org/draft/2020-12/release-notes#changes-to-items-and-additionalitems) the `items` keyword changed to `prefixItems`. 

### Additional Context:
This issue has been discovered while trying out the AJV library to validate the pattern schema against the CALM meta-schema

rocketstack-matt:
@lbulanti-ms do you mean that the core schema is wrong, which then correspondingly means the pattern is wrong or just the pattern? I think the former.

It would be great if you could show the code section and also what you ran to perform the validation if that is something that would be reproducible so that the person fixing the bug could validate they had done it correctly.

lbulanti-ms:
While there are other known issues with the [core.json](https://github.com/finos-labs/architecture-as-code/blob/main/calm/draft/2024-03/meta/core.json) meta-schema(we are raising a separate issue), this particular issue refers to the section below which uses the `items` key instead of the `prefixItems` key.
```"properties": {
    "nodes": {
      "type": "array",
      "minItems": 3,
      "items": [
        {
          "$ref": "https://raw.githubusercontent.com/finos-labs/architecture-as-code/main/calm/draft/2024-03/meta/core.json#/defs/node",
          "properties": {
            "ingress-host": {
              "type": "string"
            },
            ...
```
To validate the schema I used the AJV library. I ran this command 
`ajv compile -s api-gateway.json -r core.json -r calm.json --spec=draft2020 --strict=false` (we are not sure if we want to use the strict mode, so I set that to false while testing)
The error I am getting is `schema api-gateway.json is invalid
error: schema is invalid: data/properties/nodes/items must be object,boolean`

I am planning to raise a PR to fix this issue.

rocketstack-matt:
Thanks @lbulanti-ms so the issue you're looking at is in the pattern itself, but I think we have the same issue in the [core schema](https://github.com/finos-labs/architecture-as-code/blob/0c2fd64b267da1ebe12fb5efcbb06e066abf554c/calm/draft/2024-03/meta/core.json#L8) itself.

``` json
"properties": {
    "nodes": {
      "type": "array",
      "items": {
[...]

PR #79, opened 2024-03-25
#74 - Updating the api-gateway.json pattern to use prefixItems

lbulanti-ms:
Updating the api-gateway.json pattern to use the prefixItems keyword instead of items as per JSON Schema 2020-12

### DIFF
--- a/calm/pattern/api-gateway.json
+++ b/calm/pattern/api-gateway.json
@@ -7,7 +7,7 @@
     "nodes": {
       "type": "array",
       "minItems": 3,
-      "items": [
+      "prefixItems": [
         {
           "$ref": "https://raw.githubusercontent.com/finos-labs/architecture-as-code/main/calm/draft/2024-03/meta/core.json#/defs/node",
           "properties": {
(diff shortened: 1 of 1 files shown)


rocketstack-matt (changes requested):
Looks like you're being hit by a couple of issues not of your making . . 
* I have raised https://github.com/lbulanti-ms/architecture-as-code/pull/1 to resolve the failure of spectral rules being run against a README.md file (@willosborne FYI)
* Can I ask you to fill in the holes in the api-gateway-implementation.json to pass the new rules that are blocking this merge.

lbulanti-ms:
@rocketstack-matt I made the changes requested. There are other issues with the spectral rules so I will try to fix them. I think that @willosborne was looking into fixing these rules as well

rocketstack-matt:
> @rocketstack-matt I made the changes requested. There are other issues with the spectral rules so I will try to fix them. I think that @willosborne was looking into fixing these rules as well

Looking a lot cleaner, looks like we're just missing a couple of unique-id's now.

rocketstack-matt (approved):
This change looks good now, I had to remove a parties element that had been left in the sample (and shows why we needed the spectral validator!)

Issue #53, opened 2024-03-12
Pattern CLI for Bootstrapping Patterns


jpgough-ms:
### Description of Problem:

Patterns implemented as JSON Schema help to constrain and define reusable architectures. Developers can use tooling to "autocomplete" patterns, as demonstrated at #52 and at our monthly call.

We would like to propose the creation of a simple command line interface CLI that:

- Fetches a pattern
- Validates that it conforms to the CALM schema 
- Runs Linting against the pattern
- Generates the target pattern instantiation in JSON or YAML
- Makes it obvious to the developer the fields that they are required to contribute to

willosborne:
For the final bullet point - it would be great if the linter could pick up fields that weren't filled in and give the user a list of things to do on the CLI. Also, since Spectral has IDE plugins we could even get this information directly into their editor.

We could do this by just filling the blanks with some placeholder like MISSING_VALUE and then just having a spectral rule that looks for any property set to MISSING_VALUE.

willosborne:
One consideration here is which language we should use for the bootstrap CLI. In my mind we have two main options, Python and Node (i.e. JS or TypeScript.) I've had a think about some pros/cons - would be good to hear people's thoughts

## Python
Pros
+ Great argument parser (click) that also has strong support for subcommands
+ More native libraries for JSON Schema validation than Node
+ Can publish CLI tool to pypi for easy consumption

Cons
- No native interop with visualisation tool as that is based on JS - would need to call out via subprocess
- No native interop with Spectral, would need to call out via subprocess

## Node / JavaScript
Pros
+ Similar arg parser library based on Python's click (Commander) 
+ Native interop with visualization tool @aidanm3341 is working on
+ Native interop with Spectral - can use it from within a Node app 
+ Can publish CLI tool to NPM for easy consumption
+ Option for strong typing via TypeScript

Cons
- Strong typing may be a hindrance in early stages as CALM schema is volatile - more refactoring needed
- Main option for JSON Schema validation is ajv - if that doesn't work we will need to call out to another CLI tool via subprocess

## Other languages
Java: 
- CLI tools are weaker than Node/Python
- No interop with tools in use, 
- Heavier CLI size due to fat jar
- CLI install may be more difficult

Go: 
- Less widely used language, so experience may be limited

aidanm3341:
[...]

PR #80, opened 2024-03-26
created core cli (#53)

aidanm3341:
In order to get an initial version of the CLI merged so that people can work on the various features we have planned in parallel, I have created a simple shell of the CLI with a placeholder command


rocketstack-matt (approved):
Works per README locally.

PR #81, opened 2024-03-26
Fix spectral rules, tighten workflow and provide template parameters on api-gateway implementation

willosborne:
- Fixes rules to fit with new calm changes.
- Replaces empty or zero properties in api-gateway with {{ PROPERTY_NAME }} and -1
- Adds rules to report this at warn level
- Make workflow only run on .json and .yaml files instead of all files (incl. README.md!)


willosborne:
Rules are doing exactly what I expected - you can see warnings about placeholder properties in the spectral details.

rocketstack-matt (changes requested):
@willosborne can you update now @lbulanti-ms's PR was merged; I think you likely no longer need the spectral-validation.yml and api-gateway-implementation.json changes in your PR as they had to be resolved to fix the build for the other one.

willosborne:
@rocketstack-matt merged in those changes, thanks for the heads up

rocketstack-matt (approved):

PR #82, opened 2024-03-26
add k8s service and namespace

matthewgardner:
Added ability to generate following k8s manifests as single output
- namespaces
- service (nb: assumed port 8080!)


rocketstack-matt:
@matthewgardner is there an issue we can link this PR to?

matthewgardner:
> @matthewgardner is there an issue we can link this PR to?

https://github.com/finos-labs/architecture-as-code/issues/32 

Sorry how should I have linked these?

rocketstack-matt (dismissed):

rocketstack-matt:
@Budlee are you able to review this PR, only @matthewgardner and yourself are code owners in the translator service

translator/src/test/java/org/finos/calmtranslator/translators/TestK8sModelTranslatorShould.java
```diff
@@ -1,26 +1,37 @@
...
-		this.translator = new K8sModelTranslator();
-		this.traderxCalmModel = new ObjectMapper().readValue(this.getClass().getClassLoader().getResourceAsStream("traderx-calm.json"), Core.class);
-	}
+    @Test
+    void generateServices() {
+        String k8sManifest =  translator.translate(traderxCalmModel);
+        
+        //Simple namespace generation check
+        assertThat(k8sManifest).contains("Namespace");
+        assertThat(k8sManifest).contains("traderx-system");
```
Budlee, line 28: For the test can we compare the against a whole yaml? 

I think this test is a bit weak at present

translator/src/main/java/org/finos/calmtranslator/translators/K8sModelTranslator.java
```diff
@@ -1,21 +1,89 @@
...
+        response.append(
+            namespaces.stream().map(ns -> createNamespace(ns)).collect(Collectors.joining("")));
+
+        //Build Services 
+        response.append(
+        calmModel.getNodes().stream()
+            .filter(node -> node.getNodeType() == Node.NodeTypeDefinition.SERVICE)
+            .map(service -> createService(
+                    service,serviceToNamespaceMap)).collect(Collectors.joining("")));
+        return response.toString();
```
Budlee, line 53: I don't think there is a separation here?
The code looks like it just appends the yaml together, there is no line separation between the components.

translator/src/main/java/org/finos/calmtranslator/translators/K8sModelTranslator.java
```diff
@@ -1,21 +1,89 @@
...
-				.endMetadata()
-				.build();
-		// Get YAML string
-		System.out.println(Serialization.asYaml(p));
-		// Get JSON string
-		System.out.println(Serialization.asJson(p));
-		return null;
-	}
+    @Override
+    public String translate(final Core calmModel) {
```
Budlee, line 21: I would return a list of K8s Components over a string. 
Then let the Spring serialiser in the controller deal with converting it to Yaml or JSON to return to the user.

Additionally this will make your tests significantly easier

Budlee (changes requested):
There are some changes I think are necessary, I really think the return type should be an object and the controller will handle the serialization

translator/src/test/java/org/finos/calmtranslator/translators/TestK8sModelTranslatorShould.java
matthewgardner: I did think about that. The suggestion means our tests need to be aware of the k8s api. When that changes (most likely via the fabric8 lib) we will need to adjust the k8s sample/test manifest.

This led me to think about testing using a much smaller calm model - which then means maintaining additional calm models in our tests.

What do you think? Both sound like a maintainance nightmare

translator/src/main/java/org/finos/calmtranslator/translators/K8sModelTranslator.java
matthewgardner: I just used a simple Collector implementation to join the output. The Fabric8 lib provides the separator as part of its interface. Sample below. We could add another one but that wouldn't be help imho.

```
---
apiVersion: "v1"
kind: "Namespace"
metadata:
  name: "traderx-system"
---
apiVersion: "v1"
kind: "Service"
metadata:
  name: "web-gui"
  namespace: "traderx-system"
spec:
  ports:
  - port: 8080
  selector:
    app: "web-gui"
```

translator/src/main/java/org/finos/calmtranslator/translators/K8sModelTranslator.java
matthewgardner: We can defo start with a list of strings - this also ensure order. I'll work on those changes.

Future state we might need another plan as i can see use providing two options

- flat output as yaml/json
- zip/tar containing a structured folder with files for each manifest

 But that can be a problem for future Matt ;)

matthewgardner: actually returning  `List<KubernetesResource>` works well

translator/src/main/java/org/finos/calmtranslator/translators/K8sModelTranslator.java
matthewgardner: Ok this is getting removed in coming changes

translator/src/test/java/org/finos/calmtranslator/translators/TestK8sModelTranslatorShould.java
matthewgardner: aha - using another object will make this nicer - updating coming soon

matthewgardner:
@Budlee one change i didnt make is to the controller - are we going to provide a separate endpoint or converge them?

translator/src/test/java/org/finos/calmtranslator/translators/TestK8sModelTranslatorShould.java
matthewgardner: @Budlee this looked a lot nice now - love to get your feedback :)

Budlee (approved):
Great start to the project. Lets get this in

Issue #42, opened 2024-03-05
Create a version of the Visualizer based on Mermaid


rocketstack-matt:
### Discussed in https://github.com/finos-labs/architecture-as-code/discussions/41

<div type='discussions-op-text'>

<sup>Originally posted by **rocketstack-matt** March  5, 2024</sup>
I've been experimenting trying to use Mermaid to visualise the CALM examples. My rationale for this is:
* Mermaid is already supported for visualisation in most common IDEs including IntelliJ and VS Code.
* There is a clear map between Mermaid flowchart syntax and the syntax used for CALM.

I've been able to produce the following diagram using Mermaid for the TraderX sample.

```mermaid
flowchart TB
    subgraph internal_bank_network [Internal Network: Bank ABC Internal Network]
        subgraph traderx_system [System: TraderX]
            web_client[Webclient: Web Client]
            web_gui_process[Service: Web GUI]
            position_service[Service: Position Service]
            traderx_db[(Database: TraderX DB)]
            reference_data_service[Service: Reference Data Service]
            trading_services[Service: Trading Services]
            trade_feed[Service: Trade Feed]
            trade_processor[Service: Trade Processor]
            accounts_service[Service: Accounts Service]
            people_service[Service: People Service]
        end
    end

    traderx_trader((Actor: Trader))
    user_directory[LDAP: User Directory]
    
    traderx_trader -->|interacts| web_client
    web_client -->|connects HTTPS OAuth2| web_gui_process
    web_gui_process -->|connects HTTPS SPNEGO| position_service
    position_service -->|connects JDBC Kerberos| traderx_db
    web_gui_process -->|connects HTTPS SPNEGO| reference_data_service
    web_gui_process -->|connects HTTPS SPNEGO| trading_services
    web_gui_process -->|connects WebSocket Kerberos| trade_feed
    trade_feed -->|connects AMQP Kerberos| trade_processor
    trade_processor -->|connects JDBC Kerberos| traderx_db
    web_gui_process -->|connects HTTPS SPNEGO| accounts_service
    web_gui_process -->|connects HTTPS SPNEGO| people_service
    people_service -->|connects LDAP Kerberos| user_directory

    %% Custom styles to indicate 'composed-of' relationship visually
    style traderx_system stroke-dasharray: 5 5

```

[...]

PR #83, opened 2024-03-26
#42 removed old visualizer to make room for the new one

aidanm3341:
In order to not have one massive PR, I am first removing all the old visualiser code and I will then create a followup PR with the new visualiser, as well as integrating it with the CLI


rocketstack-matt (approved):
LGTM

PR #85, opened 2024-03-26
#42 included initial visualization command in CLI

aidanm3341:
This PR adds the code to create a visualisation of a CALM spec in the CLI. The way it is used looks like 

```
calm visualize -i <input-file> -o <output-file>
```

and it outputs an SVG in the specified output file. This SVG can then easily be opened in a browser to be inspected.

Right now the library is just installed from the local repo. Next steps will be to publish both this and the CLI to npm so that we can manage the dependencies correctly.

### DIFF
--- a/cli/commands/visualize/visualize.ts
+++ b/cli/commands/visualize/visualize.ts
@@ -1,0 +1,8 @@
+import { visualize } from 'calm-visualizer';
+import * as fs from 'fs';
+
+export default async function({input, output}: {input: string, output: string}) {
+    const calm = fs.readFileSync(input, 'utf-8');
+    const svg = await visualize(calm); 
+    fs.writeFileSync(output, svg);
+}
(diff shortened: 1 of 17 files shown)


rocketstack-matt:
I was able to run this successfully locally against the sample json files, but not the yaml - assume this is expected.

Also, what was the reasoning about jumping to SVG? Do we want to give folks the option of actually getting the mermaid output and do we plan to still have an IDE plugin?

aidanm3341:
Yes that makes sense, making it work with yaml would be a fairly easy enhancement. 

The tool always used SVG behind the scenes, it's just that with the extension I had previously shown, it immediately rendered it in the window. Specifying different output formats is something that I definitely have in mind to expand this - mermaid would be one format, but also exporting as an image would be useful I think. 

I definitely still plan to create an IDE plugin using the visualiser library. There were some issues that I ran into during the first attempt, and I thought that the CLI would be the easiest first step. Having done that, the refactoring that I've done to get this working should actually solve some of the problems I was having, and I think making the extension will be pretty easy now.

rocketstack-matt (approved):

PR #91, opened 2024-03-28
Pattern flag and fixing api-gateway pattern sample

willosborne:


willosborne:
Wasn't sure about the visualizer command because I realised I don't know if it runs on a pattern file or an instantiation. It just says 'CALM specification' which is rather ambiguous. Didn't want to break it until I was sure.

grahampacker-ms (approved):

jpgough-ms (approved):

PR #92, opened 2024-04-02
Add support for additional top level nodes

willosborne:
In the event that additional properties are desirable on a pattern, this should be included in the instantiation. This just treats extra policies in the same way as properties defined in the prefixItems section of nodes or relationships.

Other changes:
- warning when items used instead of prefixItems
- substantial improvements to tests
- migration to use winston logger more consistently and remove use of console.log


rocketstack-matt:
It would be good to discuss this, I'm still unsure if our current samples where we're defining additional properties that aren't part of the core schema are valid or if we're missing key properties and we're going to end up with duplication and lack of standardisation.

willosborne:
Locked for now

Issue #48, opened 2024-03-11
Composable Patterns using a Pattern Interface


jpgough-ms:
### Description of Problem:

In #11 we modified CALM to be a [JSON Meta-Schema](https://json-schema.org/specification#meta-schemas), which enabled the creation of patterns and the instantiation of patterns. 

In #11 we presented the idea of an API Gateway pattern, which had the nodes: `API Consumer`, `API Producer`, `API Gateway`, and `IDP`. In this model, the `API Producer` was modelled as a `Node`. However, in reality the `API Producer` could be a series of connected microservices, connect to a database etc. 

Based on the [API Gateway Pattern](calm/pattern/api-gateway.json) the `API Producer is connected to using a `host` and `port`. If we could specify a node to have an optional external reference to a CALM compliant JSON, we could model patterns that have a variety of possible node level implementations. 

### Potential Solutions:

Create the concept of an interface which defines the hand-off/boundary between two patterns, allowing patterns to be composable or contained within other patterns.

rocketstack-matt:
Ok, sounds like you're seeing the same issues as @Budlee reported in #47

rocketstack-matt:
Link to #17

rocketstack-matt:
Some thoughts / comments for discussion later:
* What we see in the API Gateway pattern shows a system level diagram of interactions, would be similar to System Context Diagram in C4.
* Each of these systems should theoretically have a full CALM model describing them.
* Currently we model the purpose of a node and the type of relationship that connects them, but we don't model the actual interface a node exposes, other than where we have added them as additional properties in the [api-gateway](https://github.com/finos-labs/architecture-as-code/blob/main/calm/pattern/api-gateway.json) pattern.
* Interfaces obviously need to exist for any software component to speak to another, this could be a host/port, a file share, a message queue, etc.
* Where do we think the interfaces should be captured on nodes / relationships / both? My initial thinking leans towards both, nodes should enumerate the interfaces they support and relationships should specify which interface they are connecting; this could be done by specifying the details in two places or we could add an explicit ID to an interface and reference the interface in the relationship?
* The validator should ensure that only declared interfaces are defined in relationships (error if they don't). It could also warn if a node exposes an interface which is not then used (should we add an ignore mechanism to document that the architect knows they are exposing an unreferenced port)?
* Assuming we agree interfaces exist, and we should be able to mark specific interfaces as external then those interfaces should also be referenced in the system level architectures with an explicit reference into the internal architecture.

[...]

PR #106, opened 2024-04-03
#48 Remove mandatory interface on connects relationship

jpgough-ms:
With https://github.com/finos-labs/architecture-as-code/pull/106 we have decided to make interfaces optional on a relationship for now. If it was mandatory that would remove the ability to use CALM as a logical tool and mean that it is only a physical modelling tool.

We may need to consider whether we have a convention to mark what type of diagram this is

### DIFF
--- a/calm/draft/2024-04/meta/interface.json
+++ b/calm/draft/2024-04/meta/interface.json
@@ -21,7 +21,7 @@
       "interface": {
         "type": "string"
       },
-      "required": ["node", "interface"]
+      "required": ["node"]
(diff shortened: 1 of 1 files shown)


rocketstack-matt (approved):
We will discuss further how we want to support use of CALM as a modelling tool vs. instantiation / control mechanism further, but for now approving this change to provide maximum flexibility.

If we were to build a security control capability over CALM it would likely need to differentiate between logical vs. physical architecture.

PR #109, opened 2024-04-03
Update API Gateway Pattern to Add Interface

jpgough-ms:


rocketstack-matt (approved):
We'll need to raise an issue to update cli calm generate (and visualise I assume) to understand interfaces.

Issue #110, opened 2024-04-03
Update CLI to understand interfaces


rocketstack-matt:
Per . . 

_Originally posted by @rocketstack-matt in https://github.com/finos-labs/architecture-as-code/pull/109#pullrequestreview-1977156635_
           
Changes have been made to CALM to introduce interfaces, we need to reflect these in the CLI tools.

jpgough-ms:
I don't think #112 fixed this issue, will leave this open for now and assign to @willosborne

willosborne:
As you say @jpgough-ms - 112 solved the problem for a minimal subset of cases but I don't think this is complete. You can work around it as is, but it's painful (needs a lot of duplication) and definitely not ergonomic. I wouldn't call this complete until we can follow refs properly; will be looking to implement this early next week.

PR #112, opened 2024-04-03
Interface instantiation of explicit properties

willosborne:
Instantiates interfaces when their properties are explicitly listed.
Does not traverse $refs or perform any lookups. 

#110

### DIFF
--- a/cli/src/commands/generate/generate.ts
+++ b/cli/src/commands/generate/generate.ts
@@ -46,6 +46,30 @@ function getPropertyValue(keyName: string, detail: any) : any {
     }
 }
 
+function instantiateNodeInterfaces(detail: any): any[] {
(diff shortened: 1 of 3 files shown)


cli/src/commands/generate/generate.ts
```diff
@@ -46,6 +46,30 @@ function getPropertyValue(keyName: string, detail: any) : any {
     }
 }
 
+function instantiateNodeInterfaces(detail: any): any[] {
+    const interfaces = [];
+    if (!('prefixItems' in detail)) {
```
aidanm3341, line 51: we should definitely revisit this once we've figured out what we're doing with our typed representation of CALM. `any` is a virus which will bleed into every corner the codebase if we're not careful, then we may as well use JS

PR #123, opened 2024-04-04
Issue 73 - adding validate command to cli

lbulanti-ms:
Adding validate function that can validate a JSON schema against an instantiation. The meta schema can be provided as an argument.

### DIFF
--- a/cli/src/commands/validate/validate.spec.ts
+++ b/cli/src/commands/validate/validate.spec.ts
@@ -1,0 +1,142 @@
+import fetchMock from 'fetch-mock';
+import validate from './validate';
+import { readFileSync } from 'fs';
+import path from 'path';
+
+jest.mock('@stoplight/spectral-core', () => {
+    const spectralCore = jest.requireActual('@stoplight/spectral-core');
+    return {
+        ...spectralCore,
+        Spectral: jest.fn().mockImplementation(() => {
+            return {
+                run: () => [],
+                setRuleset: () => {},
+            };
+        })
+    };
+});
+
+describe('validate input files', () => {
+
+    it('The JSON Schema pattern cannot be found in set path', async () => {
+        const mockExit = jest.spyOn(process, 'exit')
+            .mockImplementation((code) => { throw new Error(`${code}`); });
+
+        await expect(validate('thisFolderDoesNotExist/api-gateway.json', '../test_fixtures/api-gateway-implementation.json'))
+            .rejects
+            .toThrow();
+
+        expect(mockExit).toHaveBeenCalledWith(1);
+        mockExit.mockRestore();
+    });
+
+    it('The pattern instantiation file cannot be found in set path', async () => {
+        const mockExit = jest.spyOn(process, 'exit')
+            .mockImplementation((code) => { throw new Error(`${code}`); });
+
+        await expect(validate('test_fixtures/api-gateway.json', '../doesNotExists/api-gateway-implementation.json')) 
+            .rejects
+            .toThrow();
+
+        expect(mockExit).toHaveBeenCalledWith(1);
+        mockExit.mockRestore();
+    });
+
+    it('The pattern instantiation file does not contain JSON', async () => {
+        const mockExit = jest.spyOn(process, 'exit')
+            .mockImplementation((code) => { throw new Error(`${code}`); });
+
+        await expect(validate('test_fixtures/markdown.md', 'test_fixtures/api-gateway-implementation.json'))
+            .rejects
+            .toThrow();
+
+        expect(mockExit).toHaveBeenCalledWith(1);
+        mockExit.mockRestore();
+    });
+
+    it('The JSON Schema pattern URL returns a 404', async () => {
+        const mockExit = jest.spyOn(process, 'exit')
+            .mockImplementation((code) => { throw new Error(`${code}`); });
+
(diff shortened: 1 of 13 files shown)


cli/src/commands/validate/validate.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+
+        await expect(validate('https://exist/api-gateway-implementation.json', 'http://exist/api-gateway.json', metaSchemaLocation))
+            .rejects
+            .toThrow();
+
+        expect(mockExit).toHaveBeenCalledWith(1);
+        fetchMock.restore();
+    });
+
+    it('The meta schema location is not a directory', async () => {
```
aidanm3341, line 147: best practice for test names with jest is to write it as if it would come after the word in the `describe()` function so that the output reads like a sentence. E.g. we could name the test `exits with error when meta schema location is not a directory`, so that the test output reads as `validate exits with error... etc.`

cli/src/commands/validate/validate.ts
```diff
@@ -0,0 +1,95 @@
...
+    let exitCode = 0;
+    try {
+        const ajv = new Ajv2020({ strict: false });
+
+        loadMetaSchemas(ajv, metaSchemaPath);
+
+        console.info(`Loading pattern from : ${jsonSchemaLocation}`);
+        const jsonSchema = await getFileFromUrlOrPath(jsonSchemaLocation);
+
+        console.info(`Loading pattern instantiation from : ${jsonSchemaInstantiationLocation}`);
```
aidanm3341, line 17: We should use Winston for logging here. It's already been added as a dependency and there are examples of how to use it in the generate command that Will wrote

cli/src/commands/validate/validate.ts
```diff
@@ -0,0 +1,95 @@
...
+            ajv.addMetaSchema(meta);
+        }
+    });
+}
+
+async function runSpectralValidations(jsonSchemaInstantiation: string) {
+    const spectral = new Spectral();
+    spectral.setRuleset(await getRuleset('../spectral/calm-validation-rules.yaml'));
+    const issues = await spectral.run(jsonSchemaInstantiation);
+    if (issues !== undefined && issues.length !== 0) {
```
aidanm3341, line 63: a more idiomatic way to write this would be
```suggestion
    if (issues && issues.length > 0) {
```

cli/src/commands/validate/validate.ts
```diff
@@ -0,0 +1,95 @@
...
+        console.info('Spectral issues: ', issues);
+        if (issues.filter(issue => issue.severity === 0).length !== 0) {
```
aidanm3341, line 65:
```suggestion
        if (issues.filter(issue => issue.severity === 0).length > 0) {
```

cli/src/commands/validate/validate.ts
```diff
@@ -0,0 +1,95 @@
...
+            //Exit with 1 if any of the Spectral issues is an error
+            process.exit(1);
+        }
+    }
+}
+
+async function getFileFromUrlOrPath(input: string) {
+    const urlPattern = new RegExp('^https?://');
```
aidanm3341, line 73: you can write this as a regex literal! 

```suggestion
    const urlPattern = /^https?:\/\//;
```

cli/src/index.ts
```diff
@@ -25,4 +26,11 @@ program
         runGenerate(options.pattern, options.output, !!options.verbose);
     });
 
+program
+    .command('validate')
+    .requiredOption('-p, --pattern <pattern>', 'The location of the JSON Schema Pattern')
```
aidanm3341, line 31: I think we should use a consistent description of this field and the instantiation field - since it is going to be used in all 3 of the existing commands (I've yet to add it to visualization, but I will)

cli/src/index.ts
```diff
@@ -25,4 +26,11 @@ program
...
+    .requiredOption('-i, --instantiation <instantiation>', 'The location of the instantiation of the Pattern')
+    .option('-m, --metaSchemasLocation <metaSchemaLocation>', 'The location of the directory of the meta schemas to be loaded', '../calm/draft/2024-03/meta')
```
aidanm3341, line 33: not sure about this as a default value. It will work while we are working on this locally, but in the bundled version that folder won't exist. Granted, we haven't done the work to bundle it yet. We may just have to remember to change this later on.

cli/src/commands/validate/validate.ts
willosborne: could you not just do since [] is falsy?

```
if (!!issues) { 

}
```

cli/src/commands/validate/validate.ts
lbulanti-ms: I am planning to add Winston in a following PR. I think it would be good to extract the logic to instantiate the logger in an external file because we reuse it across the 3 commands

cli/src/index.ts
willosborne: There are a few things on this PR that we want to change down the line, but we just want to get this merged so people have something to validate with in the meantime.

cli/src/commands/validate/validate.ts
aidanm3341: technically yes that makes sense, but I personally would have to take a minute to understand that this is checking for a non-empty list

aidanm3341 (approved):

willosborne (approved):

PR #124, opened 2024-04-04
Interface spectral fix

willosborne:
Fix spectral rules for instantiations + add new rule to check interfaces exist.
Also move files to prepare for pattern flavour of spectral rules and update spectral workflow


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "yaml.schemas": {
+        "https://json.schemastore.org/yamllint.json": "file:///home/will/code/architecture-as-code/spectral/instantiation/validation-rules.yaml"
```
jpgough-ms, line 3: It looks like this one probably should not get checked in? Do we have a .gitignore for this?

jpgough-ms (changes requested):
Looks like traders.json needs a few upgrades

calm/samples/api-gateway-implementation.json
rocketstack-matt, line 1: Shouldn't the placeholders be replaced with actual values in the in the implementation and the pattern file should be updated to include the host and port placeholders in the interfaces?

willosborne: This file is meant to look like what would be output by a generate command, as I understand it. i.e. it's ready to be filled out by a user by replacing the placeholders.

The pattern file should have properties defined without a value - e.g 
host: {
  type: "string"
}

Then the placeholders are generated from this.

Sorry if I've misunderstood something!

rocketstack-matt: I don't think that's right. This folder should contain real samples that we're providing almost as documentation / guides for users and therefore implementations should be 'working' so to speak (I appreciate this is already wrong, not being changed to this in your PR).

I think if we want an example of the 'interim state' implementation that would be generated by the CLI then we should put that in the CLI project either as an example or as a test file.

willosborne: This makes sense, happy to apply this update later. But I think this should be on a separate PR.

rocketstack-matt: I'm fine with that.
- [ ] Correct the sample implementations to now have any placeholders

rocketstack-matt (approved):

lbulanti-ms (approved):

Issue #111, opened 2024-04-03
Composable Architecture References


jpgough-ms:
## Feature Request

Some patterns do not supply all the details for a given node within an architecture and indeed the node may be implemented in different ways by different pattern instantiations.  

### Description of Problem:

It would be good to provide a mechanism for a pattern instantiations to supply the implementation detail in additional CALM files. This could be something like:

```json
{
  "interfaces": [
    {
      "unique-id": "producer-ingress",
      "host": "{{ INGRESS_HOST }}",
      "port": -1
    }
  ],
  "node-type": "system",
  "name": "General API Producer",
  "description": "The API Producer serving content",
  "unique-id": "api-producer",
  "detailed-architecture": {
       "$ref": "api-producer.json"
   }
},
```

Having this feature (with interfaces) will enable us to create composable patterns within CALM.

@rocketstack-matt

rocketstack-matt:
I really like this idea, it gives us the ability to drill in and out of architectures and provides the mechanism we wanted to enable us to represent what would be different levels in C4 without polluting a single model.

jpgough-ms:
After a discussion with @rocketstack-matt I have created a proposed schema change in #132. This is a nice way of defining the additional node, however it comes with one architectural sticking point that the pattern and `detailed-architecture` node would need to use the same schema version. This could become problematic.

We will need to review how we might approach this with a view to make the `detailed-architecture` support a different version of the CALM schema

rocketstack-matt:
What is it we actually want to achieve with this reference?

I think what we're trying to achieve is to enable different systems (or indeed subsystems within a complex system) to be linked so we can split what would be an unmanageable documentation process into something that provides 'right sized' views and enables the consumer (human or tool) to be able to navigate from one to the other.

If this is the case do we need to do anything other than specify the URI of an implementation? Ideally we would be able to provide a reusable repository mechanism for organisations to store their architecture implementations (and patterns) but assuming the referenced URI is findable then the human or tool could navigate and continue their journey.

Could we therefore simply make this a string?

jpgough-ms:
In #132 we created a PR with a self referential schema validation, this could potentially cause multiple issues:

* It could pin the version of the calm together with the further detailed-architecture. This had the potential to be awkward to maintain pattern instantiations and the linked more detailed architectures 
* Referenced architectures might not be available in all situations 

[...]

PR #132, opened 2024-04-08
#111 create a detailed-architecture node, which is itself defined as …

jpgough-ms:
…the CALM document

### DIFF
--- a/calm/draft/2024-04/meta/core.json
+++ b/calm/draft/2024-04/meta/core.json
@@ -46,6 +46,9 @@
           "items": {
             "$ref": "interface.json#/defs/interface-type"
           }
+        },
+        "detailed-architecture": {
+          "$ref": "#"
         }
       },
       "required": [


jpgough-ms:
@rocketstack-matt  I think this is the right way to do this, but interested to get your feedback.

calm/draft/2024-04/meta/core.json
```diff
@@ -46,6 +46,9 @@
           "items": {
             "$ref": "interface.json#/defs/interface-type"
           }
+        },
+        "detailed-architecture": {
+          "$ref": "#"
```
rocketstack-matt, line 51: Do we want to use a $ref in this case? What we trying to achieve is to reduce duplication by being able to reference pre-existing CALM models. 

Whilst I can see it would be nice to $ref those documents, I'm not sure it's actually needed, we simply need a place to reference the URI of the models.

This actually ties in somewhat with having a tool that makes it easier to edit the schemas locally by being able to load all schemas from a public or private model repository (so that the URI itself doesn't need to be physically addressable).

jpgough-ms: @rocketstack-matt my understanding of a `ref` at this level means that the `detailed-architecture` must be defined in terms of CALM - i.e. a self referential to this meta-schema.

In a pattern instantiation they could:

1) Reference a file - using a second $ref to point to their file
2) Inline define an architecture directly within this node

jpgough-ms: @rocketstack-matt not yet, I wanted to see if you agreed with the approach. Happy for you to try an example 😄, I don't have too much time today to try it

grahampacker-ms: It feels like we need a type that represents a calm definition?

jpgough-ms:
We have decided to take this a different way and I will add the discussion to the #111

PR #133, opened 2024-04-08
Issue 73 - Switching from console log to winston logger

lbulanti-ms:
Switching from using console.log to use the winston logger

### DIFF
--- a/cli/src/commands/validate/validate.ts
+++ b/cli/src/commands/validate/validate.ts
@@ -3,39 +3,56 @@ import { existsSync, promises as fs, readFileSync, readdirSync, statSync } from 
 import pkg from '@stoplight/spectral-core';
 const { Spectral } = pkg;
 import { getRuleset } from '@stoplight/spectral-cli/dist/services/linter/utils/getRuleset.js';
+import * as winston from 'winston';
 
-export default async function validate(jsonSchemaInstantiationLocation: string, jsonSchemaLocation: string, metaSchemaPath: string) {
+let logger: winston.Logger; // defined later at startup
+
+export default async function validate(jsonSchemaInstantiationLocation: string, jsonSchemaLocation: string, metaSchemaPath: string, debug: boolean) {
+    initLogger(debug);
     let exitCode = 0;
     try {
         const ajv = new Ajv2020({ strict: false });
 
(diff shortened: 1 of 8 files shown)


cli/src/commands/validate/validate.ts
```diff
@@ -3,39 +3,56 @@ import { existsSync, promises as fs, readFileSync, readdirSync, statSync } from
...
 
+function initLogger(debug: boolean): void {
+    const level = debug ? 'debug' : 'info';
+    logger = winston.createLogger({
+        transports: [
+            new winston.transports.Console()
+        ],
+        level: level,
+        format: winston.format.combine(
+            winston.format.cli(),
```
willosborne, line 49: Why the combine()? Are you not just using one format?

Also - is there any chance that we could re-use this code across the three modules?

lbulanti-ms: Good point, I was trying to see if I could print JSON using a different format I forgot to remove the combine. Should I create and helper.ts file where I can add this logic so that it can be reused across the three commands? @willosborne

willosborne: I think this is a good approach yes

cli/src/commands/validate/validate.ts
```diff
@@ -3,39 +3,56 @@ import { existsSync, promises as fs, readFileSync, readdirSync, statSync } from
 import pkg from '@stoplight/spectral-core';
 const { Spectral } = pkg;
 import { getRuleset } from '@stoplight/spectral-cli/dist/services/linter/utils/getRuleset.js';
+import * as winston from 'winston';
 
-export default async function validate(jsonSchemaInstantiationLocation: string, jsonSchemaLocation: string, metaSchemaPath: string) {
+let logger: winston.Logger; // defined later at startup
+
+export default async function validate(jsonSchemaInstantiationLocation: string, jsonSchemaLocation: string, metaSchemaPath: string, debug: boolean) {
+    initLogger(debug);
```
grahampacker-ms, line 11: should we be passing around wether we are in debug mode or not? seems like this should be some sort of global variable, or dependency injected? This could be different because it's node or CLI tool

grahampacker-ms (approved):

willosborne (approved):

PR #143, opened 2024-04-18
Add spectral rules to check patterns

willosborne:
Note that this also includes a script to strip $refs from patterns before validation. This is necessary as spectral isn't too smart about $refs and will aggressively follow them.

CLI also needs to be updated to run these checks, I will do this later


spectral/manual-spectral.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/bash
```
jpgough-ms, line 1: Just wondering if in the medium term this would become a task in the `package.json`?

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
+functions: 
+  - node-id-exists
+  - interface-id-exists
+  - node-has-relationship
+  - numerical-placeholder
+
+rules:
+  has-nodes-relationships:
+    description: "Has key top level nodes"
+    message: Should have nodes and relationships at top level
```
jpgough-ms, line 10:
```suggestion
    message: Should have nodes and relationships as top level properties on the CALM document
```

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
+functions: 
+  - node-id-exists
+  - interface-id-exists
+  - node-has-relationship
+  - numerical-placeholder
+
+rules:
+  has-nodes-relationships:
+    description: "Has key top level nodes"
```
jpgough-ms, line 9:
```suggestion
    description: "Has nodes and relationships"
```

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
...
+    message: String placeholder detected in instantiated pattern.
+    severity: warn
+    given: "$..*"
+    then:
+      function: pattern
+      functionOptions:
+        notMatch: "^{{\\s*[A-Z_]+\\s*}}$"
+
+  connects-relationship-references-existing-nodes:
+    description: Connect relationships with const properties must reference existing nodes.
```
jpgough-ms, line 48:
```suggestion
    description: Connect relationships with const properties must reference existing nodes
```

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
...
+    message: "{{error}}"
+    # given: "$.properties.relationships.prefixItems[*].properties.relationship-type.const.connects.*.node"
+    given: "$..connects.source.node"
+    then:
+      # only container and actor are in scope, since source/destination use interfaces now
+      function: node-id-exists
+
+
+  group-relationship-references-existing-nodes:
+    description: All nodes referenced by a grouping relationship must reference existing nodes.
```
jpgough-ms, line 59:
```suggestion
    description: All nodes referenced by a grouping relationship must reference existing nodes
```

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
...
+    severity: error 
+    message: "{{error}}"
+    # given: "$.properties.relationships.prefixItems[*].properties.relationship-type.const.connects.*.node"
+    given: "$..relationship-type..nodes[*]"
+    then:
+      # only container and actor are in scope, since source/destination use interfaces now
+      function: node-id-exists
+
+  relationship-references-existing-nodes:
+    description: Relationships with const properties must reference existing nodes.
```
jpgough-ms, line 69:
```suggestion
    description: Relationships with const properties must reference existing nodes
```

spectral/pattern/validation-rules.yaml
```diff
@@ -0,0 +1,98 @@
...
+    severity: error 
+    message: "{{error}}"
+    given: "$..relationship-type.const.connects.*" # both source and destination
+    then:
+      # node uses the interface structure so has a sub object
+      - field: interface
+        function: interface-id-exists
+
+  nodes-must-be-referenced:
+    description: Nodes must be referenced by at least one relationship.
```
jpgough-ms, line 93:
```suggestion
    description: Nodes must be referenced by at least one relationship
```

jpgough-ms (changes requested):

spectral/manual-spectral.sh
willosborne: My plan was to use this script from the workflows to perform validations.
I think the eventual goal is to use the CLI to perform validations in CI - i.e. just run the calm cli against any patterns we have. That way we can get JSON schema checks too.

That of course means we need to be able to strip $refs out of the documents on the CLI too - which I'm taking a look at now.

lbulanti-ms (approved):

grahampacker-ms (approved):

grahampacker-ms:
@jpgough-ms this is pending you resolving the change requests I believe

jpgough-ms (approved):

PR #144, opened 2024-04-18
Updated Visualizer to use GraphViz

aidanm3341:
In order to get visualisation working on the CLI without the use of a headless browser, I have switched to using GraphViz.

I've also moved the visualisation code into the CLI folder directly to facilitate publishing it to NPM.


cli/package.json
```diff
@@ -25,7 +25,9 @@
     "ajv": "^8.12.0",
     "commander": "^12.0.0",
```
willosborne, line 26: Are we removing Mermaid now that it's no longer needed? That should remove the dependency on Playwright/puppeteer

cli/src/commands/visualize/Types.d.ts
willosborne, line 1: This should definitely go into a shared module eventually. Other projects should be re-using this

cli/src/commands/visualize/Types.d.ts
```diff
@@ -0,0 +1,61 @@
...
+    description: string,
+    'data-classification'?: string,
+    'run-as'?: string,
+    instance?: string
+}
+
+export type CALMRelationship = CALMInteractsRelationship | CALMConnectsRelationship | CALMDeployedInRelationship | CALMComposedOfRelationship;
+
+export interface CALMInteractsRelationship {
+    'relationship-type': {
```
willosborne, line 22: Is this top level property required in every relationship type?

willosborne: i.e. can we remove duplication

cli/src/commands/visualize/calmToDot.ts
```diff
@@ -0,0 +1,128 @@
...
+    idToNode[node['unique-id']] = graphNode;
+    G.addNode(graphNode);
+}
+
+function capitalizeFirstLetter(s: string): string {
+    return s.charAt(0).toUpperCase() + s.slice(1);
+}
+
+function addRelationship(g: Digraph, relationship: CALMRelationship): void {
+    if ('connects' in relationship['relationship-type']) {
```
willosborne, line 53: Could we potentially add methods to work on relationship types such as .getRelationshipType now that we have a model for it?

willosborne: Just a thought, this works fine

cli/package.json
aidanm3341: yeah good point I forgot to remove unused deps

cli/src/commands/visualize/Types.d.ts
aidanm3341: agreed. I'll just move it now so it's done

cli/src/commands/visualize/Types.d.ts
aidanm3341: hmmm, there is some commonality between them but I think because the ConnectsRelationship has some top level fields that set it apart from the others, it would be a little tricky to have a super type here

jpgough-ms (approved):

willosborne (approved):

PR #153, opened 2024-04-19
Devcontainers

rocketstack-matt:
Setting up devcontainers to make it easier for folks to contribute


rocketstack-matt:
Ah, my bad, I think it should be fine as long as we approve ad merge the other one first. I can always cancel this and resubmit if there is a problem.

grahampacker-ms (approved):

matthewgardner (approved):

PR #161, opened 2024-04-19
Update Translator docs to make it easier to use / contribute

rocketstack-matt:


rocketstack-matt:
@Budlee when I generate the C4 DSL from Swagger or the CLI it isn't able to render at https://structurizr.com/

I have put the output in [this gist](https://gist.github.com/rocketstack-matt/7d2afc91ed87e10dbe5e0ce9b41b2392) are you able to advise if there is a problem with the output or if I'm missing something?

jpgough-ms (approved):

Issue #139, opened 2024-04-17
CLI Generate Command Does not Generate Additional Properties in CALM Schema


jpgough-ms:
### Steps to Reproduce:

Create an additional node in a pattern template e.g.

```
{
  "$schema": "https://raw.githubusercontent.com/finos-labs/architecture-as-code/main/calm/draft/2024-04/meta/calm.json",
  "$id": "https://raw.githubusercontent.com/finos-labs/architecture-as-code/main/calm/pattern/api-gateway",
  "title": "API Gateway Pattern",
  "type": "object",
  "properties": {
    "nodes": {
      ...
    },
    "relationships": {
     .... 
    },
    "custom-pattern-information" : {
     ...
    }
  },
  "required": [
    "nodes",
    "relationships",
    "custom-pattern-information"
  ]
}
``` 

Running the generate command will not add an object for `custom-pattern-information`

### Expected Result:

All properties appear in the pattern instantiation.

Perhaps a good second step to help us discover any strange bugs or possible issues in a pattern would be to run the validate command against the produced pattern instantiation?

grahampacker-ms:
@willosborne I believe this is expected behaviour as the calm generator is opinionated based on the core meta-schema (like nodes, relationships)?

willosborne:
Indeed: the generate command will only do nodes and relationships. I did write code to have it handle additional top level properties, but there were concerned this was too custom and potentially indicative of some missing design in the CALM schema.

See #92 

We can definitely add this feature if it'd be worth having.

PR #166, opened 2024-04-22
Top level nodes

willosborne:
Partially addresses #139 by instantiating additional top-level properties as if they were node definitions.


jpgough-ms (approved):

cli/src/commands/generate/generate.ts
```diff
@@ -132,22 +135,46 @@ function instantiateRelationships(pattern: any): any {
     return outputRelationships;
 }
 
+function instantiateAdditionalTopLevelProperties(pattern: any): any {
+    const properties = pattern?.properties;
+    if (!properties) {
+        console.error('Warning: pattern has no properties defined.');
```
aidanm3341, line 141: this should be logger instead of console

willosborne: Fixed

grahampacker-ms (approved):

jpgough-ms (approved):

PR #170, opened 2024-04-23
Bugfix for incorrect model on #48

jpgough-ms:


willosborne (approved):

willosborne:
This is going to break stuff - spectral rules, visualisation and generation all need changing @aidanm3341 @lbulanti-ms .
Just FYI

PR #172, opened 2024-04-23
Issue-98 - Unify spectral and JSON Schema validation output

lbulanti-ms:
* Unify the output of spectral and Json Schema
* Enabling `strict: 'log'` option when the debug flag is provided

Example of the unified output:
```
info:    Loading meta schema(s) from ../calm/draft/2024-04/meta
info:    Loading pattern from : ../calm/pattern/api-gateway.json
info:    Loading pattern instantiation from : ../calm/samples/api-gateway-implementation.json
error:   The following issues have been found on the JSON Schema instantiation [
    {
        "code": "connects-relationship-references-existing-nodes",
        "severity": "error",
        "message": "'api-consumer' does not refer to the unique-id of an existing node.",
        "path": "/relationships/0/relationship-type/connects/source/node"
    },
    {
        "code": "json-schema",
        "severity": "error",
        "message": "must be integer",
        "path": "/nodes/0/interfaces/0/port",
        "schemaPath": "#/properties/port/type"
    },
    {
        "code": "json-schema",
        "severity": "error",
        "message": "must have required property 'unique-id'",
        "path": "/nodes/1",
        "schemaPath": "#/required"
    }
]
```

### DIFF
--- a/cli/src/commands/validate/validate.ts
+++ b/cli/src/commands/validate/validate.ts
@@ -1,16 +1,20 @@
-import Ajv2020 from 'ajv/dist/2020.js';
+import Ajv2020, { ErrorObject } from 'ajv/dist/2020.js';
 import { existsSync, promises as fs, readFileSync, readdirSync, statSync } from 'fs';
-import pkg from '@stoplight/spectral-core';
+import pkg, { ISpectralDiagnostic } from '@stoplight/spectral-core';
 const { Spectral } = pkg;
 import { getRuleset } from '@stoplight/spectral-cli/dist/services/linter/utils/getRuleset.js';
+import { DiagnosticSeverity} from '@stoplight/types';
 import * as winston from 'winston';
 import { initLogger } from '../helper.js';
+import { ValidationOutput as ValidationOutput } from './validation.output.js';
+import { SpectralResult } from './spectral.result.js';
 
 let logger: winston.Logger; // defined later at startup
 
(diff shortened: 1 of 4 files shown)


jpgough-ms (approved):

willosborne (approved):

cli/src/commands/validate/validate.ts
```diff
@@ -24,19 +28,39 @@ export default async function validate(jsonSchemaInstantiationLocation: string,
...
+        errors = spectralResult.errors;
+        validations = validations.concat(spectralResult.spectralIssues);
+
         if (!validateSchema(jsonSchemaInstantiation)) {
-            logger.error(`The instantiation does not match the JSON schema pattern. Errors: ${prettifyJson(validateSchema.errors)}'`);
-            exitCode = 1;
-        } else {
-            logger.info('The schema instantiation matches the json schema');
+            logger.debug(`JSON Schema validation raw output: ${prettifyJson(validateSchema.errors)}`);
+            errors = true;
```
aidanm3341, line 37: I wonder if we could simplify this a little bit by just throwing an error here with the error message on line 42. You could then remove that whole if statement from further below and let the catch block log the error and exit. You then wouldn't need the errors variable either. Thoughts?

aidanm3341: actually I just noticed you're using the errors variable at other places too, but I still think we can probably simplify this a little

aidanm3341: discussed offline and agreed these changes wouldn't actually simplify anything. I'm happy with these changes!

aidanm3341 (approved):

PR #174, opened 2024-04-24
Follow refs when generating definitions

willosborne:
Addresses #139 partially.
Now when a node, relationship or interface references a schema with $ref, it will be resolved and generated against a set of schemas loaded with a -s property.
Note this property is optional. For now it just skips lookups if this is not set - eventually will use the bundled metaschema.

Circular references are handled by simply terminating and returning the object as it is so far.

This does **not** instantiate refs from within nodes - e.g. if you define an object in your defs and simply reference it from a property this won't be followed. It's only definitions of nodes, interfaces and relationships. This will be added in a subsequent PR.


cli/src/index.ts
```diff
@@ -31,9 +31,10 @@ program
     .description('Generate an instantiation from a CALM pattern file.')
     .requiredOption('-p, --pattern <source>', 'Path to the pattern file to use. May be a file path or a URL.')
     .requiredOption('-o, --output <output>', 'Path location at which to output the generated file.')
+    .option('-s, --schemaDirectory <path>', 'Path to directory containing schemas to use in instantiation')
```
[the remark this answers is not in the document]
rocketstack-matt: I think we'll likely want to keep this option as an override even with a bundled schema to make it easier for folks working on the schema iself.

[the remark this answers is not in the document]
rocketstack-matt: Can we add the new option to the README

cli/src/commands/generate/schema-directory.ts
```diff
@@ -0,0 +1,134 @@
...
+        if (!newSchemaId) {
+            newSchemaId = currentSchemaId;
+            this.logger.debug(`Resolving reference ${ref} against current schema.`);
+        }
+        this.logger.debug(`Recursively resolving the reference, ref: ${ref}`);
+        const definition = this.lookupDefinition(newSchemaId, ref);
+        if (!definition) {
+            // schema not defined
+            // TODO enforce this once we can guarantee we always have schemas available
+            return {};
```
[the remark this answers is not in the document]
willosborne: Added a placeholder saying the ref was missing here

lbulanti-ms (dismissed):

cli/src/commands/generate/components/node.ts
```diff
@@ -0,0 +1,113 @@
...
+import { getPropertyValue } from './property.js';
+
+/**
+ * Instantiate an individual node from its definition, resolving $refs if appropriate. 
+ * @param nodeDef The definition of the node.
+ * @param schemaDirectory The schema directory to resolve refs against.
+ * @param debug Whether to log debug detail.
+ * @returns An instantiated node.
+ */
+export function instantiateNode(nodeDef: any, schemaDirectory: SchemaDirectory, debug: boolean = false): any {
```
aidanm3341, line 15: I also wonder if we can do better on the `nodeDef` type... for example we know that it might have a `$ref` key, so we could at least have `{ '$ref'?: string }` and whatever other optional fields there may be

cli/src/commands/generate/components/node.ts
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
```
aidanm3341, line 1: I think we should definitely remove this. I know there will be errors to then fix, but it will stop the plague of `any` getting out of control. I've left some other comments with suggestions of alternatives, but I'm happy to discuss further if you disagree

cli/src/commands/generate/components/property.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+
+export function getStringPlaceholder(name: string): string {
+    return '{{ ' + name.toUpperCase().replaceAll('-', '_') + ' }}';
+}
+
+export function getPropertyValue(keyName: string, detail: any): any {
```
aidanm3341, line 7: Would this be accurate based on what you've written in this function?
```typescript
interface Detail {
    const?: unknown,
    type?: 'string' | 'integer' | 'array',
    $ref?: string
}
```

There is the `types.d.ts` file in the root cli dir which this could maybe be added too.

With this you could also change the if statements to be the following, which is a bit nicer

```typescript
if (detail.const) { ... }
```

cli/src/commands/generate/schema-directory.ts
```diff
@@ -0,0 +1,142 @@
...
+import { mergeSchemas } from './util.js';
+import { Logger } from 'winston';
+import { initLogger } from '../helper.js';
+
+/**
+ * Stores a directory of schemas and resolves references against that directory.
+ * Can merge objects recursively and will handle circular references.
+ */
+export class SchemaDirectory {
+    private readonly schemas: Map<string, any> = new Map<string, any>();
```
aidanm3341, line 15: a schema is at least an `object` rather than just `any`, right?

cli/src/commands/generate/schema-directory.ts
```diff
@@ -0,0 +1,142 @@
...
+        const newRef: string = definition['$ref'];
+        if (visitedDefinitions.includes(newRef)) {
+            this.logger.warn('Circular reference detected. Terminating reference lookup. Visited definitions: ' + visitedDefinitions);
+            return definition;
+        }
+        const innerDef = this.getDefinitionRecursive(newRef, newSchemaId, visitedDefinitions);
+        return mergeSchemas(innerDef, definition);
+    }
+
+    private getMissingSchemaPlaceholder(reference: string): object {
```
aidanm3341, line 82: I get not wanting to write out the entire return type in instances like this where it's a bit nested and constant, so I'd suggest just taking the return type away! Typescript will at least then figure it out and not just say `object` because you overrode it to be that. It's totally valid to omit the return type in functions as trivial as this

aidanm3341 (changes requested):

cli/src/commands/generate/components/property.ts
willosborne: Interesting. What would happen if there are extra properties?

cli/src/commands/generate/components/node.ts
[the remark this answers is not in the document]
willosborne: Is unknown any better than any? Is that not just going to cause failed type checks later on if we try to introduce any types?

rocketstack-matt:
@TheJuanAndOnly99 Semgrep action appears to be working on schedule but still failing to login on PR . . . is this just a limitation and something that can't be done (because PR action triggered from fork can't have access to secrets) or is there a project where this is working?

cli/src/commands/generate/components/property.ts
aidanm3341: this is where you'd then introduce the `[key: string]: unknown` to represent additional properties

cli/src/commands/generate/components/node.ts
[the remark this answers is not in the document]
aidanm3341: it depends on how you're using it. If you don't plan on extracting any info from the thing, then unknown is definitely better because you're telling the type checker that you truly don't know what's in it and to let you know if you try to access it incorrectly. In cases where you are accessing it for something, then unknown might not be the best, but there are almost always better alternatives to `any`

cli/src/commands/generate/components/node.ts
willosborne: Revisiting later

aidanm3341 (approved):
Will and I agreed to take some time later to try and improve the typing, but that we can still merge this PR in its current state. I'm happy with the changes

PR #181, opened 2024-05-02
Improve logging and instantiate ref properties

willosborne:
Small PR

### DIFF
--- a/cli/src/commands/generate/components/node.ts
+++ b/cli/src/commands/generate/components/node.ts
@@ -115,7 +115,6 @@ export function instantiateNodeInterfaces(detail: any, schemaDirectory: SchemaDi
     const logger = initLogger(debug);
     const interfaces = [];
     if (!('prefixItems' in detail)) {
-        logger.error('No items in interfaces block.');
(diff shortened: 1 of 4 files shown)


cli/src/commands/generate/components/node.ts
```diff
@@ -115,7 +115,7 @@ export function instantiateNodeInterfaces(detail: any, schemaDirectory: SchemaDi
     const logger = initLogger(debug);
     const interfaces = [];
     if (!('prefixItems' in detail)) {
-        logger.error('No items in interfaces block.');
+        logger.debug('Instantiating a node with no interfaces defined.');
```
rocketstack-matt, line 118: I don't think we want to log on this, in fact the schema needs to be updated as interfaces are only valid on certain types of node.

rocketstack-matt (dismissed):
Approving as the comment relates to lowering the threshold, but still not sure this is a log message needed.

jpgough-ms (approved):

aidanm3341 (approved):

Issue #183, opened 2024-05-03
Support Metadata Object


jpgough-ms:
### Description of Problem:

Whilst working with systems that practically use CALM nodes and patterns, we have found a need to model additional information or metadata.

### Potential Solutions:

Introduce a common type `metadata` that has an array of objects to allow for extension within the CALM schema.

As owners we will need to be considerate that metadata does not soak from things that should become concrete types in the architecture or pattern

cc/ @rocketstack-matt from our offline discussion for additional comments as needed

willosborne:
This is done. We have generation support that supports generic objects, so I think that's enough for people to fully use this feature.

PR #186, opened 2024-05-03
Add metadata to calm #183

jpgough-ms:


calm/draft/2024-04/meta/core.json
```diff
@@ -112,6 +118,9 @@
         },
         "authentication": {
           "$ref": "#/defs/authentication"
+        },
```
rocketstack-matt, line 121: I think this should be the metadata definition rather than a ref to metadata, you have the actual definition outside the defs block at the bottom of the spec.

calm/draft/2024-04/meta/core.json
```diff
@@ -232,6 +241,12 @@
           }
         }
       }
+    },
+    "metadata": {
```
rocketstack-matt, line 245: Do you have an example of how you plan to use this, trying to understand the decision to use an array of object rather than say a key/value string pair.

calm/draft/2024-04/meta/core.json
rocketstack-matt, line 1: Do we want to add metadata to nodes / relationships / interfaces as well or we're happy to just annotate at the document level?

calm/draft/2024-04/meta/core.json
rocketstack-matt: My mistake

calm/draft/2024-04/meta/core.json
rocketstack-matt: Spoke offline

rocketstack-matt (approved):

PR #187, opened 2024-05-03
Spectral rule to detect duplicated unique IDs

willosborne:
Requested by @jpgough-ms .

Also adds @jpgough-ms as a CODEOWNER on spectral and the CLI.

### DIFF
--- a/spectral/instantiation/validation-rules.yaml
+++ b/spectral/instantiation/validation-rules.yaml
@@ -3,6 +3,7 @@ functions:
   - interface-id-exists
   - node-has-relationship
   - numerical-placeholder
+  - ids-are-unique
 
 rules:
   instantiation-has-nodes-relationships:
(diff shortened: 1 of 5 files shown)


spectral/instantiation/validation-rules.yaml
```diff
@@ -88,4 +89,12 @@ rules:
     message: "{{error}}"
     given: "$.nodes[*].unique-id"
     then:
-      function: node-has-relationship
\ No newline at end of file
+      function: node-has-relationship
+
+  unique-ids-must-be-unique:
```
lbulanti-ms, line 94: can we rename this to something like `unique-ids-must-be-unique-in-instantiation`? So we can keep it consistent with the naming convention of the other rules

lbulanti-ms (approved):

PR #188, opened 2024-05-10
Add k8s end point - issues 162

matthewgardner:
Adding a simple JSON endpoint for the translator

### DIFF
--- a/translator/src/main/java/org/finos/calmtranslator/controller/Translator.java
+++ b/translator/src/main/java/org/finos/calmtranslator/controller/Translator.java
@@ -2,9 +2,14 @@ package org.finos.calmtranslator.controller;
 
 import com.structurizr.Workspace;
 import com.structurizr.util.WorkspaceUtils;
+
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+
+import java.util.List;
+
 import org.finos.calmtranslator.calm.Core;
 import org.finos.calmtranslator.translators.C4ModelTranslator;
-
+import org.finos.calmtranslator.translators.K8sModelTranslator;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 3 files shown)


matthewgardner:
@TheJuanAndOnly99 any ideas why the semgrep act is failing?

matthewgardner:
> @TheJuanAndOnly99 any ideas why the semgrep action is failing?

Is this cos I'm using a fork and my forked repo doesn't have the token?

TheJuanAndOnly99:
@matthewgardner yes, secrets are not shared with forks!

rocketstack-matt:
> @matthewgardner yes, secrets are not shared with forks!

@TheJuanAndOnly99 this is the same conversation we were having when still in labs . . . how is anyone else using semgrep if it won't work on PRs from forks?

TheJuanAndOnly99:
@rocketstack-matt you're right it doesn't make sense. I've sent a PR removing the login step https://github.com/finos/architecture-as-code/pull/201. This should allow the semgrep check to run correctly on PRs that are coming from forks.

rocketstack-matt:
@TheJuanAndOnly99 I thought we'd added it in the lab because it was failing without it. Anyway, your PR build passed so I've merged it in and will check the others.

rocketstack-matt (approved):

rocketstack-matt:
@TheJuanAndOnly99 as I suspected, the change you have made (and we previously put in) simply means the action passes without running (see https://github.com/finos/architecture-as-code/actions/runs/9175041268/job/25227175952?pr=190) - do you have an example of any repo in FINOS where this actually works?

CC: @matthewgardner

TheJuanAndOnly99:
@rocketstack-matt we do not. This is something we're trying to roll out. See https://github.com/finos/code-scanning (were we actually have the same issue). The last change I made should allow the semgrep scan to run correctly on PRs coming from branches and though it won't run on PRs coming from forks it at least doesn't block them. Let me discuss with @maoo and get back to you.

Issue #191, opened 2024-05-13
Update "internal-network" to be "network"


jpgough-ms:
## Feature Request

Whilst reviewing CALM we noticed that `internal-network` is too specific as a `node-type-definition` it should be changed to `network` and the node created can carry more information regarding whether it is internal/external/cloud etc

PR #192, opened 2024-05-13
#191 Update internal-network to network

jpgough-ms:


jpgough-ms:
@Budlee I might need your help with this. I've bumped the version of the schema to take in all the April changes and the project fails due to changes in CALM. If you have a few mins at some point happy to jump on call to discuss

cc/ @rocketstack-matt

PR #278, opened 2024-06-10
Visualizer improvments

aidanm3341:
improvements include:
- changing the layout engine which ends up looking a lot better
- visualizer renders relationship description
- updated api-gateway pattern


cli/src/index.ts
```diff
@@ -36,7 +36,8 @@ program
     .option('-v, --verbose', 'Enable verbose logging.', false)
     .option('-a, --instantiateAll', 'Instantiate all properties, ignoring the "required" field.', false)
     .action(async (options) => {
-        await runGenerate(options.pattern, options.output, options.schemaDirectory, !!options.verbose, options.instantiateAll);
+        await runGenerate(options.pattern, options.output, !!options.verbose, options.instantiateAll, options.schemaDirectory
```
willosborne, line 39: This seems to have moved all the arguments around? Is this okay?

aidanm3341: yes - they were actually the wrong way around before. Tricky to catch because the 'options' object is just a bunch of 'any's. I think there is an add-on to commander which can help with this. I'd like to look into it at some point

willosborne (approved):

PR #280, opened 2024-06-11
Add metadata generation support

willosborne:
Adds CLI support to back up changes to spec from #183 . 
If metadata is defined it will recursively attempt to instantiate the objects set within. It will resolve refs too, but only on top-level objects. 

i.e. you can have (approx, missed some detail out)

```json
"metadata": [
   {
      "$ref": "xyz"
   }
]
```

but not 

```json
"metadata": [
   {
     property: {
       "$ref": "xyz"
     }
   }
]
```

**Note that this also removes support for arbitrary additional top level properties, since this is what metadata replaces.**

### DIFF
--- a/cli/src/commands/generate/generate.ts
+++ b/cli/src/commands/generate/generate.ts
@@ -1,5 +1,3 @@
-/* eslint-disable  @typescript-eslint/no-explicit-any */
-
 import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { mkdirp } from 'mkdirp';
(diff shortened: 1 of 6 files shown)


cli/src/commands/generate/generate.ts
```diff
@@ -26,12 +25,12 @@ function loadFile(path: string): any {
 }
 
 
-function instantiateAdditionalTopLevelProperties(pattern: any, schemaDirectory: SchemaDirectory): any {
-    const properties = pattern?.properties;
-    if (!properties) {
+function instantiateAdditionalTopLevelProperties(pattern: object, schemaDirectory: SchemaDirectory): object {
```
aidanm3341, line 28: using object over any is an improvement, but we should be able to do better. The result of this function should be a valid CALMInstantiation shouldn't it?

cli/src/commands/generate/generate.ts
```diff
@@ -26,12 +25,12 @@ function loadFile(path: string): any {
...
+    if (!('properties' in pattern)) {
         logger.error('Warning: pattern has no properties defined.');
         return [];
     }
+    const properties = pattern['properties'];
```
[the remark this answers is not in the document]
willosborne: discussed this - moving type from any to object makes this necessary.

cli/src/commands/generate/generate.ts
willosborne: removing this function entirely

aidanm3341 (approved):

PR #320, opened 2024-07-29
DRAFT Pattern Validation

lbulanti-ms:
Temporary PR to share my code with @willosborne


lbulanti-ms:
This PR was used just to share some code

Issue #321, opened 2024-07-31
Minimal (MVP) Controls Domain


jpgough-ms:
### Description of Problem:

@yt-ms and I have had a discussion about creating a minimal domain for capturing controls for nodes (as a starting point). At this stage we are looking to introduce domains discussed in #310. A controls.json to model the domain and an August schema version (to not impact existing uses of CALM).

We are going to follow the approach we used for interfaces: https://github.com/finos/architecture-as-code/blob/main/calm/draft/2024-04/meta/interface.json#L6 and create a control-type.

With this in place we can continue our discussions on what we might centralize into this domain in the future for common control types. This can be the subject of discussion at our regular working group.

### Potential Solutions:

* Create an August version of CALM
* Add the general concept of domain (outlined in #310)
* Create a control-type
* Create a control-type domain (i.e. a list of control-type)

yt-ms:
I see this in two parts - the controls defined in one file, which set out what data is needed within the system (or other) node, then the data on the nodes themselves.

Example of what the controls file might look like:
```yaml
{
  "nodes": [
    {
      "uniqueid": "cr-arch-001",
      "type": "Control Requirement",
      "name": "Architectures must be approved to ensure they are fit for purpose",
      "control-registry-id": "ABCDE",    # local ID from the company's control register
    },
    {
      "uniqueId": “ci-arch-001”,
      "type": “Control Instance”,
      "name": “Architecture review pre-production”,
      "control-requirement": "cr-arch-001",    # Reference to the Control Requirement that this is an instance of
      "scope-text": "All workloads going to production",
      "scope-rego": "input.metadata.target-deployment.environment == \"production\"",    # Use rego to support complex rules
      "data-fields": {
        # ... something that works like interfaces to define which fields are needed
        # to be captured in the system as evidence of meeting the control instance
        "review-date": {},
        "reviewed-by": {},
        "docs-link": {}
      }
    }
  ]
}
```

This defines a single control _requirement_ and the corresponding control _instance_ which describes how the requirement should be met (I'm not an expert, but apparently this is a favoured structure in the industry).

The corresponding CALM architecture model might be:
```yaml
{
  "nodes": [
    {
      "unique-id": "some-system",
      "node-type": "system",
      "name": "My awesome system",
      "description": "System with fantastic architecture",
      "interfaces": [],
      "domains": {
          "cbom": [    # Control Bill of Materials
              {
                  "control-instance": "ci-arch-001",
[...]

PR #323, opened 2024-08-02
#321 First attempt at domains, controls and structure

jpgough-ms:


calm/control-example/one-node-wonder.json
```diff
@@ -0,0 +1,24 @@
...
+  "nodes": [
+    {
+      "unique-id": "example-system",
+      "node-type": "system",
+      "name": "Example System",
+      "description": "Example System",
+      "domains": {
+        "cbom": [
+          {
+            "unique-id": "ci-arch-001",
```
yt-ms, line 14: Is this the unique ID of the control instance, or of the evidence?

calm/control-example/control-specification.json
```diff
@@ -0,0 +1,46 @@
...
+    "name": {
+      "const": "Architecture review pre-production"
+    },
+    "scope-text": {
+      "const": "All workloads going to production"
+    },
+    "scope-rego": {
+      "const": "input.metadata.target-deployment.environment == Production"
+    },
+    "data-fields": {
```
yt-ms, line 18: So this is like a mini-json-schema embedded within the control specification?

calm/control-example/control-evidence.json
yt-ms, line 1: In reality, would we expect to have multiple control evidences in one json file, or have to have a file per control evidence?

calm/control-example/control-evidence.json
```diff
@@ -0,0 +1,14 @@
...
+  "$id": "control-evidence.json",
+  "title": "Evidence of pre-production review",
+  "control-id": "ci-arch-001",
+  "data-fields": {
+    "review-date": "2024-08-02",
+    "reviewed-by": "jpgough-ms",
+    "docs-link": "https://issue-tracker.com/12345"
+  },
+  "name": "Architecture review pre-production",
+  "scope-text": "All workloads going to production",
```
yt-ms, line 12: This should be in the control instance rather than the evidence of that control for a given system.

calm/control-example/control-evidence.json
```diff
@@ -0,0 +1,14 @@
...
+  "scope-rego": "input.metadata.target-deployment.environment == Production"
```
yt-ms, line 13: This should be in the control instance rather than the evidence of that control for a given system.

calm/control-example/control-specification.json
```diff
@@ -0,0 +1,46 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "control-specification.json",
+  "title": "Example Production Control Specification",
+  "properties": {
+    "control-id": {
+      "const": "ci-arch-001"
+    },
+    "name": {
+      "const": "Architecture review pre-production"
```
yt-ms, line 10: So this gets copied into each system's evidence for this control?  What would happen if one of these const fields were changed (e.g. the scope) - do people have to regenerate all the control evidence files for all the systems?

calm/control-example/one-node-wonder.json
jpgough-ms: @yt-ms calm unique id. This would be for anything that relates to this domain (we don't have that yet)

calm/control-example/control-specification.json
jpgough-ms: The schema would no longer validate and this is similar to calm where versions would need to be considered.

calm/control-example/control-evidence.json
jpgough-ms: I think this can be up to the implementor of a specific domain. I'm not sure what the best approach is, but both could be supported. This would be where a specification can supply a list of `defs` which are `$ref`'d to from a evidence/instance file

calm/control-example/control-evidence.json
jpgough-ms: I think I'm still confused about the relationship between control, instance. Maybe `control-evidence` can just be renamed to `control-instance`?

yt-ms:
Explaining what I am meaning by some of the terms:

- **Control Requirement**: this is a high-level statement of what is needed, e.g. "all code in production must meet appropriate architecture quality characteristics (a.k.a. NFRs)".
- **Control Instance**: this is the way a particular requirement is implemented, and there may be multiple of these for a single requirement.  E.g. "All systems in MyDepartment are reviewed by an approved architect before entering production, and a record of that is captured in the central review log repository".
- **Control Evidence**: this is the data that shows that a particular Control Instance was satisfied.  E.g. "On 2024-05-27, jpgough-ms did an architecture review of MySystem and approved it meets appropriate AQCs, as evidenced by document http://myfirm/reviewlog-123" (except expressed as data rather than text).

yt-ms:
There's also the potential for automated testing of the evidence, which might be done by a REGO expression in the Control Instance operating on the Control Evidence data (with hooks to pull additional info from linked repositories).

calm/control-example/one-node-wonder.json
jpgough-ms: @yt-ms we can go this way, however this would mean that we'd need the core CALM schema to understand the types of the controls domain and future domains that we add. What is there at the moment is represented by the `domain-type`. We could modify the `domain-type` to be like (may not quite be syntax correct):

```
"domain-type": {
      "type": "object",
      "properties": {
        "unique-id": {
          "type": "string"
        },
        "description": {
          "type": "string"
        },
        "detail": {
          "type": "string"
        },
        "evidence": {
          "type": "array",
          "items": {
                  "evidence-id": {
                      "type": "string"
                  },
                 "evidence-link": {
                      "type": "string"
                  },
          }
        }
      },
      "required": [
        "unique-id",
        "description",
        "detail"
      ]
    }
```
Would this be sufficient?

calm/draft/2024-08/meta/core.json
willosborne, line 1: So for now, domains are only modelled on nodes? That seems reasonable as a first cut, but also gets me thinking about what we want domains to apply to.

Nodes/relationships/top level properties I can see the point of - e.g. security or resiliency domain info on a relationship.

Interfaces I'm less convinced, but might still be useful

calm/draft/2024-08/meta/core.json
```diff
@@ -53,6 +53,17 @@
             "$ref": "interface.json#/defs/interface-type"
           }
         },
+        "domains": {
+          "type": "object",
+          "patternProperties": {
+            "^[a-zA-Z0-9_]+$": {
```
willosborne, line 59: Nice! I didn't know you could do this :octocat:

calm/draft/2024-08/meta/core.json
jpgough-ms: @willosborne  `interfaces` are on relationships, but only on certain types of relationships (to define what interface is used by a `connects` relationship for example). 

It is a good question about domains on nodes, in the discussions with @yt-ms for the controls domain it would just be `nodes` at this point. Relationships already carry a connectivity type or protocol. 

It might be worth brainstorming on a separate issue if this is needed or not. It would be trivial to add, I can move `domain` out too a definition that can be referenced from `nodes` so this would be an easy change

jpgough-ms:
Couple of things to change on the PR based on this feedback:

- [ ] Add `evidence-id` and `evidence-link` if @yt-ms agrees to the general approach
- [ ] Move the definition out of `nodes`, so it could be referenced from `relationships` if we decide to go that way

calm/draft/2024-08/meta/core.json
jpgough-ms: `JSONSchema` is a hard language to learn 😄

calm/draft/2024-08/meta/core.json
willosborne: Sounds good. I think we'll probably want to do this once we get around to doing Security as I can see a lot of potential stuff applying to connections there

willosborne (approved):

PR #330, opened 2024-08-08
Add new Interfaces for WIP Domain Example

### LINKED ISSUE #328: New Interfaces
### Description of Problem:

Whilst building a new domains-example, I needed a few additional interfaces to cover `ports` and `container-image`

### Potential Solutions:

Add to August interfaces

jpgough-ms:


willosborne (approved):
Looks good to me. 

- image name should contain the URL for the registry. We could probably put a regex on this, but probably no need for now.
- we now have port and host-port - this is fine, though

PR #338, opened 2024-08-12
Fixes #333 Apply feedback from Matt

### LINKED ISSUE #333: Should we introduce a domain-reference object?
The 2024-08 draft introduces a [domain model](https://github.com/finos/architecture-as-code/blob/main/calm/draft/2024-08/meta/domain.json). 
* The design allows new domains to be defined, as can be seen in the [control-specification](https://github.com/finos/architecture-as-code/blob/main/calm/control-example/control-specification.json) example.
* We then declare instances of a domain, as can be seen in the [control-instance](https://github.com/finos/architecture-as-code/blob/main/calm/control-example/control-instance.json) example.
* Domain instances can then be specified in CALM instance documents either at either the architecture (document), node or relationship level, as can be seen with the cbom example in [one-node-wonder](https://github.com/finos/architecture-as-code/blob/main/calm/control-example/one-node-wonder.json)

e.g.
```json
"domains": {
        "cbom": [
          {
            "unique-id": "ci-arch-001",
            "description": "Evidence of Pre Production Review",
            "requirement-schema": "control-specification.json",
            "requirement-instance": "control-instance.json",
            "evidence": [
              {
                "evidence-link": "https://evidence.com/12345",
                "evidence-id": "12345"
              }
            ]
          }
        ]
      }
```

The implementation requires the author to specify the domain specification a specific domain should adhere to and the instance being referenced, but we also duplicate the unique-id and description. Could we reduce duplication and make the document more readable by creating a new `domain-reference` object (open to different names) which would be the schema/instance and evidence container rather than referencing the domains themselves?

Tag: @jpgough-ms, @willosborne

jpgough-ms:


rocketstack-matt:
As discussed offline, what I would suggest is:

Domains should ensure all requirements map to the same domain, but should still contain evidence e.g.:

```json
"domains": {
        "cbom": 
          {
            "requirement-schema": "control-specification.json",
            "requirements": [
                {
                  "requirement-instance": "control-instance.json",
                   "evidence": [
                    {
                      "evidence-link": "https://evidence.com/12345",
                      "evidence-id": "12345"
                    }
                  ]
                },
                {
                  "requirement-instance": "control-instance-2.json",
                   "evidence": [
                    {
                      "evidence-link": "https://evidence.com/6789",
                      "evidence-id": "6789"
                     }
                   ]
                 } 
              ]
          },
          "resilience-bom": {
            "requirement-schema": "resilience-specification.json",
            "requirements": [
                {
                  "requirement-instance": "resilience-instance.json",
                   "evidence": [
                    {
                      "evidence-link": "https://evidence.com/56789",
                      "evidence-id": "56789"
                    }
                  ]
                }
             ]
          }
      }
```

I think this makes it clearer that all requirements in a domain must map to the same domain as opposed to the current implementation where they could map to different domains.

On the same line, shouldn't the requirement specification schemas reference the domain schema? e.g.

```json
{
  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/domain.json",
  "$id": "control-specification.json",
  "title": "Example Production Control Specification",
  "properties": {
    "control-id": {
      "const": "ci-arch-001"
    },
....
```

calm/draft/2024-08/meta/control.json
```diff
@@ -0,0 +1,68 @@
...
+            "base-path": {
+              "type": "string"
+            },
+            "description": {
+              "type": "string"
+            },
+            "requirements": {
+              "type": "array",
+              "items": {
+                "type": "#defs/control-requirement"
```
rocketstack-matt, line 55:
```suggestion
                "type": "#/defs/control-requirement"
```

calm/draft/2024-08/meta/control.json
```diff
@@ -0,0 +1,68 @@
...
+      },
+      "required": [
+        "control-id",
+        "name",
+        "description"
+      ]
+    },
+    "evidence": {
+      "type": "object",
+      "properties": {
```
rocketstack-matt, line 27: I think we need to add a control-configuration-uri or control-requirement-uri in the evidence object so we know what the evidence is for?

calm/control-example/pre-prod-review-specification.json
rocketstack-matt, line 1: I think we should remove all but the mandatory fields for now to make this example as simple as possible.

calm/control-example/pre-prod-review-specification.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control.json#defs/control-requirement",
   "$id": "control-specification.json",
```
rocketstack-matt, line 3: I think anywhere we're using $id we should be using an absolute URI, I think that's what we've done everywhere else we've used them. I think this use effectively changes the base URI for the retrieval URI - https://json-schema.org/understanding-json-schema/structuring#id

calm/control-example/pre-prod-review-configuration.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "control-specification.json",
+  "$schema": "pre-prod-review-specification.json",
   "$id": "control-evidence.json",
```
rocketstack-matt, line 3: Given this is an JSON instance document, not a schema, I don't think we need an id?

calm/control-example/one-node-wonder.json
```diff
@@ -8,21 +8,16 @@
...
-            "evidence": [
+      "controls": {
+        "cbom": {
+            "base-path": ".",
+            "description": "Control requirements for delivering patterns",
+            "requirements": [
               {
-                "evidence-link": "https://evidence.com/12345",
-                "evidence-id": "12345"
+                "$ref": "pre-prod-review-configuration.json"
```
rocketstack-matt, line 17: I've been playing with this locally, once I fixed the refs link in the schema I got this to correctly recognise it is an array of control-requirements, but I can only enter them inline, at least in IntelliJ, using a $ref like this is not resolving, even though I think it should.

We introduced it as a way to effectively allow tools to look for configuration files all in the same place and therefore presumably be of the same 'domain'. However, I don't think this actually works. If we can get the external refs to work correctly and provide absolute URIs then that would enable architects to traverse to the actual instance (assuming the URL is physically accessible). 

Something to discuss further.

calm/draft/2024-08/meta/control.json
rocketstack-matt, line 1: We should add a reference to the control.json into the calm.json the same way core.json is referenced otherwise the one-node-wonder doesn't know what a control-requirement is because the schema reference is to calm which currently knows nothing about controls.

calm/control-example/one-node-wonder.json
rocketstack-matt, line 17: Shouldn't this be a link to the pre-prod-review-specification.json rather than the configuration?

calm/control-example/pre-prod-review-specification.json
rocketstack-matt, line 1: Not specific to this change, but are the data-fields effectively audit data? If so wouldn't we want them on the base control-requirement specification rather than as added into this specific control-requirement?

calm/control-example/pre-prod-review-evidence.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control.json#defs/evidence",
```
rocketstack-matt, line 2:
```suggestion
  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control.json#/defs/evidence",
```

calm/draft/2024-08/meta/control.json
jpgough-ms: That's interesting, is this something we've broken with `interface.json`!

yt-ms (approved):
Looks good as far as I can see

dc-ms (approved):
Approved!

PR #349, opened 2024-08-14
Add initial resiliency domain specification 

develontopia:
This is an initial basis for discussion capturing some ideas. Not yet in line with latest domain proposal.


jpgough-ms:
It would be good to review and combine some of this work with #430. This would enable us to capture both architecture and business resiliency. @develontopia it would be good to get your input. 

@develontopia also worth checking the CLA

PR #358, opened 2024-08-24
Upgrade micromatch to remove vulnerable version

rocketstack-matt:


rocketstack-matt:
Subsequently fixed by dependabot

PR #369, opened 2024-08-28
Domains example

jpgough-ms:


calm/domains-example/security/schema/micro-segmentation.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control-requirement.json",
```
rocketstack-matt, line 2: I don't think this is the correct way to refer to the 'parent' schema . . . this says that micro-segmentation.json is an instance of the schema and expects to have the required properties. I think instead this should be like a pattern when the schema is to the calm.json and it includes a control requirement as a $ref, e.g. https://github.com/finos/architecture-as-code/blob/main/calm/pattern/api-gateway.json - do you think?

calm/domains-example/security/schema/permitted-connection.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control-requirement.json",
```
rocketstack-matt, line 2: Same comment as on micro-segmentation.json

rocketstack-matt (changes requested):

calm/control-example/pre-prod-review-evidence.json
```diff
@@ -1,8 +1,10 @@
 {
-  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/control.json#/defs/evidence",
+  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-08/meta/evidence.json",
   "evidence": {
     "unique-id": "pre-prod-review-evidence",
-    "evidence-path": "https://evidence.com/12345",
-    "control-configuration-uri": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/control-example/pre-prod-review-configuration.json"
+    "control-configuration-uri": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/control-example/pre-prod-review-configuration.json",
```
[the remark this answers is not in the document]
jpgough-ms: @rocketstack-matt it looks like this has been missed in the refactor, will fix this now

calm/domains-example/security/schema/micro-segmentation.json
jpgough-ms: I agree, there's something wrong with this as it is. Will fix this too, but I think there is a better way to do it.

rocketstack-matt (approved):

PR #385, opened 2024-09-13
Fix CVE-2024-45296

rocketstack-matt:
Suppress CVE-2023-29827 per https://github.com/mde/ejs/issues/720


rocketstack-matt:
The updated fetch-mock breaks our tests; as discussed offline we will need to find a new test framework but in the meantime this dependency will be made dev time only and warning suppressed.

PR #413, opened 2024-09-25
Update README.md to link to the docs instead of the default readme

djohnston89:
Update README.md to link to the docs instead of the default readme page


rocketstack-matt (approved):
Thanks for the contribution!

Issue #103, opened 2024-04-03
Visualizer VSCode Plugin


aidanm3341:
## Feature Request

Using the existing visualizer functionality, it would be useful to have an extension which immediately turns the CALM file to its diagram representation in a VSCode window.

### Requirements
* The extension should be triggered by running a command from the command palette 
    * "CALM: Visualize Current File" 
* The visualizer panel should open along side the existing panel
* The diagram should refresh when you save the CALM file
* Validation Errors from the CALM file should be displayed in the window so the user can address them

### Notes
* This could be implemented using the VSCode WebView capability https://code.visualstudio.com/api/extension-guides/webview 
* This existing visualisation logic lives here https://github.com/finos/architecture-as-code/tree/main/cli/src/commands/visualize 
* We might need to add an additional piece of logic to determine if the current file is a pattern or an instantiation so we know which path through the logic to take. (This detail is currently provided by the user when they execute the CLI)
    * Alternatively, the simpler approach would be instead to have **two** commands:
        * "CALM: Visualize Current Pattern"
        * "CALM: Visualize Current Instantiation"

rocketstack-matt:
As we approach v1 and the tooling is being used in anger, we want to revisit the option for enabling more productive use of CALM through offering an integrated visualization tool in VS Code

markscott-ms:
I agree @rocketstack-matt .
Whilst CALM Hub is strategic, and use of MCP for architecture & pattern creation has been demonstrated, the very industries that CALM would benefit are the ones with the most regulations, policies and procedures that will hinder adoption of those tools.

A VSCode extension running completely offline, allowing quick visual feedback, would be an excellent addition.

PR #460, opened 2024-10-08
103/vscodeplugin

Adwoa-Konadu-Appiah:
[Issue 103](https://github.com/finos/architecture-as-code/issues/103)
An initial implementation to visualize CALM instantiation when **_CALM: Calm Visualize Current Instantiation File_**  is run in the VS code command pallet.


Thels:
> Thanks for the PR @Adwoa-Konadu-Appiah!
> 
> @aidanm3341 can we get your review on this PR please.
> 
> An initial thought from me, should we place this in a top level folder called `plugins` then `vscode` that way if we create extensions for IntelliJ or other tools they can be grouped together?

Completely agree - we should go with `plugins -> vscode -> validate / visualuze whatever`

Thels:
I think we should hold fire on merging this - it looks like the `calmToDot` was copied into this extension folder.

There are problems having the extension depend on the CLI module, but this is something I believe we really need to overcome prior to merging.

calm-extension/src/types.d.ts
Thels, line 1: Ideally - this file, `visualize.ts` and `calmToDot.ts` aren't copy pasted into this directory for this to work.

I understand for a PoC this was done to produce an output, but we shouldn't merge in this state unless others disagree.

calm-extension/src/extension.ts
```diff
@@ -0,0 +1,77 @@
+import * as vscode from 'vscode';
+import { visualize } from './visualize.js';
+
+export function activate(context: vscode.ExtensionContext) {
+    let panel: vscode.WebviewPanel | undefined = undefined;
+    let svg: string | undefined;
+
+    context.subscriptions.push(
+        vscode.commands.registerCommand('calm.visualize.instantiation', async () => {
```
Thels, line 9: So how does this behave? Do we load into a `.json` and then run the `Calm Visualize Instantiation File` command from the palette.  It loads the extra panel - does this panel update as we navigate files or will it show whatever was initially rendered? 

Obviously all of this can be improved in time, just curious how you envisioned it working.

calm-extension/src/types.d.ts
aidanm3341: agreed. The approach we should take if possible is to pull in the CLI as a dependency in the package.json and pull in these files through that. As you mentioned in your other comment, there are some issues that come up regarding the mismatch between ESM and CJS, due to VSCode extensions only supporting CJS.

@Adwoa-Konadu-Appiah it would be good if you could also try doing it this way and see if you can solve the issues that come up

calm-extension/src/extension.ts
Adwoa-Konadu-Appiah: VS Code provides an API that enables extensions to read values from an active text or JSON tab and visualize the diagram in a new tab. The visualization panel will automatically update whenever the user saves changes to the file.

calm-extension/src/types.d.ts
Adwoa-Konadu-Appiah: I believe the optimal approach is to extract the logic that represents CALM as a graph into a separate npm package. This package can be maintained independently and utilized by the CLI, plugin, and any future requirements.

Thels: I agree, but I think that can wait until we have figured out the dependency issue. 

It looks to be possible to have our VSCodeExtension package depend on the `cli` but it involves compiling the `typescript` within CLI into `commonJS` format which is icky, it's stop-gap until it handles importing ESM modules ( which does look to be on the horizon - https://github.com/microsoft/vscode/issues/130367 )

Going to play around with this problem this weekend, as it's a major blocker to introducing the first IDE support for CALM which would be huge.

Adwoa-Konadu-Appiah: @Thels Has there been any update?

Thels: https://github.com/finos/architecture-as-code/pull/476

I've raised another PR which covers quite a lot of changes but the two major ones in a nutshell are;

- Introducing a shared module which both the plugin and cli depend on.  ( you can see from the PR how much of a pain-point this was - a lot of the CLI code wasn't built to be used programmatically ) 
- Plugin code which depends on this new shared module.

Thels:
I'm going to close this PR ( along with my other PR in the similar vein ) - we can still refer to the code and can re-open once the extension framework is in place.

PR #474, opened 2024-10-14
Remove database, add data-asset

### LINKED ISSUE #450: Revisit CALM Node Types
### Description of Problem:

With the introduction of `interfaces` and `controls`, we have started to challenge the `node-type-definition` that exists in CALM. The current list is:

```
"node-type-definition": {
      "enum": [
        "actor",
        "system",
        "service",
        "database",
        "network",
        "ldap",
        "webclient"
      ]
    },
```

We should revisit this to set what the types would be

arhy:
Fixes #450

### DIFF
--- a/calm/draft/2024-10/meta/core.json
+++ b/calm/draft/2024-10/meta/core.json
@@ -177,10 +177,10 @@
         "actor",
         "system",
         "service",
-        "database",
         "network",
         "ldap",
-        "webclient"
+        "webclient",
+        "data-assset"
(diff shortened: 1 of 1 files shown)


rocketstack-matt (changes requested):
I would suggest for now we add data-asset but don't remove database until we have broader agreement, I think that will also require us to update some examples to change where it is referenced.

rocketstack-matt (approved):

Issue #442, opened 2024-10-07
VSCode Extension for CALM Validation


grahampacker-ms:
### Description of Problem:
With the CALM CLI we are currently able to validate patterns and pattern instantiations.

To improve the user experience, we will create a VSCode extension that will perform continuous validations on calm patterns and patten instantiations.

markscott-ms:
This would be a feature in #103

rocketstack-matt:
This was implemented long ago

PR #476, opened 2024-10-15
#442 VSCode Extension + Other Refactors - WIP

Thels:
1. Introduction of a shared module - this is the migration of all core type-script logic from the CLI previously as we want to share it between the CLI and other purposes.
2. Shared package compiles to CommonJS - this is so that the VSCode extension can depend on it.
3. CLI now is a relatively empty module, but all CLI specific logic can live in there ( and we can use some of the fancier CLI features in the future )
4. In the shared module ( formally CLI ) made the validate function less of a god function. It now only cares about performing validations - system exit logic and how those are output/interpreted fall to others.


Thels:
There are still many things to do to clean up this PR ( some mentioned above ) - but I wanted to get the WIP up and visible so that it was easier to share with others working on similar tasks.

Ideally, to make this easier to review and reduce the blast radius of the change this would be handled in multiple PR's instead of one mega-one.

1. The first PR would simply move from CLI -> CLI + Shared modules ( whatever we end up naming shared ) - no extensions.
2. Second PR would introduce the extension is isolation.

cli/src/index.ts
```diff
@@ -16,9 +19,8 @@ program
     .addOption(new Option('-i, --instantiation <file>', 'Path to an instantiation of a CALM pattern.').conflicts('pattern'))
     .addOption(new Option('-p, --pattern <file>', 'Path to a CALM pattern.').conflicts('instantiation'))
     .requiredOption('-o, --output <file>', 'Path location at which to output the SVG.', 'calm-visualization.svg')
-    .option('-v, --verbose', 'Enable verbose logging.', false)
-    .action(async (options) => { 
-        if (!options.instantiation && ! options.pattern) {
+    .option('-v, --verbose', 'Enable verbose logging.', false)    .action(async (options) => {
```
aidanm3341, line 22: looks like a new line was accidentally deleted? `.action(...` should be on a new line

plugins/calm-vscode/README.md
```diff
@@ -0,0 +1,74 @@
+# calm-validate README
+
+
+TODO - this needs to be populated once initial features are agreed upon.
```
aidanm3341, line 4: What are we thinking for initial features then:

- command to visualize instantiation and pattern
- the in-file validation errors and warnings

That should be plenty for a first release

plugins/calm-vscode/src/extension.ts
```diff
@@ -0,0 +1,211 @@
...
+                null,
+                context.subscriptions
+            );
+        }
+    });
+
+    const validateCommand = vscode.commands.registerCommand('calm-vscode.validate', () => {
+        const diagnosticCollection = languages.createDiagnosticCollection('calm');
+        vscode.workspace.onDidSaveTextDocument(async (document: vscode.TextDocument) => {
+            if (document.uri.fsPath.endsWith('.json')) {
```
aidanm3341, line 52: not super familiar with building these extensions still so could be wrong - but is this required given the command is configured in the package.json to only work with json files?

Thels: Fair spot, I added in the logic into the package.json far later.

plugins/calm-vscode/src/extension.ts
```diff
@@ -0,0 +1,211 @@
...
+            line_end = data[0].line_end;
+            character_start = data[0].character_start;
+            character_end = data[0].character_end;
+        }
+    }
+
+    return new Diagnostic(new Range(new vscode.Position(line_start, character_start), new vscode.Position(line_end, character_end)), calmValidation.message, convertSeverity(calmValidation.severity));
+}
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
aidanm3341, line 114: how come we need this? all variables look like they're being used to me

Thels: Another fair spot, this was back when paths were hard-coded and `pageData` wasn't actually being used - it is now obviously!

plugins/calm-vscode/src/extension.ts
```diff
@@ -0,0 +1,211 @@
...
+    return await validate(instantiation, pattern, schema_location, false);
+}
+
+async function validationIssueCount(pageData: string, absFilePath: string): Promise<number> {
+    const validationOutcome = await callCalmValidate(pageData, absFilePath);
+    return validationOutcome.allValidationOutputs().length;
+}
+
+function singleValueFrom(pageData: string, jsonPath: string): string {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
aidanm3341, line 141: same question here, can we remove it?

plugins/calm-vscode/vsc-extension-quickstart.md
aidanm3341, line 1: the tips in this file are probably useful to keep around given most of us aren't too experienced with extensions. Maybe we could rename to `developer-guide.md` or something similar instead

shared/.eslintrc.json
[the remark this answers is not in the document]
Thels: Was already looking into this, given that the three of them are duplicated

cli/src/index.ts
Thels: Yeah not going to pretend I understand how this happened - it's resolved now.

PR #487, opened 2024-10-17
WIP Interactive Visualizer

aidanm3341:
This is a **WIP PR**. There are some obvious issues with what is here, but I just wanted to share what I have so far.
 
This is a tool I threw together as part of a recent hackday for CALM. The idea of this tool is to be an interactive visualizer which gives the user a much more control over the layout of their visualization.

The main feature is the ability to drag around the nodes and save a "layout" file which remembers the locations that you dragged them to. When you load an instance into the UI, you can also upload that layout file and it will place the nodes back in the correct spots.

Here's an example using traderx:
Screenshot 2024-10-17 at 14 49 19

TODO:
- investigate and fix why there are some duplicated relationship arrows
- add support for nested nodes to represent the `deployed-in` and `contains` relationships.
- make a nicer looking toolbar
- write tests (sorry, no time in a hackday for tests :D) 
- save the layout to the place it was loaded from, not just the downloads folder
- improve how the user uploads the files. Can we make it so the user only needs to upload one file instead of two?
- add zoom feature, the traderx example is already too cramped 
- rely on the types from the shared module once [this PR](https://github.com/finos/architecture-as-code/pull/476) gets merged
- add **some** kind of default layout. Currently on a new instance file, all the nodes will just overlap in the middle of the screen
- render patterns as well as just instantiations


aidanm3341:
I've created a branch in the main repo called "visualizer" that we can merge this into for now so that others can more easily work on this code. Once it's in a better state we can then merge that into main.

rocketstack-matt (approved):

PR #493, opened 2024-10-18
Update docs to publish to calm.finos.org via s3 and configure to use …

### LINKED ISSUE #491: Publish CALM docs to calm.finos.org
### Description of Problem:
We currently publish our docs to GitHub pages, it would provide a better branding experience to publish them to calm.finos.org along side the schema and samples.

### Potential Solutions:
Add a GitHub Action to build the docusaurus portal and sync to S3 instead of publishing to Pages.

The following secrets have been added to the repository:
AWS_S3_ACCESS_KEY
AWS_S3_SECRET_ACCESS_KEY

The bucket address is in repository variable AWS_S3_BUCKET without a trailing slash.

rocketstack-matt:
…Git LFS for .mov files.

Resolves #491

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.mov filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/s3-docs-sync.yml
+++ b/.github/workflows/s3-docs-sync.yml
@@ -1,0 +1,38 @@
+name: Sync Docs to S3
+
+on:
+  workflow_dispatch: {}
+  push:
+    branches: [ "main" ]
+    paths: [ "docs/**" ]
+
+jobs:
+  sync-to-s3:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: docs
+
+    steps:
+    - name: Checkout repository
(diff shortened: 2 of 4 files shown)


rocketstack-matt:
Will remove the GitHub Pages publishing in a later PR assuming this all publishes as expected

Thels:
First time seeing a `.mov` as part of a PR - next up is the entire Shrek movie!

Thels (approved):
Overall looks fine - I worry that if we add more folders into docs that we will forget to update `s3-docs-sync.yml` and end up publishing things we shouldn't to the FINOS S3 bucket but it's a small risk that is mitigated by proper review processes.

rocketstack-matt:
> Overall looks fine - I worry that if we add more folders into docs that we will forget to update `s3-docs-sync.yml` and end up publishing things we shouldn't to the FINOS S3 bucket but it's a small risk that is mitigated by proper review processes.

Everything in docs would automatically get included in the sync. The excludes are to prevent the --delete flag from removing other directories we sync (draft and samples); this is because we're copying to contents of docs/build to root rather than into it's own subdir.

PR #503, opened 2024-10-18
Update how we pull version from package.json

### LINKED ISSUE #283: Version reported is incorrect
## Bug Report
--version always prints 0.1.0 because it's hard coded. This should report current version (without hard coding)

### Steps to Reproduce:
- Install version 0.1.4
- Run --version
- See output

### Expected Result:
0.1.4

### Actual Result:
0.1.0

### Proposed solution
Use the following code in index.ts:
```
import {version} from './package.json';
```
and then print that instead of a hardcoded string.

rocketstack-matt:
Resolves #283 (hopefully)


cli/src/index.ts
```diff
@@ -1,13 +1,17 @@
...
 
 import { Option, program } from 'commander';
+import { createRequire } from 'module';
 import { visualizeInstantiation, visualizePattern } from './commands/visualize/visualize.js';
 import { runGenerate } from './commands/generate/generate.js';
 import validate  from './commands/validate/validate.js';
 import { CALM_META_SCHEMA_DIRECTORY } from './consts.js';
 
+const require = createRequire(import.meta.url);
+const { version } = require('../package.json');
```
Thels, line 11: Modern linters will block this from being merged 

image

Can I suggest adding `resolveJsonModule: true` to `tsconfig.json` and then changing this to `import { version } from '../package.json';`

willosborne (dismissed):

cli/src/index.ts
rocketstack-matt: Updated per suggestion

Thels (approved):
Thanks for making the suggested changes :)

PR #511, opened 2024-10-22
Fix CLI issue caused from resolving version from package.json

rocketstack-matt:
Referencing a file (in our case package.json) outside of the TS project has changed the dist structure, pushing the commands below /src. Reverting that change to fix the CLI.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -24,7 +24,6 @@
   "license": "ISC",
   "type": "module",
   "dependencies": {
-    "@finos/calm-cli": "^0.2.3",
     "@rollup/plugin-commonjs": "^28.0.1",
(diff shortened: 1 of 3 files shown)


cli/package.json
```diff
@@ -24,6 +24,7 @@
   "license": "ISC",
   "type": "module",
   "dependencies": {
+    "@finos/calm-cli": "^0.2.3",
```
willosborne, line 27: Maybe misunderstanding - is this not a circular dependency?

rocketstack-matt: Yep, erroneous npm install when testing the module, removed

willosborne (approved):

Issue #486, opened 2024-10-17
Properly generate $ref types on nodes, interfaces


willosborne:
### Description of Problem:
Currently the generator can generate arrays, strings, integers, consts properly. But if you have a schema with a $ref to a type, it just outputs `{{ REF_[ key name ] }}` and doesn't recursively instantiate.

For example, generating an interface that is based on the `rate-limit-interface` defined [here](https://github.com/finos/architecture-as-code/blob/main/calm/draft/2024-04/meta/interface.json#L102). The usage example is the following. Note that only the interfaces are shown for brevity:

```json
{
    "$ref": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-04/meta/interface.json#/defs/rate-limit-interface",
    "type": "object",
    "properties": {
        "unique-id": {
            "const": "rate-limit-info"
        }
    }
}
```

```json
{
          "unique-id": "unique-id-here",
          "key": "{{ REF_KEY }}",
          "time": -1,
          "time-unit": "{{ REF_TIME_UNIT }}",
          "calls": -1
 }
```

You can see `key` is a flat string that contains none of the structure of the real definititon.
What this should look like is:

```json
{
    "key": {
        "rate-limit-key": "{{ RATE_LIMIT_KEY }}"
    }
}
```

This is really confusing because it implies it's just a single field, not an object - and obscures optional/additional fields.

### Potential Solutions:
We can already do this recursive instantiation for metadata objects.
A relatively simple solution would be to unify the instantiation code to all use the generic/recursive system which would then bring these benefits to nodes and relationships.
Note that we would still need to keep the node/relationship-specific logic and merge the final objects together.

willosborne:
This is done as of #521

PR #521, opened 2024-10-25
#486 Unify instantiation mechanisms

willosborne:
Make nodes, relationships, metadata and interfaces all use the same code for instantiation

Addresses #486


rocketstack-matt (commented):
> Make nodes, relationships, metadata and interfaces all use the same code for instantiation
> 
> Addresses #486

Is #486 actually addressed in this PR? Interfaces seem to be generated exactly the same way, using the following pattern

```json
{
  "$schema": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-04/meta/core.json",
  "title": "Test CLI Pattern",
  "type": "object",
  "properties": {
    "nodes": {
      "type": "array",
      "minItems": 1,
      "prefixItems": [
        {
          "$ref": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-04/meta/core.json#/defs/node",
          "type": "object",
          "properties": {
            "unique-id": {
              "const": "system-001"
            },
            "name": {
              "const": "Demo System"
            },
            "description": {
              "const": "Demo System to test generation"
            },
            "node-type": {
              "const": "system"
            },
            "interfaces": {
              "type": "array",
              "minItems": 1,
              "prefixItems": [
                {
                  "$ref": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/draft/2024-04/meta/interface.json#/defs/rate-limit-interface",
                  "type": "object",
                  "properties": {
                    "unique-id": {
                      "const": "rate-limit-info"
                    }
                  }
                }
              ]
            }
          }
        }
      ]
    },
    "required": [
      "nodes"
    ]
  }
}
```

I get exactly the same instantiation whether generating with the 0.2.4 version of the CLI from npm or the locally built and installed version:

```json
{
  "nodes": [
    {
      "unique-id": "system-001",
      "node-type": "system",
      "name": "Demo System",
      "description": "Demo System to test generation",
      "interfaces": [
        {
          "unique-id": "rate-limit-info",
          "key": "{{ REF_KEY }}",
          "time": -1,
          "time-unit": "{{ REF_TIME_UNIT }}",
          "calls": -1
        }
      ]
    }
  ],
  "relationships": [],
  "metadata": []
}
```

If the intent of this PR is purely refactor and not change the behaviour then seems correct.

We should probably also bump the patch version of the CLI in package.json and index.ts to 0.2.5 to ensure folks realise there was a change.

willosborne:
That's interesting, I have not seen the same behaviour. Let me try to reproduce, maybe there's something I've missed

cli/src/commands/generate/components/instantiate.ts
```diff
@@ -0,0 +1,92 @@
...
+    const logger = initLogger(debug);
+    let fullDefinition = definition;
+    if (definition['$ref']) {
+        const ref = definition['$ref'];
+        const schemaDef = schemaDirectory.getDefinition(ref); 
+
+        fullDefinition = mergeSchemas(schemaDef, definition);
+    }
+    // TODO rework to properly separate 'verbose' from 'debug' level logging
+    // logger.debug('Generating ' + objectType + ' object from ' + JSON.stringify(fullDefinition));
```
Thels, line 17:
```suggestion
```

I suggest we remove this, and instead raise an issue to address the general logging across the CLI invoked methods. What are your thoughts?

cli/src/commands/generate/util.ts
```diff
@@ -24,4 +25,12 @@ export function logRequiredMessage(logger: Logger, required: string[], instantia
     } else {
         logger.debug('Required properties: ' + required);
     }
+}
+
+export function appendPath(path: string[], element) : string[] {
```
Thels, line 30: **minor** : You've typed everything here but element. Can I suggest you make this generic potentially?



```suggestion
export function appendPath<Type>(path: Type[], element: Type) : Type[] {
```

Then usage would be `appendPath<string>(array, element)`

willosborne:
Alright, so some considerable more diving has revealed my problem.
The schema directory is smart enough to recursively resolve $refs when it's pulling out a definition.
But once you go to actually generate that definition later on, it's already thrown away the info about which schema those definitions belong to.

i.e. a $ref: #/defs/abcd within interface.json is indistinguishable from one that references a local def at the bottom of the current pattern.

So what I'm working on now is a means to disambiguate the local references (i.e. without a schema ID, beginning with just a #) by turning them into absolute references including the full URI.

Note that just using file references is probably going to break this for now. e.g. interface.json#/defs/rate-limit-key from another file won't get fully disambiguated to https://calm.com/pattern/etc 

This is a little more gnarly than I thought; I'm wondering if maybe I should introduce a secondary data structure to track where things came from when they get generated.

willosborne:
Made some big improvements here. It now qualifies all relative references (at least, within the same file) with the full ID.
Also can do enums now too:
image
I also added a logging hint as to the potential values for an instantiated enum: 
image

cli/src/commands/generate/components/instantiate.ts
willosborne: Agreed!

shared/src/commands/generate/schema-directory.spec.ts
```diff
@@ -45,6 +45,16 @@ describe('SchemaDirectory', () => {
         expect(interfaceDef.properties).toHaveProperty('port');
         expect(interfaceDef.properties).toHaveProperty('unique-id');
     });
+    
+    it('qualify relative references within same file to absolute IDs', async () => {
+        const schemaDir = new SchemaDirectory();
+        
+        await schemaDir.loadSchemas('../calm/draft/2024-04');
```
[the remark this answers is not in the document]
willosborne: No particular reason. Would you mind if I got this larger PR merged before going back and making these changes? It keeps getting broken by other changes 😆

grahampacker-ms (approved):

PR #565, opened 2024-11-06
Update Docs badge to reflect using S3; rename npm publishing action t…

rocketstack-matt:
…o make it clearer what it does.


Thels (approved):
Approving and merging - this is a great reminder that I need to introduce a new badge and remove the one from CLI with my refactor PR.

PR #568, opened 2024-11-07
Added labels to interactive visualiser (finos#568)

oliviajanejohns:
See screenshot below:
Screenshot 2024-11-07 at 14 39 00

I've looked further into JSPlump and it seems that it doesn't provide an easy way of preventing overrides of labels out of the box. Will improve in a future PR but also would suggest utilising a different graph library for example D3

### DIFF
--- a/calm-visualizer/src/components/graph/Graph.tsx
+++ b/calm-visualizer/src/components/graph/Graph.tsx
@@ -10,8 +10,6 @@ interface Props {
     relationships: RelationshipLayout[]
 }
 
-// const groups: string[] = [];
-
 function createConnectsRelationship(instance: BrowserJsPlumbInstance, relationship: CALMConnectsRelationship) {
     const r = relationship["relationship-type"]["connects"];
     instance.connect({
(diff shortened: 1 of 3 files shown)


calm-visualizer/src/components/graph/Graph.tsx
```diff
@@ -29,7 +27,8 @@ function createConnectsRelationship(instance: BrowserJsPlumbInstance, relationsh
             options: {}
         },
         overlays: [
-            { type:"Arrow", options:{location:1}}
+            { type:"Arrow", options:{location:1}},
+            {  type: "Label", options: { location: 0.5, label: relationship.description }}
```
willosborne, line 31: Might want to make the formatting a bit more consistent here

willosborne, line 31: Will this cause any problems if there isn't a description defined on the relationship?

willosborne:
Looks cool and adds a lot of detail. 
Some of the lines of text are very long and quite hard to follow; is there any way we could wrap (or even truncate with an on-hover expansion) text that goes over a certain size?

willosborne (changes requested):

calm-visualizer/src/components/graph/Graph.tsx
oliviajanejohns: label also accepts undefined as a value in the case the description isnt defined

oliviajanejohns:
Adjusted styling slightly - however looking at perhaps using d3.js or other libraries:
Screenshot 2024-11-12 at 00 24 38

aidanm3341 (approved):

Issue #313, opened 2024-07-26
Unify command interfaces to use the same flags


willosborne:
### Description of Problem:
Currently the different commands have some inconsistencies in their usage that may be confusing to users.
For example, the validate command uses -m or --metaSchemasLocation to set a directory of schemas to load, whereas generate uses -s or --schemaDirectory.

These should be made consistent.

### Potential Solutions:
Use --schemaDirectory in the case given. But we should also consider other parameters.
Documentation should be updated, too.

willosborne:
We should also make the -m flag on validate behave the same as -s on generate - it should recursively traverse the directory and load all schemas. Currently it only looks at the top level folder.

rocketstack-matt:
I may have inadvertently resolved this with https://github.com/finos/architecture-as-code/pull/530 as I moved the params to be const strings.

aidanm3341:
I was just looking at that s well @rocketstack-matt, but I think the extracted flags still have the same issue. I'll delete one and update the commands to use the same const.

aidanm3341:
addressed in #571

PR #571, opened 2024-11-08
(#313) use consistent argument names

aidanm3341:


cli/src/index.ts
```diff
@@ -55,7 +54,7 @@ program
     .description('Validate that an instantiation conforms to a given CALM pattern.')
     .requiredOption(PATTERN_OPTION, 'Path to the pattern file to use. May be a file path or a URL.')
     .option(INSTANTIATION_OPTION, 'Path to the pattern instantiation file to use. May be a file path or a URL.')
-    .option(META_SCHEMA_OPTION, 'The location of the directory of the meta schemas to be loaded', CALM_META_SCHEMA_DIRECTORY)
+    .option(SCHEMAS_OPTION, 'The location of the directory of the meta schemas to be loaded', CALM_META_SCHEMA_DIRECTORY)
```
rocketstack-matt, line 57: If we're making the param name the same, shouldn't we do the same with the description?

willosborne (approved):

PR #578, opened 2024-11-11
[#422] - Migrate CLI to Shared+CLI modules.

Thels:
Splitting up of the VSCode Extension PR. This  PR is the migration of the logic within CLI into Shared ( with the CLI only logic remaining in CLI ).

Majority of the code is a move, all tests are updated along with associated GH workflows.


cli/package.json
```diff
@@ -4,71 +4,43 @@
...
   "files": [
-    "dist"
+    "dist/"
   ],
   "scripts": {
-    "build": "tsc && npm run copy-calm-schema && npm run copy-spectral-rules",
-    "test": "jest --verbose",
+    "build": "tsc",
+    "watch": "tsc -watch -p ./",
+    "clean": "rimraf node_modules dist",
```
willosborne, line 12: What does this do?

shared/jest.config.js
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line no-undef
+module.exports = {
```
willosborne, line 2: I can't see anything in here about matching *.spec.ts files. 
Does this inherit from the base one for 'cli' somehow or is there some smarter way of doing it? 
(I set it up from google first time around, so there probably is!)

shared/package.json
```diff
@@ -0,0 +1,76 @@
+{
+    "name": "@finos/calm-shared",
+    "version": "0.2.1",
```
willosborne, line 3: Do we now have multiple versions to keep track of? Or is the whole project on one version?

cli/package.json
Thels: Just gives you the ability to go back to how the folder looked post-clone, `rimraf` is just a fancy wrapper around `rm -rf` :)

shared/package.json
Thels: This is something I hadn't really considered, currently `shared` is entirely internal.  Potentially should consider marking the `package.json` private somehow.

shared/jest.config.js
Thels: Earlier in the file there is a `transform` to pass over anything as `.*.ts` to be handled by `ts-jest` which has that configured by default.

cli/package.json
Thels: It's honestly not required, it's a personal preference of mine to sometimes go back to how the code-base looks at clone to ensure that I don't have anything local holding things together.

shared/package.json
willosborne: Do we even _need_ a version beyond the top level one that the entire project uses?

shared/jest.config.js
willosborne: Ah i see, so you basically just pass it along. Nice

willosborne (approved):

shared/package.json
Thels: What about the following;

- Increment minor version within shared - so that CLI and Shared can have different versions.
- Once this PR is merged,  I'll be updating the CLI `package.json` to refer to an explicit version of `shared` module instead of a file reference ( so that version upgrades are managed and CLI version increments can occur when needed ) 
- Mark `shared` `package.json` to have `"private": true`

PR #619, opened 2024-11-22
Update vizualizer to include latest from main

oliviajanejohns:


aidanm3341:
given that we are going to continue with the cytoscape branch instead, I am closing this PR

PR #634, opened 2024-11-27
Pretty Human Readable Output of Validate CLI

Thels:
Introducing a `-pretty` flag to the validate command.

image

Also introduced the ability to `watch` coupled with `tsup`. While ideally these wouldn't be coupled into one PR I made sure `watch` was working when developing this very PR.

### DIFF
--- a/shared/src/commands/validate/validate.ts
+++ b/shared/src/commands/validate/validate.ts
@@ -9,7 +9,8 @@ import * as winston from 'winston';
 import { initLogger } from '../helper.js';
 import { ValidationOutput, ValidationOutcome } from './validation.output.js';
 import { SpectralResult } from './spectral.result.js';
-import createJUnitReport from './junit-report/junit.report.js';
+import createJUnitReport from './output-formats/junit-output.js';
+import prettyFormat from './output-formats/pretty-output';
 
 let logger: winston.Logger; // defined later at startup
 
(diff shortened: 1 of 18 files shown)


willosborne (dismissed):

shared/src/commands/validate/validate.ts
```diff
@@ -42,38 +43,34 @@ export function exitBasedOffOfValidationOutcome(validationOutcome: ValidationOut
     process.exit(0);
 }
 
-export function getFormattedOutput(
+export enum OutputFormat {
```
aidanm3341, line 46: enum's are generally avoided in Typescript - unless there's a reason I'm missing, I think we should do `type OutputFormat = 'junit' | 'json' | 'pretty'` instead. That also eliminates the need for a default branch in the switch statement below

shared/src/commands/validate/validate.ts
```diff
@@ -295,7 +279,7 @@ export async function validate(
 
         await loadMetaSchemas(ajv, metaSchemaPath);
 
-        logger.info(`Loading pattern from : ${jsonSchemaLocation}`);
+        logger.info(`Loading MARKER pattern from : ${jsonSchemaLocation}`);
```
aidanm3341, line 282: `MARKER`?

Thels: Ah, yes - this was what I was adding/removing to determine if the multi-package `watch` was working with the custom `watch.mjs` script :) Will remove this prior to seeking final approval/merge.

Thels: This is now removed - resolving.

aidanm3341 (approved):

Issue #639, opened 2024-11-30
GitHub Action Review - Consistent Naming Conventions - Coverage Gaps


Thels:
### Description of Problem:
There are a number of Github actions within the project - however they are needing a minor touch up;

- Naming convention of the actions don't always align - like between how Shared and CLI are built.
- Some actions are missing like build-time checking of the `docs` module.

### Solution:

1. Introduce consistent naming and document these standards within the root level `README.md`.
2. Ensure if we rename any files/jobs that the root level `README.md` URL's are updated for the status badges.
3. Identify ( and document on this issue ) any missing Github actions, implementing them if it isn't too much effort.

Thels:
I think it's a little out of the scope of this chore issue - but from looking at other projects in the real world - they tend to use a github action around `https://github.com/release-drafter/release-drafter` - which uses the content of our commits to govern version changes so that we don't need to bother. What do folks think?

PR #643, opened 2024-12-03
#639 - Github Action Chore - Standardizing Existing - Introducing Docs Build for PRs

Thels:


.github/workflows/build-shared.yml
```diff
@@ -0,0 +1,39 @@
+name: Build CLI
```
rocketstack-matt, line 1: Do you want this to have the same name as the build-cli action?

rocketstack-matt (approved):

Issue #650, opened 2024-12-03
Support for Ecosystem Node Type and Define a flow section within core.json


LeighFinegold:
## Feature Request

The current implementation of the architecture-as-code framework expects all nodes to be defined within a single document. However, in enterprise systems, ecosystems are composed of multiple interconnected systems, and business flows often span these systems. To better represent real-world scenarios and facilitate enterprise-wide architectural documentation, I propose two enhancements:

**Enhancement 1:** Introduce a Node Type for Ecosystem

A new node type, eco-system, should be added to represent a collection of related systems within a larger ecosystem. This will help organize and abstract complex architecture into manageable layers, aligning better with enterprise structures.

At a larger scale, Traderx could be considered as an ecosystem consisting of a Trading System, Ref Data System, Trade System, and Position System. 
Example -  ![traderx_ecosystem](https://github.com/user-attachments/assets/0226ee9a-9a6f-46f7-9910-09d3e92e9e2d)

Each can be defined as a system node and grouped under an overarching ecosystem node.

**Enhancement 2:**  Define Flows Within Core.json:

As part of https://github.com/finos/architecture-as-code/issues/426, The ability to define business flows was added. There was an expectation that eventually something would bundle flow definitions with node architectures so that node references could be derived. The proposal here is actually to define the flows section in core.json such that there is a direct link to everything that spans the ecosystem. 

### Benefits

1. Enhanced Documentation: The ecosystem and flow enhancements would enable more intuitive and comprehensive documentation for enterprise systems.
2. Improved Traversal with CALM-Hub: With the planned rollout of CALM-Hub, these features will enable automatic traversal to bundle related CALM documents for building documentation websites.
3. Scalability: Allows for modular and distributed/FrontToBack documentation, reflecting the distributed nature of modern enterprise systems.

### Alternative approaches
The flow.json spec is enhanced such that at the top of the flow document every architecture spec (core.json) that contains all referenced nodes is specified.

rocketstack-matt:
**Enhancement 1**: Introduce a Node Type for Ecosystem
tl;dr I agree with the proposal to add an ecosystem node-type,

However, I don't agree that current implementation expects all nodes to be defined in a single document.

You could already have a document that shows

```json
{
  "$schema": "https://calm.finos.org/draft/2024-10/meta/calm.json",
  "nodes": [
    {
      "unique-id": "traderx-trading-system",
      "node-type": "system",
      "name": "Trading System",
      "description": "Simple Trading System",
[...]

PR #651, opened 2024-12-03
Creating 2024-12 Spec

LeighFinegold:
Additionally moving existing traderx sample into previous released draft folder.

This is pre-req to changing the spec with proposed enhancements like https://github.com/finos/architecture-as-code/issues/650


LeighFinegold (commented):
Intentionally added sample to 2024-10 and not 2024-12 once confirmed folks are happy with proposal mentioned in https://github.com/finos/architecture-as-code/issues/650

rocketstack-matt (approved):

PR #660, opened 2024-12-09
Add ecosystem to node spec and tie flows into core.json

LeighFinegold:
This PR addresses the feature request detailed in issue [650](https://github.com/finos/architecture-as-code/issues/650) by implementing the following enhancements to CALM

1. Adds ecosystem definition to node-type
2. Add flows to core.json. This will allow the eventual traversal of calm-hub by calm-docify
3. It additionally contains some small changes to $id and $url references with the intention that these documents will be hosted network addressable and in calm-hub

### DIFF
--- a/calm/draft/2024-12/meta/core.json
+++ b/calm/draft/2024-12/meta/core.json
@@ -20,6 +20,12 @@
     },
     "controls": {
       "$ref": "control.json#/defs/controls"
+    },
+    "flows": {
+      "type": "array",
+      "items": {
+        "$ref": "flow.json#/defs/flow"
+      }
     }
   },
   "defs": {
@@ -175,6 +181,7 @@
     "node-type-definition": {
       "enum": [
         "actor",
+        "ecosystem",
         "system",
         "service",
         "database",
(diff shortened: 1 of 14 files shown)


rocketstack-matt:
I know this isn't added in this PR, but I do think we should break out the way controls are modelled here, each control should be a separate document rather than all bundled, but happy to discuss that in a separate PR.

calm/draft/2024-12/meta/flow.json
```diff
@@ -76,6 +39,45 @@
...
+          "description": "Link to a detailed requirement document"
+        },
+        "transitions": {
+          "type": "array",
+          "items": {
+            "$ref": "#/defs/transition"
+          }
+        },
+        "controls": {
+          "$ref": "https://calm.finos.org/draft/2024-12/meta/control.json#/defs/controls"
```
rocketstack-matt, line 69: We should change to a local file ref
```suggestion
          "$ref": "control.json#/defs/controls"
```

LeighFinegold: Fixed on latest commit

calm/draft/2024-12/meta/core.json
```diff
@@ -175,6 +181,7 @@
     "node-type-definition": {
       "enum": [
         "actor",
+        "ecosystem",
```
jpgough-ms, line 184: I'm not sure what this is for? Will continue looking through the review.

jpgough-ms (approved):

calm/draft/2024-12/meta/core.json
LeighFinegold: Explained on linked issue. If we don't want to pull over that terminology, happy to go with something else

PR #665, opened 2024-12-11
Prototype of core schema validation

jpgough-ms:


calm/spectral/calm-rules.yaml
```diff
@@ -0,0 +1,66 @@
...
+      functionOptions:
+        schema:
+          type: object
+          properties:
+            $schema:
+              type: string
+              pattern: ^https://json-schema\.org/draft/2020-12/schema$
+            $id:
+              type: string
+              pattern: ^https://calm\.finos\.org/draft/\d{4}-\d{2}/.*\.json$
```
willosborne, line 53: This seems to duplicate the logic of the last 3 rules - not sure if i'm missing something as to what this is for?

calm/spectral/calm-rules.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+  calm-cross-references:
+    description: 'Validates cross-references between schemas'
+    message: 'Invalid cross-reference detected: {{error}}'
+    severity: error
+    given: $..$ref
+    then:
+      function: pattern
+      functionOptions:
+        match: ^(https://calm\.finos\.org/draft/\d{4}-\d{2}/.*\.json|#/.*)$
```
willosborne, line 66: Don't think this would work for local file refs eg `interface.json#defs/xyz` but i think that's okay, generator can't properly handle those yet anyway because they make things really complicated haha

calm/spectral/calm-rules.yaml
```diff
@@ -0,0 +1,66 @@
+extends: spectral:oas
+functions: []  # We'll implement custom functions in the next step
+
+rules:
+  calm-schema-version:
+    description: 'Ensures correct JSON Schema version'
+    message: 'Must use JSON Schema Draft 2020-12'
+    severity: error
+    given: $.['$schema']
```
grahampacker-ms, line 9: I don't think this will raise an error if `$schema` is missing, I think it should be:
```
rules:
  calm-schema-version:
    description: 'Ensures correct JSON Schema version'
    message: 'Must use JSON Schema Draft 2020-12'
    severity: error
    given: $
    then:
      field: $schema 
      function: pattern
      functionOptions:
        match: https://json-schema.org/draft/2020-12/schema
```

calm/spectral/calm-rules.yaml
```diff
@@ -0,0 +1,66 @@
...
+    then:
+      function: pattern
+      functionOptions:
+        match: https://json-schema.org/draft/2020-12/schema
+
+  calm-schema-id-format:
```
grahampacker-ms, line 15: I think this has the same issue as `calm-schema-version`

.github/workflows/validate-calm-schemas.yml
```diff
@@ -0,0 +1,36 @@
+name: Validation of CALM Schemas
```
grahampacker-ms, line 1: do we have any tests for this ruleset?

calm/spectral/calm-rules.yaml
```diff
@@ -0,0 +1,66 @@
+extends: spectral:oas
```
grahampacker-ms, line 1: do we have any tests for this ruleset?

PR #666, opened 2024-12-11
Alternative approach to json lint

jpgough-ms:


rocketstack-matt:
What's this the alternative to?

Issue #645, opened 2024-12-03
calm-cli install failing on Mac OS


djohnston89:
### Steps to Reproduce:
1. fresh install of Node / NPM
2. following docs: https://calm.finos.org/working-with-calm/installation/

### Expected Result:
successful install

### Actual Result:

```shell
% sudo npm install -g @finos/calm-cli
Password:
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated npmlog@5.0.1: This package is no longer supported.
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated are-we-there-yet@2.0.0: This package is no longer supported.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm warn deprecated gauge@3.0.2: This package is no longer supported.
npm error code 1
npm error path /usr/local/lib/node_modules/@finos/calm-cli/node_modules/canvas
npm error command failed
npm error command sh -c node-pre-gyp install --fallback-to-build --update-binary
npm error Failed to execute '/usr/local/bin/node /usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js configure --fallback-to-build --update-binary --module=/usr/local/lib/node_modules/@finos/calm-cli/node_modules/canvas/build/Release/canvas.node --module_name=canvas --module_path=/usr/local/lib/node_modules/@finos/calm-cli/node_modules/canvas/build/Release --napi_version=9 --node_abi_napi=napi --napi_build_version=0 --node_napi_label=node-v127' (1)
npm error node-pre-gyp info it worked if it ends with ok
npm error node-pre-gyp info using node-pre-gyp@1.0.11
npm error node-pre-gyp info using node@22.11.0 | darwin | arm64
npm error (node:41606) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
npm error (Use `node --trace-deprecation ...` to show where the warning was created)
npm error node-pre-gyp http GET https://github.com/Automattic/node-canvas/releases/download/v2.11.2/canvas-v2.11.2-node-v127-darwin-unknown-arm64.tar.gz
[...]

PR #669, opened 2024-12-14
docs: update CLI dev env setup instructions

yt-ms:
- Add details about the version of node, as installation of canvas seems to fail with later versions.
- Update `npx link` to include the folder as it doesn't work without, and clarify that you run these commands from the `cli` folder (if run from the parent folder then it needs to be `npx link cli`).

### DIFF
--- a/cli/DEVELOPER_GUIDE.md
+++ b/cli/DEVELOPER_GUIDE.md
@@ -1,13 +1,17 @@
 # Developing the CALM CLI
 
+## Prerequisites
+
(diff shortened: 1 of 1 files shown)


rocketstack-matt (approved):
You should check why your CLA check isn't passing

rocketstack-matt:
I've merged this in as you're the second person to be having a problem with canvas @djohnston89 [here](https://github.com/finos/architecture-as-code/issues/645) but I don't have this issue using 22.11.0 so think we still need further investigation to get to the bottom of this.

PR #670, opened 2024-12-14
tests: fix problem with running tests when not root

yt-ms:
In `cli.spec.ts`, it was doing an `npm link` even though in the docs is says not to do that.  This means the tests were failing if you weren't running as an account that could modify the global `node_modules`.  I've changed this to use `npx link` instead.


yt-ms:
Closing PR and trying again with a new one to trigger the EasyCLA to reevaluate.

PR #671, opened 2024-12-14
Simple, repeatable dev env set-up with Flox

yt-ms:
[Flox](https://flox.dev) makes development environment set-up simple and repeatable, as it explicitly installs dependencies locally to the project rather than relying on whatever versions of things you may happen to have installed globally.

Setting up the CALM environment on a fresh Ubuntu image with Flox and git already installed:
```bash
git clone https://github.com/finos/architecture-as-code.git
cd architecture-as-code
flox activate
```
And that's it!  Flox will install the right version of Node, `libuuid.so`, do the `npm install`, and even do the initial build and link for you.  And we'll know exactly what versions of everything are being used, which makes debugging environment problems much much easier.

### DIFF
--- a/.flox/.gitignore
+++ b/.flox/.gitignore
@@ -1,0 +1,4 @@
+run/
+cache/
+lib/
+log/
--- a/.flox/env.json
+++ b/.flox/env.json
@@ -1,0 +1,4 @@
+{
+  "name": "calm",
+  "version": 1
+}
--- a/.flox/env/manifest.lock
+++ b/.flox/env/manifest.lock
@@ -1,0 +1,351 @@
+{
+  "lockfile-version": 1,
+  "manifest": {
+    "version": 1,
+    "install": {
+      "gum": {
+        "pkg-path": "gum"
+      },
+      "libuuid": {
+        "pkg-path": "libuuid",
+        "systems": [
+          "aarch64-linux",
+          "x86_64-linux"
+        ]
+      },
+      "nodejs": {
+        "pkg-path": "nodejs"
+      }
+    },
+    "hook": {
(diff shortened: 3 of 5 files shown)


rocketstack-matt:
@yt-ms how does this differ to using dev containers? What's the additional benefit?

yt-ms:
> @yt-ms how does this differ to using dev containers? What's the additional benefit?

The benefit is you don't need dev containers!  More seriously, it provides absolute certainty around the build dependencies, and reduces the amount of toil in setting up and maintaining the dev environment.  An example of the dependencies is that when Canvas was added to the project, I'm sure everyone was fine because they happened to be on an older version of Node, but when I came to set up from scratch then I lost a couple of hours trying to figure out why the dev instructions didn't work - it turns out because on my machine it was on a later version of Node.  This kind of thing just doesn't happen with Flox.  I'm also very comfortable blowing away my dev env and reconstructing regularly as I don't need to remember a whole sequence of installs and hacks to get everything working properly.

And if you want a dev container, then you can still run Flox inside that - there's even a minimal (Nix-based) OCI image at ghcr.io/flox/flox, or you can get Flox to generate you the OCI image from your Flox environment.

yt-ms:
Just trying to nudge the EasyCLA bot to reevaluate...

Thels:
New to Flox - at first this felt like a pretty large wrapper around what was three commands previously ( install / build / link ) but I do see the value if it also manages the build tool versioning for us.

One thing to note is that being able to do a root level `npm install/build/link` is due to `npm workspaces` - **which currently doesn't contain the docs module**. Should we add into the `hook` to potentially also cd to that directory and build it separately too? This could then be removed when we migrate docs to also be a part of the workspace.

Thels (approved):

Issue #693, opened 2024-12-19
Terminology Change - Architecture replaces instantiation


grahampacker-ms:
### Description of Problem:
The term `instantiation` is a high level concept and is difficult to conceptualise to people new to CALM. The term `architecture` accurately describes the document, and is more understandable.

### Potential Solutions:
Update any references of `instantiation` to `architecture` and update the CLI.

PR #694, opened 2024-12-19
issue-#693 changing terminology to architecture

grahampacker-ms:
Hopefully this captures all instances of the previous terminology


rocketstack-matt:
Do we want to rename `shared\src\commands\generate\components\instantiate.ts` and associated functions to include along the lines of instantiateArchitecture or we're happy it isn't confusing in the case of the component (as it's actually doing instantiation)?

rocketstack-matt (approved):
Other than my earlier comment looks good, I don't see any other remaining uses of the word instantiate or instantiation other that in the actually instantiation logic. 

I'm approving so you're welcome to merge if you're happy this piece of logic wasn't intended to be renamed, but won't merge in case you do want any further changes.

grahampacker-ms:
Thanks @rocketstack-matt. I've kept the scope of this to renaming the resulting document to architecture. I think the only cases left are where we're actually instantiating the architecture from the pattern, which I'm fine with.

Issue #701, opened 2024-12-24
Explore Options for CALM Schema to TypeScript Generation - Library Evaluation


LeighFinegold:
### Description of Problem:

There are many use cases for the CALM (Common Architecture Language Model) language, such as generating architecture implementation files (#444), documentation (#478), and more. Since the CALM specification is defined using JSON Schema, it is often necessary to convert the CALM model into strongly typed representations, making it easier to manipulate and use for generating code outputs or documentation.

Currently, there is no clear or standardised approach for converting CALM schemas (JSON Schema) into usable TypeScript types. It should consider the potential for extension. 

### Potential Solutions:

1. **Define our own internal model manually**: We could parse the CALM schema into a custom internal model, but this would require significant time and effort to implement and maintain.

2. **Use an existing JSON Schema to TypeScript generator**: We could leverage an established npm library (such as [json-schema-to-typescript](https://www.npmjs.com/package/[json-schema-to-typescript) ) to automatically generate TypeScript types from the CALM schemas. This approach would save time and reduce the complexity of creating a custom solution, but it may not always be a perfect fit for CALM’s unique and extension needs.

3. **Write our own TypeScript generator**: We could write our own custom generator (AST) tailored specifically to the CALM schema. While this would give us the most control, it would also require significant effort to develop and maintain, as well as deal with any edge cases specific to CALM.

### Focus of this Issue
This issue will focus on Solution 2. Will explore [json-schema-to-typescript](https://www.npmjs.com/package/[json-schema-to-typescript) which seems the most widely adopted npm library for JSON Schema to TypeScript conversion, as it is the most straightforward and efficient approach to meet our needs. We want to evaluate whether this approach will work effectively with CALM schemas and whether there are any limitations or modifications needed to integrate it fully into our toolchain.

LeighFinegold:
Linked to this issue is a PR containing a basic setup of the library to generate models 
- https://github.com/finos/architecture-as-code/pull/702

The library seems a little buggy when it comes to generating type definitions for defs that are not referenced via any schemas.

However, it has been really useful to do as I think it will encourage a conversation around
1. Do we have the schema types defined well enough? Perhaps core.json should contain node, relationship, , but there should be seperate architecture and patterns schemas refererencing those. 
2. The library struggles with pattern field concepts in json schema
[...]

PR #702, opened 2024-12-24
DO NOT MERGE: POC using json-schema-to-typescript library

LeighFinegold:
This PR contains the early proof-of-concept (PoC) work with model generation #701 

Lessons learned will be attached to issue rather than PR.

### DIFF
--- a/model-gen/build/calm/2024-12/core.d.ts
+++ b/model-gen/build/calm/2024-12/core.d.ts
@@ -68,27 +68,7 @@ export interface ControlDetail {
 export interface Relationship {
   "unique-id": string;
   description?: string;
-  "relationship-type":
-    | {
-        kind: "deployed-in";
-        details: DeployedInType;
-        [k: string]: unknown;
-      }
-    | {
-        kind: "composed-of";
-        details: ComposedOfType;
-        [k: string]: unknown;
-      }
-    | {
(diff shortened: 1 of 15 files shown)


model-gen/build/calm/2024-12/core.d.ts
```diff
@@ -0,0 +1,162 @@
+export type Metadata = {
```
LeighFinegold, line 1: These file souldn't be checked in (but doing so, for folks to review and see output)

model-gen/build/calm/2024-12/flow.d.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
LeighFinegold, line 1: The library seems to have a bug whereby just defs alone in a file will not generate any definitions. It will generate definitions via a schema which is why core.d.ts has flow definition generated.

model-gen/build/calm/2024-12/core.d.ts
```diff
@@ -0,0 +1,162 @@
...
+  [k: string]: unknown;
+}[];
+export type Metadata1 = {
```
LeighFinegold, line 4: Should MetaData be a general concept which is leveraged by flow.json and core.json to prevent this

PR #704, opened 2024-12-27
maint: bump recommended Node version to 21 and auto-install Java dependencies

yt-ms:
Node v21 is also compatible with `canvas`, so bumping the recommended version and updating the Flox environment to match.


jpgough-ms (approved):

rocketstack-matt (approved):
Have tested the setup with this node version with Flox and without and works in both cases

Issue #710, opened 2024-12-31
CalmHub: Modify Tests to Remove @Ignored Test


jpgough-ms:
### Steps to Reproduce:

Running all tests causes issues with tests failing. 

Modification to tests, as identified by @grahampacker-ms and noted in this PR: https://github.com/jpgough/calm-hub/pull/3/files

grahampacker-ms:
@jpgough-ms I'm reassigning to myself and will move the PR to this repo

Issue #714, opened 2024-12-31
CalmHub: Add a GitHub Action to Trigger Build


jpgough-ms:
### Description of Problem:

Run a full build on any PR to run both the `unit tests` and `integration-tests` of the CalmHub project. This will need to use the TestContainers action to run both styles of test (once the integration tests are expected, Docker must be available or the tests will not work).

### Potential Solutions:

Use a GitHub action with docker required for TestContainers:

```
# Step 3: Set up Docker (required for Testcontainers)
      - name: Set up Docker
        uses: docker/setup-buildx-action@v2
```

The run command for both types of test is (from the calm-hub directory):

```shell
../mvnw -P integration clean verify
```

Verify both test plugins produce test results:

* `maven-surefire-plugin`
* `maven-failsafe-plugin`

At the time of writing there are 88 unit tests and 13 integration tests

PR #726, opened 2025-01-01
Fix tests to allow all tests to run as part of suite (#710)

grahampacker-ms:


jpgough-ms:
@grahampacker-ms before we merge this, I’d like to setup the config for GH Actions #714 

I can pick this up tomorrow and review the above.

jpgough-ms (approved):

PR #734, opened 2025-01-02
Remove demo data from mongo init file (#712)

### LINKED ISSUE #712: CalmHub: Init Mongo Script
### Steps to Reproduce:

Init Mongo script is not correct (what is in the integration tests is) for setting up counters. The script was put together to have a few examples.

### Fix

Update the init mongo script to not add any architectures or patterns, but only setup the latest schema.

grahampacker-ms:
Fixes #712


calm-hub/mongo/init-mongo.js
```diff
@@ -1192,280 +757,5 @@ db.schemas.insertMany([               // Insert initial documents into the schem
...
-        ]
-    },
-    {
-        namespace: "custom",
-        patterns: [
-        ]
-    }
-]);
-
-db.architectures.insertMany([
```
jpgough-ms, line 1462: Does `db.architectures` and `db.patterns` still need to be inserted as empty? Not sure what the code assumptions are.

grahampacker-ms: I tested inserts and they worked fine

grahampacker-ms: I tested GETs on a new DB and it errors. I'll add the empty inserts to initialise the collection

grahampacker-ms: @jpgough-ms fixed, I've put some code protections so an empty collection doesn't cause null pointers. I did try to initialize the collection as empty, but it still cause NPEs

jpgough-ms (approved):

Issue #169, opened 2024-04-22
Spectral ruleset to check that referenced interfaces are defined on the specific node


willosborne:
### Description of Problem:
Currently spectral rules check that referenced interfaces are defined.
It does not check that they're defined on the specified node - so this could potentially cause some confusion.

### Potential Solutions:
Either enhance the existing rule to check the node that owns that interface, or make a new rule that performs this check.

willosborne:
Another rule - check for erroneous ref values. Everything should be $ref and 'ref' should be an error.

jpgough-ms:
@willosborne was this implemented?

willosborne:
@jpgough-ms no, this is still unimplemented

willosborne:
Picking this up

willosborne:
This is now done.

PR #743, opened 2025-01-02
Add Spectral rules to validate that interfaces are defined on the specific node referenced (169)

willosborne:
Addresses #169 

Adds a new rule; technically this overlaps with the 'not defined anywhere' rule which I could probably cut.
But 'not defined anywhere' and 'not defined on the right node' are different data points in my mind.
Could merge them into one, what do people think?


shared/src/commands/generate/components/property.spec.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable  @typescript-eslint/no-explicit-any */
```
willosborne, line 1: These exclusions aren't actually used, so lint-fix automatically cleaned them up.

grahampacker-ms (dismissed):

willosborne:
Would be good to get input from @Thels here before I merge - I needed to npm install rollup for the spectral CI job to work, and I also needed to add link to the dev dependencies to prevent an error in the CI jobs. Not sure how that was merged in - I was seeing the issues on CI too so not just my machine. 

RE the CI jobs, I think we should probably just replace the Spectral CI job with running the CLI directly anyway - make things more consistent and use our own tools.

.github/workflows/validate-spectral.yml
```diff
@@ -39,6 +39,9 @@ jobs:
 
       - name: Install Spectral-CLI
         run: npm install @stoplight/spectral-cli
+        
+      - name: Install Spectral-CLI
```
Thels, line 43: Can we hop on a call to go over this? I worry that other dependency changes have made this workflow not work anymore ( when it was previously ) - which could mean that the conditions in which this workflow is triggered could be incorrect.

.github/workflows/validate-spectral.yml
```diff
@@ -39,6 +39,9 @@ jobs:
 
       - name: Install Spectral-CLI
```
Thels, line 40:
```suggestion
      - name: Install Dependencies to execute Spectral Rules
```

.github/workflows/validate-spectral.yml
```diff
@@ -39,6 +39,9 @@ jobs:
...
         run: npm install @stoplight/spectral-cli
```
Thels, line 41:
```suggestion
        run: npm install @stoplight/spectral-cli rollup
```

We can do both in the single install - no need for an additional step.

.github/workflows/validate-spectral.yml
willosborne: Discussed offline, it's working fine but I misunderstood some stuff.

PR #770, opened 2025-01-07
[Early Early Draft] WIP - CLI Server 

Thels:


cli/src/routes/validation-route.ts
```diff
@@ -0,0 +1,19 @@
+// import { CALM_META_SCHEMA_DIRECTORY, getFormattedOutput, validate } from '@finos/calm-shared';
+import { Router } from 'express';
+
+export const validationRouter: Router = Router();
+
+
+validationRouter.get('/example', (req, res) => {
```
Thels, line 7: Just me playing around with an example - as I was having major issues getting the end-points outside of basic GETS to be hittable during end to end tests.

cli/src/cli.spec.ts
```diff
@@ -161,6 +163,40 @@ describe('CLI Integration Tests', () => {
...
+        const serverProcess = exec(serverCommand, async (_error, _stdout, _stderr) => {
+            const response = await axios.get('http://localhost:3000/api/validate/health');
+            expect(response.status).toBe(200);
+            done();
+        });
+
+        serverProcess.kill();
+
+
+    }, millisPerSecond * 30);
```
Thels, line 180: This test passed, however the one below wouldn't. 

I think if we only test that the express server ( or whatever framework we use ) behaves then maybe we just want to validate that the server itself has come up when calling `calm server`

Issue #783, opened 2025-01-10
PR Clarity - Labelling


Thels:
### Description of Problem:
It's not very clear when a PR is raised by either a human or renovate what section of the code that it's impacting.

Sometimes titles can be indicative, but it would nice if we labelled each PR accordingly if it changes one/more of the packages.

### Potential Solutions:

We could update Renovate to ensure that it has the package name in the title of the PR.
We could introduce a GitHub action that uses the auto-labler to tag things based on the major folder that is changed. Potentially only supporting major packages like;

- cli
- shared
- calmhub
- translator

This could be extended in the future to handle other modules that are added.

PR #784, opened 2025-01-10
#783 Pr Labelling

Thels:
Ensure that changes being made in PR's are labelled based on the module being modified.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
-CLI:
+cli:
   - "cli/**"
 shared:
   - "shared/**"
@@ -6,7 +6,7 @@ docs:
   - "docs/**"
 calm-hub:
   - "calm-hub/**"
-CALM:
+calm:
   - "calm/**"
 config:
   - "pom.xml"
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,14 @@
+CLI:
+  - "cli/**"
+shared:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
yt-ms, line 1: I think the normal convention is lower-case for labels.  Should we standardise on that (e.g. CLI->cli, CALM->calm) and migrate the existing labels (Generator, Refactor, Roadmap:*, Validator, Visualiser) over?

.github/workflows/pr-labelling.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  label:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Label PR based on changes
+      uses: actions/labeler@v4
```
yt-ms, line 16: Why not go straight to v5?

Thels: It's a fair point - was just a miss. Will update.

.github/labeler.yml
Thels: I'll be transparent here and didn't realise Github allowed such easy editing of labels.

Will modify to be lower-case, have moved the labels on the project itself to be in-line.

jpgough-ms (changes requested):
Changes agreed on PR

jpgough-ms (approved):

PR #793, opened 2025-01-13
Adding support to validate an architecture without relying on a pattern

lbulanti-ms:
In this PR I added the support to validate an architecture without the need of a pattern. The architecture validation for now is based on the spectral rules that already exist in the project.


shared/src/commands/validate/validate.ts
```diff
@@ -267,71 +267,105 @@ export async function validate(
...
+            return await validatePatternOnly(jsonSchemaLocation, metaSchemaPath, debug);
+        } else if (jsonSchemaArchitectureLocation) {
+            return await validateArchitectureOnly(jsonSchemaArchitectureLocation);
+        } else {
+            logger.debug('You must provide at least an architecture or a pattern');
+            throw new Error('You must provide at least an architecture or a pattern');
         }
+    } catch (error) {
+        logger.error('An error occured:', error);
+        process.exit(1);
```
aidanm3341, line 283: I wonder if at some point it would make sense to not exit the process here, but instead in the CLI component instead which calls this code. I'm thinking from a reuse perspective, there could potentially be an app which doesn't want to exit when an error occurs, but instead just displays it

shared/src/commands/validate/validate.ts
```diff
@@ -267,71 +267,105 @@ export async function validate(
...
- * This essentially tries to compile the pattern, and returns the errors thrown if it fails.
+ * This essentially runs the spectral validations and tries to compile the pattern.
  * 
- * @param spectralValidationResults The results from running Spectral on the pattern.
- * @param patternSchema The pattern as a JS object, parsed from the file.
- * @param ajv The AJV instance to compile with.
- * @param failOnWarnings Whether or not to treat a warning as a failure in the validation process.
+ * @param jsonSchemaLocation the location of the patterns JSON Schema to validate.
+ * @param metaSchemaPath the path of the meta schemas to use for ajv.
+ * @param debug the flag to enable debug logging.
```
aidanm3341, line 332: small thing - might be good to have a dash after the param name, I was ever so briefly confused because I thought this was telling me to 'debug the flag' 😆

aidanm3341 (approved):

grahampacker-ms (approved):

shared/src/commands/validate/validate.ts
willosborne: Yes. This is going to be important for the CLI server, I would imagine. We want to just return the validation output back to the calling module

willosborne (approved):

aidanm3341 (approved):

Issue #716, opened 2024-12-31
CalmHub: Store Architecture Decision Records (ADR)


jpgough-ms:
### Description of Problem:

Whilst centralising the store of patterns and architectures, it also makes sense to centralise the store of architecture decision records that accompany these assets. 

### Potential Solutions:

* Pick a JSON structure to store ADRs
* Create a RESTful API for managing ADRs and changing their status

jpgough-ms:
Proposal for what the API could look like:

```
# Retrieve all ADRs in a namespace
GET /calm/namespaces/{namespace}/adrs
- Fetch all ADRs within a specific namespace.

# Create a new ADR in draft state
POST /calm/namespaces/{namespace}/adrs
- Create a new ADR in the draft state. Returns an ADR id.

# Update ADR state to proposed
PUT /calm/namespaces/{namespace}/adrs/{id}/propose
- Transition the ADR to the "Propose" state.

# Update ADR state to accepted
PUT /calm/namespaces/{namespace}/adrs/{id}/accept
- Transition the ADR to the "Accepted" state.

# Mark ADR as superseded
PUT /calm/namespaces/{namespace}/adrs/{id}/supersede
- Mark an ADR as superseded by another ADR. Accepts the id of the superseding ADR in the request body.

# Mark ADR as outdated
PUT /calm/namespaces/{namespace}/adrs/{id}/outdate
- Mark an ADR as outdated.

# Reject ADR
PUT /calm/namespaces/{namespace}/adrs/{id}/reject
- Transition the ADR to the "Rejected" state.
```

jpgough-ms:
I'd propose that the `body` content of these is the MADR specification for ADRs in JSON: https://adr.github.io/madr/

yt-ms:
What is the difference between outdate and supersede?

yt-ms:
> I'd propose that the `body` content of these is the MADR specification for ADRs in JSON: https://adr.github.io/madr/

JSON to make it schema-verifiable?  Markdown would be more human-friendly.

Also need to think about extensibility, as an organisation may want to add additional metadata, and potentially additional sections.

yt-ms:
Presumably you'll also need `GET /calm/namespaces/{namespace}/adrs/{id}`...

yt-ms:
What metadata would we expect to be returned for all ADRs?  E.g. created by, last updated data, etc.

yt-ms:
Do we need to support querying the version history?

[...]

PR #798, opened 2025-01-15
Add ADR support to CalmHub

grahampacker-ms:
Addresses #716


calm-hub/src/main/java/org/finos/calm/store/mongo/MongoAdrStore.java
```diff
@@ -0,0 +1,224 @@
...
+        Document namespaceDocument = adrCollection.find(Filters.eq("namespace", namespace)).first();
+
+        //protects from an unpopulated mongo collection
+        if(namespaceDocument == null || namespaceDocument.isEmpty()) {
+            return List.of();
+        }
+
+        List<Document> adrs = namespaceDocument.getList("adrs", Document.class);
+
+        return adrs.stream().map(adr -> adr.getInteger("adrId")).toList();
```
jpgough-ms, line 67: Just use `id`?

calm-hub/src/test/java/org/finos/calm/resources/TestAdrResourceShould.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+
+    @ParameterizedTest
+    @MethodSource("provideParametersForCreateAdrTests")
+    void respond_correctly_when_creating_adr(String namespace, Throwable exceptionToThrow, int expectedStatusCode, Matcher<?> locationHeader) throws NamespaceNotFoundException, JsonProcessingException {
+        if (exceptionToThrow != null) {
+            when(mockAdrStore.createAdrForNamespace(any(Adr.class))).thenThrow(exceptionToThrow);
+        } else {
+            Adr adr = AdrBuilder.builder()
```
jpgough-ms, line 90: Can we just use one ADR for all these tests and compress the file?

jpgough-ms (dismissed):

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoAdrStore.java
grahampacker-ms: this was to match the other objects ("architectureId", "patternId" etc.) for consitency

calm-hub/src/main/java/org/finos/calm/domain/NewAdr.java
```diff
@@ -0,0 +1,18 @@
+package org.finos.calm.domain;
```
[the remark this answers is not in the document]
grahampacker-ms: I struggled with this, naming being the hardest problem in software engineering 😆 

`Adr` contains the metadata and the ADR content
`AdrContent` would be the actual ADR document we would render when we have a UI
`NewAdr` is the first, or updated, `AdrContent`, I made it distinct from `AdrContent` as there are fields (namely `status`, `creationDateTime` and `updateDateTime` that are not modifiable by the requestor. I used a different type so they can't be set on the request but will be returned in the responses

calm-hub/src/main/java/org/finos/calm/domain/Adr.java
```diff
@@ -0,0 +1,13 @@
+package org.finos.calm.domain;
+
+import io.soabase.recordbuilder.core.RecordBuilder;
+
+/**
+ * Represents an ADR and the associated namespace, id, and revision.
+ * The ADR is represented as a String in JSON format.
+ */
+@RecordBuilder.Options(enableWither = false)
```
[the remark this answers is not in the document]
Thels: Wouldn't any builder implementation of a record just require setting the same attributes in a mutable object and then calling the constructor for us **anyway**? 

+1 to @jpgough-ms 's comment - I don't think this is required.

calm-hub/src/main/java/org/finos/calm/domain/AdrContent.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer;
+import io.soabase.recordbuilder.core.RecordBuilder;
+
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Objects;
+
+@RecordBuilder.Options(enableWither = false)
+@RecordBuilder
+public record AdrContent(
```
Thels, line 16: This is a HUGE record - is there any other parts of the data structure that we could group together? 

Such as `AdrMetaData` etc

calm-hub/src/main/java/org/finos/calm/domain/AdrStatus.java
```diff
@@ -0,0 +1,10 @@
+package org.finos.calm.domain;
+
+public enum AdrStatus {
+    DRAFT,
```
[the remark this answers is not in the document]
Thels: Can I ask why? I thought the justification for them being uppercase was the fact that they're constants and the standard is for constants to be capitalized. 

Supporting lowercase is fair, just curious as to why you prefer the lowercase options.

calm-hub/src/main/java/org/finos/calm/domain/AdrContent.java
```diff
@@ -0,0 +1,55 @@
...
+                .decisionOutcome(newAdr.decisionOutcome())
+                .links(newAdr.links());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if(this == o) return true;
+        if(o == null || getClass() != o.getClass()) return false;
+        AdrContent that = (AdrContent) o;
+        return Objects.equals(title, that.title) && status == that.status  && Objects.equals(contextAndProblemStatement, that.contextAndProblemStatement) && Objects.equals(decisionDrivers, that.decisionDrivers) && Objects.equals(consideredOptions, that.consideredOptions) && Objects.equals(decisionOutcome, that.decisionOutcome) && Objects.equals(links, that.links);
```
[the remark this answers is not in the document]
Thels: Yeah some parts are randomly double spaced?

calm-hub/src/main/java/org/finos/calm/store/AdrStore.java
```diff
@@ -0,0 +1,22 @@
...
+import org.finos.calm.domain.exception.AdrNotFoundException;
+import org.finos.calm.domain.exception.AdrPersistenceError;
+import org.finos.calm.domain.exception.AdrRevisionNotFoundException;
+import org.finos.calm.domain.exception.NamespaceNotFoundException;
+
+import java.util.List;
+
+public interface AdrStore {
+
+    List<Integer> getAdrsForNamespace(String namespace) throws NamespaceNotFoundException;
```
Thels, line 15: Not sure of the naming used in the rest of the project ( and would appreciate @jpgough-ms 's feedback on this ) but I always tend to avoid repeating the name of the object being fetched when the caller **knows** they're using an ADR store.

For example;

`getAdrsForNamespace` becomes `getForNamespace`
`createAdrForNamespace` becomes `createForNameSpace`

It also means that these `stores` become something we can put behind a generic interface, as I'm assuming a lot of our objects will support `CRUD for namespace`

calm-hub/src/test/java/org/finos/calm/resources/TestAdrResourceShould.java
Thels: It would be nice to have the JSON representation of the ADR in the test resources for when folks come in to easier understand the structure.

calm-hub/src/main/java/org/finos/calm/domain/AdrContent.java
grahampacker-ms: This is the MADR format

calm-hub/src/main/java/org/finos/calm/domain/Adr.java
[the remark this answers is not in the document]
grahampacker-ms: The benefits of a builder as I see it are:
a) easy of changing the type over time. e.g. adding a new field doesn't cause all the constructors to have syntax errors
b) you can pass in an existing object, and use that as the basis of a new object
c) immutable objects, in my opinion, produce less bugs than mutable ones

grahampacker-ms:
Made some changes based on review comments. Any additional changes I think should be done as follow up changes.

calm-hub/src/main/java/org/finos/calm/domain/adr/Status.java
```diff
@@ -0,0 +1,10 @@
+package org.finos.calm.domain.adr;
+
+public enum Status {
```
lbulanti-ms, line 3: is there any specific reason to have the members of this enum lowercase instead of uppercase?

grahampacker-ms: Jim asked for it 😄

calm-hub/src/main/java/org/finos/calm/store/AdrStore.java
jpgough-ms: Sounds like a decent refactor, one to do across the board. This is one of the code complete pests 😄, especially when generated from tests/mocks

jpgough-ms (approved):

Issue #314, opened 2024-07-26
Log a warning when wrong CALM core schemas are being loaded


willosborne:
### Description of Problem:
The CALM CLI bundles the meta-schemas it was built against alongside the app. 
When a user tries to generate/validate a pattern, these schemas will be made available to the command.

However, if the pattern in use has an outdated schema URL, it will either try to load that schema or report it as missing.
For example, a pattern pointing at an older version of the spec, a version hosted on the old finos-labs repo, etc.

This is really confusing for users and indicates something wrong with how the spec is loaded, rather than just their pattern having the wrong $schema version.

### Potential Solutions:
Keep track of the bundled schemas, and log an ERROR warning very clearly if a user is trying to run commands against a different version (and has not manually loaded these via the --schemaDirectory flag.)

This will involve somehow detecting is a pattern URL is the CALM pattern; perhaps if it's got the same path but has the wrong date in the URL.

Thels:
From office hours - potentially notification to the user to upgrade their pattern if it's outdated.

willosborne:
Going to take a look at this. I think this boils down to two main parts:
- Determining which version is the 'bundled' version of CALM
- Detecting when the schema being loaded is a core schema, and which version it is

The first part is the bit that has me scratching my head; I guess we could just hard-code this somewhere since we have a task that copies the schemas in.

willosborne:
My initial try at this (#800) was a bit too specific and I closed the PR.
Instead my ongoing refactor for the SchemaDirectory class will solve this by simply logging a warning when _any_ schema is not loaded. See #838

Issue #478, opened 2024-10-16
Creating an Architecture Documentation Website


LeighFinegold:
# Feature Request

## Description of Problem:
As of 2024-10, CALM provides a highly flexible framework for capturing detailed technical architectures through nodes,  relationships, flows and associated control requirements. This has enabled us to effectively map and apply constraints onto system architectures and reusable patterns. We have already demonstrated the ability to transition from these architectural constraints to infrastructure as code, specifically in k8s environments.

However, as we scale to larger and more complex system architectures, especially when incorporating additional stakeholders and cross-functional teams, there is a growing need for a more dynamic and accessible format for architecture documentation. The current CALM specification, while robust, primarily exists as a static artifact that lacks interactivity and user-friendly navigation, which is essential for cross-team collaboration, onboarding, and maintaining long-term architectural knowledge.

This is particularly critical as developers, architects, and operations teams require more intuitive access to architecture details, business logic, and relationships between systems. Without a centralized, interactive view, it can be challenging to effectively communicate architectural changes, technical flows, and associated control requirements to both new and existing team members.

## Potential Solutions:
Proposing introducing a feature to generate a document based website (using library like https://docusaurus.io/) from the CALM specification. This solution would allow teams to automatically generate a navigable, interactive site that documents the system architecture in an accessible format, making it easier to understand, explore, and communicate complex architectures across various teams and stakeholders.

### The website should capture the following key aspects of the CALM specification:

- **High-Level Business Flows**: Clear documentation and visualization of business processes supported by the system, including how different flows interact with the architecture. This enables stakeholders to easily understand the relationship between business goals and the technical infrastructure.
  
- **Architecture Decision Records (ADRs)**: A section that showcases key architectural decisions and their impact on the system. This allows users to track the rationale behind decisions and view how they affect nodes, relationships, flows, and patterns.

- **Node Focus**: The ability to drill down into individual nodes within a business flow or system. This would provide details about how a specific node functions, its role in the flow, and how it interacts with surrounding components.

- **System Views**: Comprehensive views of the entire system architecture, including nodes, relationships, and control requirements, allowing users to zoom in and out of technical details based on their needs.

- **Interactive Visualization**: Allowing users to explore business flows and system components dynamically, toggling between high-level and detailed views.

- **Data Dictionary**: With https://github.com/finos/architecture-as-code/issues/462 be able to clearly show the data assets linke to the high-level/detailed views.

## Proposed Solution

`calm docify calm.json`

[...]

PR #800, opened 2025-01-16
Log warning when validating a spec against the wrong core schema versions

willosborne:
Initial stab at #314 

This looks at the bundled files and then intercepts any load to look at the URL. It's quite crude, but wanted to get an initial PR up to see people's thoughts.

### DIFF
--- a/shared/src/commands/validate/validate.ts
+++ b/shared/src/commands/validate/validate.ts
@@ -11,6 +11,7 @@ import { ValidationOutput, ValidationOutcome } from './validation.output.js';
 import { SpectralResult } from './spectral.result.js';
 import createJUnitReport from './output-formats/junit-output.js';
 import prettyFormat from './output-formats/pretty-output';
+import { checkCoreSchemaVersion, getBundledSchemaVersion } from '../bundled-schemas';
 
 let logger: winston.Logger; // defined later at startup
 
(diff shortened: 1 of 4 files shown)


shared/src/commands/validate/validate.ts
```diff
@@ -64,9 +65,12 @@ export function formatOutput(
 
 function buildAjv2020(debug: boolean): Ajv2020 {
     const strictType = debug ? 'log' : false;
+    const bundledSchemaVersion = getBundledSchemaVersion(debug);
     return new Ajv2020({
         strict: strictType, allErrors: true, loadSchema: async (uri) => {
             try {
+                // validate schema version
+                checkCoreSchemaVersion(uri, bundledSchemaVersion, debug);
```
LeighFinegold, line 73: How would this work if organisations want to extend the core calm schema but add their own spectral rules for their extensions?

willosborne: So you mean if there is effectively a second set of core schemas - the extensions offered by an org - and I guess they would need versioning too?

I guess what I have here is a pretty narrow definition of a 'core schema'. 
The reason we embarked on this was that Jim was doing a validate and getting a ton of weird errors because he was using something still pointing at the first release.

What would you say to instead having it just log a warning if you do validation against a spec not found/registered in the schema directory, instead of all this weirdness with 'core schemas'?

LeighFinegold: That would make a lot more sense to me. One scenario I was also concerned about (although more a future roadmap end state ) is if I am referencing someone else's CALM spec for node information do I care if its not on the same version of core version anyways. I'm not sure we would want to mandate it, but also not sure what the implications would be on our current tooling. It's kind of why on this [comment](https://github.com/finos/architecture-as-code/issues/478#issuecomment-2538465256) I was indicating that maybe the tooling goes via a defined model but then that also implies every piece of tooling is extendable.

willosborne:
I'm going to close this PR for now. This work will be immediately superseded by the incoming refactors around SchemaDirectory (see #830 #831)

Once that's done this will be much easier

PR #801, opened 2025-01-17
Improve Visualizer Layout

aidanm3341:
Updated the layout engine to try something new, I think its much better.
Also in the latest CALM draft relationships use 'unique-id' not 'uniqueId', so I think the types were just outdated and I have updated.

Screenshot 2025-01-17 at 15 20 21
Screenshot 2025-01-17 at 15 57 34

still not perfect, but an improved layout

### DIFF
--- a/calm-visualizer/src/components/drawer/Drawer.tsx
+++ b/calm-visualizer/src/components/drawer/Drawer.tsx
@@ -139,7 +139,7 @@ function Drawer({ calmInstance, title, isConDescActive, isNodeDescActive }: Draw
                 if (isInteracts(relationship)) {
                     return {
                         data: {
-                            id: relationship['unique-id'],
+                            id: relationship.uniqueId,
                             label: relationship.description || '',
                             source: relationship['relationship-type'].interacts.actor,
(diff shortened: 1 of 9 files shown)


oliviajanejohns (approved):

calm-visualizer/src/components/drawer/Drawer.tsx
```diff
@@ -139,7 +139,7 @@ function Drawer({ calmInstance, title, isConDescActive, isNodeDescActive }: Draw
                 if (isInteracts(relationship)) {
                     return {
                         data: {
-                            id: relationship['unique-id'],
+                            id: relationship["unique-id"],
```
willosborne, line 142: is this not conflicting with the lint rules? or does it have to be double quotes for jsx

willosborne (approved):

calm-visualizer/src/components/drawer/Drawer.tsx
aidanm3341: apparently not, the lint passes... but you make a good point, this kinda flew under my radar. I'll take a look at making this consistent an an upcoming PR.

PR #803, opened 2025-01-17
Highlight the small nodes on selection

aamanrebello:
I noticed that the small nodes, which are white in colour, don't get highlighted when you click on them, while the edges do. It would be nice to see which node corresponds to the sidebar which opens, hence I made this PR where the selected white nodes are highlighted.

The reason highlighting was not happening earlier was because these nodes were all being given a static html template for a label. I tweaked this a little to give a new node label template if the node has the `:selected` class. The result looks like this:

image

Note that:
- If we close the sidebar, the node on the graph does not get deselected.
- If we deselect on the graph (by clicking outside it), the sidebar does not get closed

This behaviour is easy enough to change, but I thought I'd keep things simple in this PR e.g. do people agree with the selection colour?

### DIFF
--- a/calm-visualizer/src/components/cytoscape-renderer/cytoscape.css
+++ b/calm-visualizer/src/components/cytoscape-renderer/cytoscape.css
@@ -8,6 +8,10 @@
     background-color: white;
 }
 
+.selected-node {
+    background-color: #CCE1F9;
+}
+
 :focus {
     border: 1px solid;
     border-color: blue;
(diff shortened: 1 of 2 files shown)


calm-visualizer/src/components/cytoscape-renderer/cytoscape.css
```diff
@@ -8,6 +8,10 @@
     background-color: white;
 }
 
+.selected-node {
+    background-color: aliceblue;
```
aidanm3341, line 12: from looking at the screenshot, I wonder if a more bold blue would be even better. It's maybe a little subtle right now

aamanrebello: How about this?


it matches the colour when you select one of the larger nodes (hopefully anyway, I used a colour picker):


I think I'll push the change for now, can redo the commit if something different is wanted.

aidanm3341: looks great

aidanm3341 (approved):

Issue #817, opened 2025-01-23
Constraining Detailed Architecture Usage in Patterns


jpgough-ms:
### Description of Problem:

`detailed-architecture` provides a "drill down" feature in CALM. A `node` can provide a detailed architecture link, which corresponds to another CALM architecture with more details of a node. 

When using the `detailed-architecture` feature in patterns, it would be helpful to constrain the `detailed-architecture` to a specific pattern. 

### Potential Solutions:

One solution is to provide an optional `detailed-pattern` or `detailed-architecture-pattern`, this would constrain the downstream architecture and enable the architect to define any input and output `nodes` or `relationships` relevant to the pattern.

It would also enable the downstream pattern to constrain the downstream architecture where necessary and have that also conform to a pattern.

jpgough-ms:
Changing `node` to have something like

```json
"details": {
          "type": "object",
            "properties": {
                "detailed-architecture": {
                  "type": "string"
                },
                "detailed-pattern": {
                  "type": "string"
                }
            }
        }
```

Could work really well with patterns. 

* It enables a pattern author to say you **must** provide a detailed-architecture where on is needed. 
* It can be constrained to point at a specific detailed architecture (for known systems that have a well established set of components and controls)
* It can be used to say you **must** provide a detailed architecture and it is constrained by this pattern

cc/ @Budlee @rocketstack-matt

rocketstack-matt:
Would the name `reference-pattern` or `required-pattern` be more appropriate in terms of intent?

Budlee:
The required-pattern is a nice add here and I believe should resolve what we require

Budlee:
@rocketstack-matt is this going to be in the new draft?

jpgough-ms:
@Budlee the plan is to get it into version 1. 

There is a bit of work left open on this issue to validate how it behaves in the CLI. I've assigned to myself to test this and will either close the issue or create a new issue for CLI support.

rocketstack-matt:
Still open as we need to validate if this works correctly in the cli

YoofiTT96:
So say you specify detailed architecture that links to another CALM spec, what would denote the entry point and exit point

If say Node A connects to Node B and node B has a detailed architecture, it follows that there is a specific node that Node A is connecting to in Node B.

Likewise if Node B connects to C, C would be receiving a connection from a node from the detailed architecture in B

markscott-ms:
[...]

PR #821, opened 2025-01-23
Proposal for #817 adding pattern to the previously declared detailed-architecture

jpgough-ms:


willosborne (approved):

jpgough-ms:
@rocketstack-matt you're absolutely right. If it's OK, I will merge this PR and before closing the issue I will link this work that needs to be done. 

For our #832 next week, we should have this discussion.

rocketstack-matt:
Sounds good, thanks.

PR #829, opened 2025-01-23
remove visualize command and dependencies. Fixes #827

### LINKED ISSUE #827: Deco CLI Visualize Command
## Request

Remove the old visualize command is it exists today in the CLI. 

Now that we've added the dedicated visualizer app which works in a completely different way, we should be using that instead.
Eventually there may be a need for a visualization from the CLI, at which time we can look at re-adding it in a way that reuses the other code to generate a consistent output. 

In addition, some dependency issues that we've been facing are coming from dependencies of the visualize command specifically. Removing this will allow us to more consistently install the CLI across different environments.

aidanm3341:
remove old visualize command


package.json
```diff
@@ -24,7 +24,6 @@
   "devDependencies": {
     "link": "^2.1.1",
     "npm-run-all2": "^5.0.0",
-    "vitest": "^2.1.8",
-    "jsdom": "^25.0.1"
+    "vitest": "^2.1.8"
```
willosborne, line 27: is this actually part of this PR? or just something left over in your dev env?

willosborne: ignore

willosborne (approved):

lbulanti-ms (approved):

Issue #830, opened 2025-01-23
Pull SchemaDirectory out of generate command to make available across tools


willosborne:
### Description of Problem:
Currently many commands use their own mechanisms to load schemas.
SchemaDirectory is the most flexible of these and is only used by generate.
See #790 for more details on the refactor.

This issue is just to refactor the SchemaDirectory to be at a higher level. Smaller issues will then adapt the different commands to use it for loading schemas

### Potential Solutions:
Pull out from generate to top level within `shared`.

Identify any generate-specific logic within the class.

willosborne:
This is now done. Two commands - generate and validate - are now using the schema directory.

PR #838, opened 2025-01-24
#830: Schema directory refactor

willosborne:
Pulls the SchemaDirectory class out of the generate command and makes it behave a little better - doesn't call Process.exit(1) on failure any more and just rethrows the exception so consumers can handle it how they want.


LeighFinegold (approved):

shared/src/schema-directory.ts
```diff
@@ -57,7 +57,7 @@ export class SchemaDirectory {
             } else {
                 this.logger.error(err);
             }
-            process.exit(1);
+            throw err;
```
aidanm3341, line 60: now that we're just throwing this here, should we be catching it in the CLI? I figured there would be some additional change along with this to handle the error

willosborne: I decided to let it get to the top because the error is pretty good. But maybe the stack trace is not what we want

aidanm3341 (approved):

lbulanti-ms (approved):

shared/src/schema-directory.ts
willosborne: Updated with more error handling

PR #844, opened 2025-01-25
Removed Record Builder annotations (#716)

grahampacker-ms:
addresses #716


willosborne:
Is there a reason we're removing these? Is it a code style thing or to do with supporting pre-java 17? Seems to introduce significantly more boilerplate by removing all the records.

willosborne (approved):

rocketstack-matt:
As discussed, we don't want the annotation processor

rocketstack-matt (approved):

PR #873, opened 2025-01-29
HOLD: Update Flox env

yt-ms:
- Update the Flox environment with the currently required Java version, and add docker and docker-compose for calmhub-server
- Define Flox services so the Mongodb, Calmhub server and Calmhub UI can all be started with a simple `flox services start`


yt-ms:
Hold off on this - need to double-check original problem.

rocketstack-matt:
@yt-ms is this still in progress or should it be closed?

Issue #689, opened 2024-12-17
Expose the CALM CLI as a RESTful API


Budlee:
### Description of Problem:

The problem that I face is that I am working with dockerized Java appliation and would like to use the calm validation that has been created within the CALM cli.
Currently there is no easy way to access the CLI from a dockerized application as it requires the cli to be added to the container then calling to the system process to run the cli and looking at the output.
This is a less than optimal solution.

### Potential Solutions:

I would like to propose is to add an option to the calm cli labelled `server`.

```shell
calm server -p 8080
```

This new option would then spin up a simple server to host the calm cli validation behaviour as an API endpoint that can be called by other applications.

The validation endpoint as an API will provide better interopoability for other services that want to utilise what has been built.
Additionally the process of dockerizing and running such an application would be more straight forward, it would be great id the CLI could be put onto dockerhub
I see this being used as a sidecar that an application can call locally.

```shell
calm server -p 8080
java myApp --validate-calm localhost:8080
```

TAG: @jpgough-ms @rocketstack-matt

rocketstack-matt:
@Thels have you already picked this up? If so can you assign to yourself, thanks!

Thels:
I did play around with spinning up an `expressJS` server as part of a `calm server` command - and that's went fine ( and I can raise a PR for that small amount of work ). But ran into some problems with trying to test it in the `cli.spec.ts` - but a proper implementation tackling this issue would need;

- OpenAPI support - likely moving to `https://hono.dev/` as it has great OpenAPI support and is very Express like.
- Definition of what API's we want to expose ( for example do we want to expose all formats of output validation or only JSON ) 

I'll take another stab at this tomorrow as my day is looking quite empty, at least if it's formalizing the "starting point" I mentioned at the top of this PR.

Thels:
> @Thels have you already picked this up? If so can you assign to yourself, thanks!

But fair point on the assignee - fixed it to myself now.

Budlee:
After the discussion in the Office Hours call (#776) we want to re-scope this issue. @jpgough-ms shall we close this open a new one or just edit the title and provide and update in the initial description

Thels:
Discussed briefly on a call;

- Currently validate CLI requires both an architecture and a pattern file path. @Budlee  would like to be able to have the architecture be just JSON in the body. So we need to think about saving it down as part of the server into a temporary location.
[...]

PR #877, opened 2025-01-29
Features/#689 CLI Server 

Budlee:
Adding in a first iteration of the CLI as a server for issue #689


cli/package.json
```diff
@@ -28,8 +28,12 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/supertest": "^6.0.2",
+    "axios": "^1.7.9",
+    "canvas": "^3.0.0",
```
willosborne, line 33: How come canvas is needed? IIRC this is the library that was breaking the NPM install finos/calm-cli previously

cli/package.json
```diff
@@ -28,8 +28,12 @@
...
     "commander": "^13.0.0",
     "copyfiles": "^2.4.1",
+    "jest-environment-node": "^29.7.0",
```
willosborne, line 36: jest is a test dependency and should go in devDeps

cli/src/index.ts
```diff
@@ -75,6 +80,33 @@ async function runValidate(options) {
...
 }
 
+program
+    .command('server')
+    .description('Start a HTTP server to proxy CLI commands. (experimental)')
+    .option('-p, --port <port>', 'Port to run the server on', '3000')
+    .requiredOption(SCHEMAS_OPTION, 'Path to the directory containing the meta schemas to use.')
+    .option(VERBOSE_OPTION, 'Enable verbose logging.', false)
+    .action((options) => {
+
```
willosborne, line 90: Is there anyway we could extract this server logic into a separate module? Feels wrong having all this business logic inside the arg parsing callback

cli/package.json
```diff
@@ -28,8 +28,12 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "@types/supertest": "^6.0.2",
+    "axios": "^1.7.9",
```
willosborne, line 32: is axios needed at runtime?

cli/src/routes/validation-route.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+const validationRouter = new ValidationRouter(router, schemaDirectory);
+app.use('/calm/validate', router);
+
+describe('ValidationRouter', () => {
+    test('should return 400 when $schema is not specified', async () => {
+        const expectedFilePath = path.join(__dirname, '../../test_fixtures/validation_route/invalid_api_gateway_instantiation_missing_schema_key.json');
+        const invalidArchitectureMissingSchema = JSON.parse(
+            fs.readFileSync(expectedFilePath, 'utf-8')
+        );
+        const response = await request(app)
```
willosborne, line 58: We use supertest request in some places, and axios in others. Other tests in the cli use just plain fetch.
Would be good to centralise on just one framework if possible

cli/src/routes/validation-route.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+//     });
+
+const schemaDirectory = '/Users/matthewauburn/Coding/Java/architecture-as-code/calm';
+const app: Application = express();
+app.use(express.json());
+
+const router: express.Router = express.Router();
+const validationRouter = new ValidationRouter(router, schemaDirectory);
+app.use('/calm/validate', router);
```
willosborne, line 50: What do you think about maybe putting this stuff in a beforeEach() style lifecycle hook so the tests don't potentially interfere with one another? Not sure if it's needed.

cli/src/routes/validation-route.ts
```diff
@@ -0,0 +1,70 @@
...
+    private validateSchema = async (req: Request, res: Response) => {
+        const schema = req.body['$schema'];
+        if (!schema) {
+            return res.status(400).json({ error: 'The "$schema" field is missing from the request body.' });
+        }
+        console.info("Path loading schemas is " + this.schemaDirectoryPath)
+
+        await this.schemaDirectory.loadSchemas(this.schemaDirectoryPath)
+        console.info("Loaded schemas: " + this.schemaDirectory.getLoadedSchemas())
+        const foundSchema = this.schemaDirectory.getSchema(schema)
```
willosborne, line 34: Can't remember off the top of my head - does this throw an exception here?

cli/src/routes/validation-route.ts
```diff
@@ -0,0 +1,70 @@
...
+            return res.status(400).json({ error: 'The "$schema" field referenced is not available to the server' });
+        }
+        const tempInstantiation = await createTemporaryFile()
+        const tempPattern = await createTemporaryFile()
+        try {
+
+            await fs.writeFile(tempInstantiation, JSON.stringify(req.body, null, 4));
+            await fs.writeFile(tempPattern, JSON.stringify(foundSchema, null, 4));
+
+            const outcome = await validate(tempInstantiation, tempPattern, this.schemaDirectoryPath, true);
```
willosborne, line 45: That is nice!!! I was thinking this would need you to somehow rework the validate command to take loaded in data (that is still an option, I guess if we have problesm with the file loading approach)

cli/src/routes/validation-route.ts
```diff
@@ -0,0 +1,70 @@
...
+            const content = getFormattedOutput(outcome, 'json');
+            res.status(201).type('json').send(content);
+        } catch (error) {
+            res.status(500).json({ error: error.message });
+        } finally {
+            [tempInstantiation, tempPattern].forEach(element => {
+                fs.unlink(element).catch(() => { });
```
willosborne, line 52: What does this do? is this the node/fs equivalent of running 'rm'?

cli/src/routes/validation-route.ts
```diff
@@ -0,0 +1,70 @@
...
+        console.info("Loaded schemas: " + this.schemaDirectory.getLoadedSchemas())
+        const foundSchema = this.schemaDirectory.getSchema(schema)
+        if (foundSchema == undefined) {
+            return res.status(400).json({ error: 'The "$schema" field referenced is not available to the server' });
+        }
+        const tempInstantiation = await createTemporaryFile()
+        const tempPattern = await createTemporaryFile()
+        try {
+
+            await fs.writeFile(tempInstantiation, JSON.stringify(req.body, null, 4));
```
willosborne, line 42: Probably fine, but are there any security concerns from writing a file to the FS? I think there is no way for the user to set the path which is probably fine

cli/src/routes/validation-route.ts
willosborne: Also I see we don't do anything if they fail - basically swallowing the exception. I get that we don't want a failure in cleanup to return a bad status code, but we should at least log. Otherwise the host could fill up with temp files

docs/README.md
```diff
@@ -2,3 +2,4 @@
 
 https://calm.finos.org/
 
+I
```
willosborne, line 5: Is this needed

shared/src/schema-directory.ts
```diff
@@ -38,12 +39,14 @@ export class SchemaDirectory {
             this.logger.debug('Loading schemas from ' + dir);
             const files = await readdir(dir, { recursive: true });
 
-            const schemaPaths = files.filter(str => str.match(/^.*(json|yaml|yml)$/))
+            const schemaPaths = files.filter(str => str.match(/^.*(json)$/))
                 .map(schemaPath => join(dir, schemaPath));
-
+            console.info("Schema Paths found: " + schemaPaths)
```
willosborne, line 44: Should use the winston logger instead of console.info

shared/src/schema-directory.ts
```diff
@@ -38,12 +39,14 @@ export class SchemaDirectory {
...
             for (const schemaPath of schemaPaths) {
                 const schema = await this.loadSchema(schemaPath);
-                map.set(schema['$id'], schema);
+                if (schema){
```
willosborne, line 47: So this means that if we somehow load an empty schema, we don't break the schema dir?

willosborne (changes requested):

cli/src/routes/routes.ts
```diff
@@ -0,0 +1,18 @@
...
+    router: Router;
+
+    constructor(schemaDirectoryPath: string) {
+        this.router = Router();
+        this.initializeRoutes(schemaDirectoryPath);
+    }
+
+    private initializeRoutes(schemaDirectoryPath: string) {
+        new ValidationRouter(this.router, schemaDirectoryPath);
+        this.router.use(VALIDATE_ROUTE_PATH, this.router);
```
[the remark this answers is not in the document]
Budlee: Updated to make this neater

cli/src/routes/validation-route.ts
Budlee: Nope. returns undefined if it does not exist

cli/src/server/cli-server.ts
```diff
@@ -0,0 +1,20 @@
+
```
aidanm3341, line 1: random new lines? you might need to configure your vscode to run the formatting. 'Prettier' should be configured, but your IDE might not be picking it up

cli/src/server/cli-server.ts
```diff
@@ -0,0 +1,20 @@
...
+
+
+import express, { Application } from 'express';
```
aidanm3341, line 4: I think express is fine for now because this is so simple. In the future we might want to consider something more advanced like nestjs

cli/README.md
```diff
@@ -161,6 +161,30 @@ You would get an output which includes a warning like this:
...
+
+It may be required to have the operations of the CALM CLI available over rest.
+The `validate` command has been made available over an API
+
+```shell
+calm server --schemaDirectory calm
+```
+
+```shell
+curl http://127.0.0.1:3000/calm/validate/health
```
aidanm3341, line 174: seems odd that the health endpoint is under the 'validate' path rather than at the root of the API? I reckon it would make more sense to just be `http://127.0.0.1:3000/health`

cli/src/server/cli-server.ts
aidanm3341, line 1: General question: is the CLI directory the right place for all this logic? Currently this dir only contains a few files which just delegate out to the shared folder - maybe it makes sense for this to live there as well? I could see someone wanting to run this without having to go through the CLI to start it

cli/src/server/routes/validation-route.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.logger = initLogger(debug);
+        this.initializeRoutes(router);
+    }
+
+    private initializeRoutes(router: Router) {
+        router.post('/', this.validateSchema);
+        router.get('/health', this.healthCheck);
+    }
+
+    private validateSchema = async (req: Request, res: Response) => {
```
Thels, line 28: I'm not sure what the best long-term approach to do here is - but ExpressJS ( and I assume other frameworks ) support casting the requests into specific objects.

So if you were to define a `ValidateRequest` - with the attributes you expect you can then have it be `async (req: ValidateRequest)`. Keen to know others opinions on this though.

cli/src/cli.spec.ts
```diff
@@ -164,28 +164,37 @@ describe('CLI Integration Tests', () => {
...
-                console.log('Response received');
-                expect(response.status).toBe(200);
-                serverProcess.kill();
-                done();
-            } catch (error) {
-                serverProcess.kill();
-                done(error);
-            }
-        }, 5000);
+        await new Promise(resolve => setTimeout(resolve, 10000));
```
willosborne, line 191: Nice

cli/src/server/routes/validation-route.ts
```diff
@@ -0,0 +1,72 @@
...
+        } finally {
+            [tempInstantiation, tempPattern].forEach(element => {
+                fs.unlink(element).catch(() => {
+                    this.logger.warn('Failed to delete temporary file ' + element);
+                });
+            });
+        }
+    };
+
+    private healthCheck(_req: Request, res: Response) {
```
Thels, line 62: I only had a `healthCheck` on my WIP PR as it was easy to ping something which just ACK'd. Do we need it?

cli/src/server/cli-server.ts
willosborne: the way i thought about it was that the shared has the logic, and the cli module is about the execution environment. so for the cli version that's parsing logic and passing it along. but i think it also kinda makes sense for this to be in the same place since ultimately its just about gathering input from the user

Thels: I think this is a fair suggestion - maybe a new folder under the shared module would make sense. Do we then want to call out in the CLI README that it only exists as a delegator?

cli/src/cli.spec.ts
```diff
@@ -13,7 +15,7 @@ describe('CLI Integration Tests', () => {
     const millisPerSecond = 1000;
     const integrationTestPrefix = 'calm-test';
     const projectRoot = __dirname;
-    jest.setTimeout(30 * millisPerSecond);
+    jest.setTimeout(30 * millisPerSecond); // Uncomment this line to set the timeout for Jest tests
```
Thels, line 18: Do you mean we should comment the line if we don't want a timeout?

cli/src/cli.spec.ts
```diff
@@ -182,6 +184,18 @@ describe('CLI Integration Tests', () => {
         });
     });
 
+    test('example server command - starts server and responds to requests', async () => {
+        const serverCommand = 'calm server -p 3001 --schemaDirectory ../../dist/calm/';
+        const serverProcess = exec(serverCommand);
+        // Give the server some time to start
+        await new Promise(resolve => setTimeout(resolve, 10000));
```
Thels, line 191:
```suggestion
        await new Promise(resolve => setTimeout(resolve, 10 * millisPerSecond));
```

Is this timeout required? I only ask we await the npx command from returning in the call to `exec` - does the express server really take another 10s to start up?

cli/src/server/cli-server.ts
Budlee: removed

cli/README.md
Budlee: Corrected

Thels: I'm not sure health is required - it was something I had in my original PR to debug issues with the CLI and end to end tests. If we want proper health endpoints we should use something like `terminus`.

Thels: If we do want to keep it though until we introduce a proper health check - I do agree it should be at the root.

Thels (changes requested):
Overall looking really good for a v1 - if we decide to keep `health` endpoint it should move to root so that when we introduce a proper health check framework ( like terminus ) they're using the same URLs

willosborne (dismissed):

cli/src/cli.spec.ts
Budlee: Updated to 5 seconds

cli/src/cli.spec.ts
Budlee: Removed

cli/src/server/routes/validation-route.ts
Budlee: I think its useful to have that we know the service is up, for liveness I would use this

cli/src/server/cli-server.ts
aidanm3341: It just seems to me that the CLI just happens to be the way that we're starting the server, rather than it being intrinsically to do with the CLI. There's nothing to stop someone from just starting this up with node directly

cli/src/server/routes/validation-route.ts
aidanm3341: yeah good shout, I agree. Looks like Express does support this where we can give extra type information as generics to the Request type. That would be a bit better

cli/src/server/cli-server.ts
Budlee: I completely disagree, this is a mini server option for the CLI, not a full blown production grade offering. If it's shared then we move it late. However, I do not think that this something that someone would want in a shared lib

cli/src/server/routes/validation-route.ts
Budlee: I accept the Response. I disagree with the request to cast it to anything but a string. The request needs to be saved to a document so if it's converted to a type (a CALM doc) then I need to transform it back again. If we transform it then potentially the transform will be lossy

aidanm3341: from an API design perspective, should the request body itself even be the architecture? I wonder if it makes more sense to have it be something like
```json
{
    "architecture": "{\"$schema\": \"https://...etc }"
}
```

so that in the future we can extend this to also potentially take a pattern in the request as well. This also means that the schema for the API is well defined as opposed to just some random json object that we're checking

Budlee: The initial proposal is that the patterns you allow to validate are all pre-loaded into the server. We did not want to have the pattern being passed in

aidanm3341 (approved):

cli/src/server/cli-server.ts
Budlee: @Thels, I know you mentioned in our conversation this was not a blocker and I also discussed with @aidanm3341. @aidanm3341 was happy that this could be moved if required due to my point made above

PR #901, opened 2025-02-13
First version of calm template for early review (#890)

LeighFinegold:
## **Summary**  
This PR introduces the initial version of calm template, enabling the generation of various outputs from a CALM model using Handlebars templates. The goal is to provide a flexible and extensible way to generate configurations (e.g., Kubernetes YAML), documentation (e.g., Markdown), and other artifacts from architecture models.

## **Key Features**  
- ✅ Supports different types of output generation (single vs. multi-file).  
- ✅ Uses **Handlebars** as the templating engine, with support for partials.  
- ✅ Allows users to define custom template bundles via `index.json`.  
- ✅ Implements a **parser and visitor pattern** for transforming the CALM model into a template-friendly format.  
- ✅ CLI integration (initial setup, further refinements expected).  
- ✅ **Adds document name support for repeated output types** (ensuring uniqueness in multi-file generation).  

## **Next Steps / Areas for Feedback**  
- 🔹 General feedback on the approach and design.  
- 🔹 Validation of the visitor pattern for model transformation or using other languages like [Zod](https://www.npmjs.com/package/zod) 
- 🔹 Add unit tests for CLI endpoint to ensure robustness and correct output generation.
- 🔹 Additional template examples (per initial demo) and developer.md 
- 🔹 Write additional unit test for logger.ts

Sample visual of cli output
image

Issue #899, opened 2025-02-13
CLI Model Discussion


LeighFinegold:
As part of follow up meeting comments on https://github.com/finos/architecture-as-code/issues/890#issuecomment-2653028005, there was a discussion the model that people can use for their own custom templates.

The current model in [types.ts](https://github.com/finos/architecture-as-code/blob/main/shared/src/types.ts) is not great.

-	Mainly of interfaces sitting in types.ts. Interfaces are not great because it gives you no safety at runtime i.e. the interface information is lost when compiled into javascript. So you could be parsing a document into an interface which is completely invalid.
-	Kebab-case use in json schema spec makes class modelling into camelCase meh as we need to stick single quotes ‘’ around field definitions.  
-	It is also missing a lot of current core.json definitions and relationship-type is not modelled how its defined in json-schema currently. See [@McPhelim, Aidan (Application Infrastructure)](mailto:Aidan.McPhelim@morganstanley.com) comment on https://github.com/finos/architecture-as-code/issues/826.

Most code so far e.g. generator builds a whole parsing layer and instantiate classes we want.

Some options
1.	We continue with the manual approach https://github.com/finos/architecture-as-code/commit/1b7ba293bea70e387ea63f1781462f1e3d5d1e3b (similar to generator.ts code). Note at this point not really sure the need to maintain interfaces at all as can’t make class implement the interface unless on class definitions we maintain kebab-case. Note: assuming we go with classes, you can’t just do const model: CalmNode = JSON.parse(x) so you need your own conversion method. 
2.	We create a generic kebab-case to camelCase converter of a document before parsing into interfaces. No real validation but less likely to run into runtime issues. 
3.	[class-transformer - npm](https://github.com/finos/architecture-as-code/issues/class-transformer%20-%20npm) – converts a document into class hierarchy  (prob too complicated) https://github.com/LeighFinegold/architecture-as-code/pull/5/commits/b0d16e3c9910b35403f635fd81770f911149c9e1
4.	[zod - npm](https://www.npmjs.com/package/zod?activeTab=versions) – this was mentioned by [@McPhelim, Aidan (Application Infrastructure)](mailto:Aidan.McPhelim@morganstanley.com) and seems very powerful in giving compile time and runtime validation of our model. [Zod | Documentation](https://zod.dev/?id=objects).  Trialed using jsonschema-to-zod [here](  https://github.com/LeighFinegold/architecture-as-code/pull/6)
[...]

Issue #790, opened 2025-01-12
Typescript CALM Module Refactoring - Visualizer, Doc Publisher, Schema Directory, CALM Model


LeighFinegold:
#### **Background**
Over recent discussions with developers ([@aidanm3341](https://github.com/aidanm3341), [@oliviajanejohns](https://github.com/oliviajanejohns), [@willosborne](https://github.com/willosborne), [@Thels](https://github.com/Thels), [@Budlee](https://github.com/Budlee), [@YoofiTT96](https://github.com/YoofiTT96), [@Adwoa-Konadu-Appiah](https://github.com/Adwoa-Konadu-Appiah)), we’ve been reviewing the current CALM codebase and future pull requests around documentation generation, CALM visualisation and the [CLI server](https://github.com/finos/architecture-as-code/issues/689).

The current structure of the project uses the following organization:
- **Workspaces**:
  - `shared`: Contains submodules like `generate`, `validate`, and `visualize`.
  - `cli`: Published separately to npm and includes CLI-specific logic.
  - `docs`: Documentation-related functionality.

Relevant future contributions include:
- **Visualization PR [#786](https://github.com/finos/architecture-as-code/pull/786)**: Recent improvements to visualization.
- **Docify prototype branch [calm-docify](https://github.com/LeighFinegold/architecture-as-code/tree/calm-docify)** as part of [Issue #478](https://github.com/finos/architecture-as-code/issues/478).

---

#### **Refactoring Discussed**
1. **Visualizer as a module to remain.**:
   - The **CALM Visualizer tool** is an essential component to help users build CALM documents. It is due to be committed as its own workspace. The current visualiser module under shared was due to be deleted. However, there is consensus that a **separate visualizer module** should be created to support and integrate with **docify** workflows. This modular visualizer would allow docify and other tools to leverage visualization capabilities independently of the CLI and the tool and would make https://github.com/finos/architecture-as-code/issues/782 possible.
   - The [**translator module**](https://github.com/finos/architecture-as-code/tree/main/translator), currently written in Java, was also discussed. We agreed it would be imported into the **visualizer module**
   - It is possible that the visualiser cli endpoint is not needed as will be exposed by document generator commands.

2. **SchemaDirectory Duplication**:
[...]

PR #902, opened 2025-02-14
New centralised model for CALM (#790)

LeighFinegold:
## 🔍 Summary  
This PR introduces a centralized and structured data model for CALM. Per discussion on https://github.com/finos/architecture-as-code/issues/899, The new model provides well-defined TypeScript classes and enums for core CALM components, ensuring seamless integration and validation for cli and visualizer componenets. 

## ✨ Key Changes  
- Added types that match the current schema files.
- Introduced `CalmCore` as the centralized entry point for nodes, relationships, metadata, controls, and flows.  
- Standardized structure for `CalmNode`, `CalmRelationship`, `CalmFlow`, `CalmControl`, and `CalmMetadata`.  
- Each model includes a `fromJson()` factory method to enable seamless conversion from JSON.  
- Ensures consistent handling of nested objects, such as relationships and interfaces.  
- Added unit tests for all models to ensure stability when we start to amend them to support potential new concepts like visitor pattern etc. 

## ❌ Out of Scope for this PR
- Not updating the existing codebase to move away from `types.ts` yet, as this will wait for the completion of SchemaDirectory changes as per https://github.com/finos/architecture-as-code/pull/897
- Once SchemaDirectory changes are  finalised, all existing tooling will be updated to leverage this model and expose it via CLI.  
- This update will then enable additional template-bundles related to Infrastructure as Code and Solution Architecture document summaries into Code as a follow up PR to  https://github.com/finos/architecture-as-code/pull/901

## Observations around CALM Specification
- We should look to move out metadata  into separate json file as definition duplicated in flows.json
- This PR corrects a spelling mistake on data-asset node-type
- The current core.json doesn't set any requirements on any of the attributes e.g. nodes, relationships. Should it also specify type: object as part of the specification. (Note: generators have required that to be defined)
- Should core.json actually have a collection of flowUrls instead of collection of flows? In the same way control detail has a requirementUrl which we expected to be linked to control-requirement.json

### DIFF
--- a/shared/src/model/control-requirement.spec.ts
+++ b/shared/src/model/control-requirement.spec.ts
@@ -1,0 +1,35 @@
+import { CalmControlRequirement } from './control-requirement.js'; // Importing the model
+import { CalmControlRequirementSchema } from '../types/control-requirement-types.js'; // Importing the schema
+
+// Mock data for testing
+const controlRequirementData: CalmControlRequirementSchema = {
+    'control-id': 'control-123',
+    name: 'Test Control Requirement',
+    description: 'This is a description of the test control requirement.'
+};
+
+describe('CalmControlRequirement', () => {
+    it('should create a CalmControlRequirement instance from JSON data', () => {
+        const controlRequirement = CalmControlRequirement.fromJson(controlRequirementData);
+
+        expect(controlRequirement).toBeInstanceOf(CalmControlRequirement);
+        expect(controlRequirement.controlId).toBe('control-123');
+        expect(controlRequirement.name).toBe('Test Control Requirement');
+        expect(controlRequirement.description).toBe('This is a description of the test control requirement.');
+    });
+
+    it('should handle missing description', () => {
+        const controlRequirementWithNoDescription: CalmControlRequirementSchema = {
+            'control-id': 'control-124',
+            name: 'Control with No Description',
+            description: ''
+        };
+        const controlRequirement = CalmControlRequirement.fromJson(controlRequirementWithNoDescription);
+
+        expect(controlRequirement).toBeInstanceOf(CalmControlRequirement);
+        expect(controlRequirement.controlId).toBe('control-124');
+        expect(controlRequirement.name).toBe('Control with No Description');
+        expect(controlRequirement.description).toBe('');
+    });
+
+});
--- a/shared/src/model/control-requirement.ts
+++ b/shared/src/model/control-requirement.ts
@@ -1,0 +1,17 @@
+import {CalmControlRequirementSchema} from '../types/control-requirement-types.js';
+
+export class CalmControlRequirement {
+    constructor(
+        public controlId: string,
+        public name: string,
+        public description: string
+    ) {}
+
+    static fromJson(data: CalmControlRequirementSchema): CalmControlRequirement {
+        return new CalmControlRequirement(
+            data['control-id'],
+            data['name'],
+            data['description']
+        );
+    }
+}
--- a/shared/src/model/metadata.ts
+++ b/shared/src/model/metadata.ts
@@ -1,0 +1,13 @@
+import { CalmMetadataSchema } from '../types/metadata-types.js';
+
+export class CalmMetadata {
+    constructor(public data: Record<string, unknown>) {}
+
+    static fromJson(data: CalmMetadataSchema): CalmMetadata {
+        const flattenedData = data.reduce((acc, curr) => {
+            return { ...acc, ...curr };
+        }, {} as Record<string, unknown>);
+
+        return new CalmMetadata(flattenedData);
+    }
+}
--- a/shared/src/model/node.ts
+++ b/shared/src/model/node.ts
@@ -1,0 +1,55 @@
+import {CalmInterface} from './interface.js';
+import {CalmControl} from './control.js';
+import {CalmMetadata} from './metadata.js';
+import {CalmNodeDetailsSchema, CalmNodeSchema} from '../types/core-types.js';
+
+export type CalmNodeType = 'actor' | 'ecosystem' | 'system' | 'service' | 'database' | 'network' | 'ldap' | 'webclient' | 'data-asset';
+
+export type CalmDataClassification = 'Public' | 'Confidential' | 'Highly Restricted' | 'MNPI' | 'PII';
+
+export class CalmNodeDetails {
+    constructor(
+        public detailedArchitecture: string,
+        public requiredPattern: string
+    ){}
+    static fromJson(data: CalmNodeDetailsSchema): CalmNodeDetails {
+        return new CalmNodeDetails(
+            data['detailed-architecture'],
+            data['required-pattern']
+        );
+    }
+}
+
+
+export class CalmNode {
+    constructor(
+        public uniqueId: string,
+        public nodeType: CalmNodeType,
+        public name: string,
+        public description: string,
+        public details: CalmNodeDetails,
+        public interfaces: CalmInterface[],
+        public controls: CalmControl[],
+        public metadata: CalmMetadata,
+        public dataClassification?: CalmDataClassification,
+        public runAs?: string,
+        public instance?: string,
+
+    ) {}
+
+    static fromJson(data: CalmNodeSchema): CalmNode {
+        return new CalmNode(
+            data['unique-id'],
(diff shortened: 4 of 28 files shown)


shared/src/model/control-requirement.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
```
aidanm3341, line 1: hmm, it would be good to not have this linting suppression. What if we use `unknown` on line 10 instead of `any`.

shared/src/model/control-requirement.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+        };
+
+        const control = CalmControlRequirement.fromJson(jsonData);
+
+        expect(control).toBeInstanceOf(CalmControlRequirement);
+        expect(control.controlId).toBe('456');
+        expect(control.name).toBe('JSON Control');
+        expect(control.description).toBe('Description from JSON');
+    });
+});
```
aidanm3341, line 26: how are we thinking about invalid input? Should these data classes be responsible for basic validation that the fields in the input data exist before using them? If not, maybe they should throw some parsing error

shared/src/model/relationship.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+
+export enum Protocol {
```
aidanm3341, line 3: enum = bad! Can we use a type of unioned strings instead please? E.g.

```typescript
export type Protocol = 'HTTP' | 'HTTPS' | 'FTP' | etc...
```

shared/src/model/node.ts
```diff
@@ -0,0 +1,70 @@
...
+    constructor(
+        public uniqueId: string,
+        public nodeType: CalmNodeType,
+        public name: string,
+        public description: string,
+        public details?: CalmNodeDetails,
+        public dataClassification?: CalmDataClassification,
+        public runAs?: string,
+        public instance?: string,
+        public interfaces?: CalmInterface[],
```
aidanm3341, line 50: to make using these model objects a little nicer to use, we could get rid of the optional aspect of these lists and instead default them to an empty list. That would mean that any client code using these models doesn't have to worry about checking for undefined before iterating over it.

```typescript
public interfaces: CalmInterface[] = []
```

shared/src/model/node.ts
```diff
@@ -0,0 +1,70 @@
...
+    Public = 'Public',
+    Confidential = 'Confidential',
+    HighlyRestricted = 'Highly Restricted',
+    MNPI = 'MNPI',
+    PII = 'PII'
+}
+
+export class CalmNodeDetails {
+    'detailed-architecture'?: string;
+    'required-pattern'?: string;
```
aidanm3341, line 29: why are we using the kebab case in our model object here rather than camel case? One of the benefits of this approach is that we don't have to do that right?

shared/src/model/node.ts
```diff
@@ -0,0 +1,70 @@
...
+
```
aidanm3341, line 30: also this class doesn't have a constructor like the others? Not sure if I'm missing something with this particular class

shared/src/model/node.ts
LeighFinegold: Yeah missed that one - is because originally had interfaces .

shared/src/model/relationship.ts
LeighFinegold: I actually moved from type to enum as liked you get that intellisense check, but will go with the above if we are saying that enums are avoided.

shared/src/model/node.ts
LeighFinegold: Yeah thinkng about it, the optionality I put in because it matched the json-schema btu given we are parsing I like the idea of avoiding such checks.

shared/src/model/control-requirement.ts
LeighFinegold: Ah yes - that was a topic wanted to raise but forgot. In the end I went with status quo in other classes so let's reiterate this on next meet-up. Will go with unknown

shared/src/model/control-requirement.spec.ts
aidanm3341: As long as we're consistent either way, e.g. I noticed that in the flow model you are doing some basic validations "// Ensure `value` is an object". If we're assuming that the input is valid, then we don't need that validation check there

shared/src/model/control-requirement.ts
LeighFinegold: Discussed this further with @aidanm3341 further and we agreed it would have been better to go with object than unknown. However, on further thought, I decided that actually it might be better to define types in the schema structures after all and have added this on latest commit.

shared/src/types/core-types.ts
```diff
@@ -0,0 +1,87 @@
...
+    controls?: CalmControlsSchema;
+};
+
+//TODO: There is no required section.
+export type CalmCoreSchema = {
+    nodes?: CalmNodeSchema[];
+    relationships?: CalmRelationshipSchema[];
+    metadata?: CalmMetadataSchema;
+    controls?: CalmControlsSchema;
+    flows?: CalmFlowSchema[];
```
LeighFinegold, line 83: One thing I considered was stating that this could be an array of CalmFlowSchema[] | String[]. For now I kept it as this, but it requires the calm doc to be de-referenced before parsing.

shared/src/parser/parser.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class CalmParser {
+
+    private static logger = initLogger(process.env.DEBUG === 'true');
+
+    parse(coreCalmFilePath: string): CalmCore {
+        const logger = CalmParser.logger;
+        try {
+            const data = fs.readFileSync(coreCalmFilePath, 'utf8');
+            const dereferencedData: CalmCoreSchema = JSON.parse(data); // TODO: this needs to use SchemaDirectory to parse the other documents e.g. flows.
```
LeighFinegold, line 14: I don't think this needs to block the PR to get the models in, but we really do need to ensure that the document is dereferenced prior to trying to parse into a model. If we decide we don't want to do that then likely CalmCore.fromJson doesn't get called. I think the individual Calm method fromJson e.g. CalmFlow.fromJson would though.

shared/src/types/core-types.ts
```diff
@@ -0,0 +1,87 @@
...
+    protocol?: CalmProtocolSchema;
+    authentication?: CalmAuthenticationSchema;
+    metadata?: CalmMetadataSchema;
+    controls?: CalmControlsSchema;
+};
+
+//TODO: There is no required section.
+export type CalmCoreSchema = {
+    nodes?: CalmNodeSchema[];
+    relationships?: CalmRelationshipSchema[];
```
aidanm3341, line 80: aren't nodes and relationships required?

LeighFinegold: Not atm on core.json which is why I put that an observation on this PR description

shared/src/model/metadata.ts
```diff
@@ -0,0 +1,13 @@
+import { CalmMetadataSchema } from '../types/metadata-types.js';
+
+export class CalmMetadata {
+    constructor(public data: Record<string, unknown>) {}
+
+    static fromJson(data: CalmMetadataSchema): CalmMetadata {
+        const flattenedData = data.reduce((acc, curr) => {
+            return { ...acc, ...curr };
```
willosborne, line 8: Very nifty.

shared/src/model/metadata.ts
```diff
@@ -0,0 +1,13 @@
+import { CalmMetadataSchema } from '../types/metadata-types.js';
+
+export class CalmMetadata {
+    constructor(public data: Record<string, unknown>) {}
+
+    static fromJson(data: CalmMetadataSchema): CalmMetadata {
+        const flattenedData = data.reduce((acc, curr) => {
```
willosborne, line 7: Does this flatten only the top level? We probably don't want it to recursively flatten all the nested objects

willosborne (approved):

shared/src/model/metadata.ts
LeighFinegold: Should be only first level, but can add a test case to confirm on subsequent commit.

Issue #761, opened 2025-01-03
CalmHub: Sanitize Outputs to prevent XSS attacks


grahampacker-ms:
### Description of Problem:
As we will be rendering ADRs on a web app, we need to consider the sanitisation of the content we are displaying.
We will be accepting Markdown content which can contain embedded HTML which can be an attack vector for XSS attacks.

### Description of Problem:

- Sanitize HTML elements before storage in CalmHub
- Sanitize HTML elements on GET operation in CalmHub
- Sanitize HTML elements in the CalmHub UI

grahampacker-ms:
I propose we use this library for sanitization: https://owasp.org/www-project-java-html-sanitizer/

grahampacker-ms:
@jpgough-ms the Snyk Source Advisor doesn't seem to support Java dependencies: https://snyk.io/advisor/

jpgough-ms:
Looks like the library is well supported @grahampacker-ms

jpgough-ms:
#923 has been opened @grahampacker-ms to address part of this issue, but we are aware that there is more to fix. Can you let us know if you have opportunity to work on this issue? I'm going to merge @rocketstack-matt PR for now.

grahampacker-ms:
yes I'll get on it @jpgough-ms

grahampacker-ms:
I took a look at the issue with #923 and I wonder if this should be dealt with by having stricter input validation rather than HTML escaping? For example, should a version be restricted to the #.#.# format? should a schema name be restricted to alphanumeric plus some specific characters(perhaps spaces and hyphens?

PR #936, opened 2025-02-27
Sanitize ADR inputs for HTML elements (#761)

grahampacker-ms:
part of #761


calm-hub/src/main/java/org/finos/calm/sanitizer/AdrSanitizer.java
```diff
@@ -0,0 +1,50 @@
...
+import org.finos.calm.domain.adr.NewAdrRequest;
+import org.finos.calm.domain.adr.Option;
+import org.owasp.html.HtmlPolicyBuilder;
+import org.owasp.html.PolicyFactory;
+
+@ApplicationScoped
+public class AdrSanitizer {
+
+    //strict policy that strips all HTML elements
+    private final PolicyFactory strictPolicy = new HtmlPolicyBuilder().toFactory();
```
rocketstack-matt, line 15: Is the default what we want, won't this remove the ability for even formatting HTML?

calm-hub/src/main/java/org/finos/calm/sanitizer/AdrSanitizer.java
rocketstack-matt, line 1: Would it be more idiomatic to either have sanitized record equivalents (e.g. SanitizedNewAdrRequest) or else implement a sanitized method implemented in each Record rather than having a utility class dealing with the sanitisation of each record type?

grahampacker-ms: I was initially trying to keep the record as a POJO, but I'm now thinking it would be better change to a class and sanitize in the constructor. Thoughts?
This would avoid a utility class and also duplicate types

calm-hub/src/main/java/org/finos/calm/sanitizer/AdrSanitizer.java
grahampacker-ms: For ADRs, my understanding is that we will support markdown. This should be safer than allowing HTML. If there's certain tags we want to allow in future we can relax the policy. There's some default policies we can use that can be found here: https://javadoc.io/doc/com.googlecode.owasp-java-html-sanitizer/owasp-java-html-sanitizer/latest/org/owasp/html/Sanitizers.html#BLOCKS

calm-hub/pom.xml
```diff
@@ -53,6 +53,11 @@
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jsr310</artifactId>
     </dependency>
+    <dependency>
```
jpgough-ms, line 56: If we're introducing this sanitiser can we now remove this block added by @rocketstack-matt and do this consistently?

```
<dependency>
  <groupId>org.apache.commons</groupId>
  <artifactId>commons-text</artifactId>
  <version>1.13.0</version>
  </dependency>
```
I think I tagged you in the original PR #923

calm-hub/src/main/java/org/finos/calm/resources/AdrResource.java
```diff
@@ -83,7 +86,9 @@ public Response getAdrsForNamespace(@PathParam("namespace") String namespace) {
             description = "Creates an ADR for a given namespace with an allocated ID and revision 1"
     )
     public Response createAdrForNamespace(@PathParam("namespace") String namespace, NewAdrRequest newAdrRequest) {
-        Adr adr = new Adr.AdrBuilder(newAdrRequest)
+        NewAdrRequest safeNewAdrRequest = adrSanitizer.sanitizeNewAdrRequest(newAdrRequest);
```
jpgough-ms, line 89: Wondering if this can just form part of the `Adr.AdrBuilder` it's a valid step required to build a compliant ADR. Trouble with this separation is folks will forget if we use this elsewhere

calm-hub/src/main/java/org/finos/calm/resources/AdrResource.java
```diff
@@ -83,7 +86,9 @@ public Response getAdrsForNamespace(@PathParam("namespace") String namespace) {
...
+
+        Adr adr = new Adr.AdrBuilder(safeNewAdrRequest)
```
jpgough-ms, line 91: Concerns me that we don't have a unit test here to verify on `createAdrForNamespace` any characters are escaped

calm-hub/src/main/java/org/finos/calm/sanitizer/AdrSanitizer.java
jpgough-ms: See my similar comment in `AdrSanitizer`

calm-hub/src/test/java/org/finos/calm/sanitizer/TestAdrSanitizerShould.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@QuarkusTest
+class TestAdrSanitizerShould {
+
+    @Inject
+    AdrSanitizer adrSanitizer;
+
+    @Test
+    public void remove_html_tags_from_adr() {
```
jpgough-ms, line 25: I hadn't spotted this before, but in JUnit5 tests should not be `public` the can just be:

`void remove_html_tags_from_adr()`

Can you correct this where it has been introduced across the ADR work?

jpgough-ms (changes requested):
As a good practice, it would be helpful for a PR to contain a summary of the code related changes and any assumptions as it relates to the issue. It takes a while to get into the context otherwise when reviewing change.

I think there is a discussion with @rocketstack-matt to have on whether we support the storing of markdown vs html. It might be useful in this context to discuss what is stored by CalmHub and what will be rendered in the UI. I was thinking that the UI handles all the formatting and we just store the raw JSON. A UI mockup/discussion on office hours would help with this.

calm-hub/pom.xml
rocketstack-matt: Agreed we should have a consistent approach.

calm-hub/src/main/java/org/finos/calm/sanitizer/AdrSanitizer.java
rocketstack-matt: I think this is a better use of encapsulation and avoids littering the code with utility classes.

calm-hub/pom.xml
grahampacker-ms: I'll do this as a seperate PR. This one is focussed on ADRs

calm-hub/src/main/java/org/finos/calm/domain/adr/Link.java
```diff
@@ -1,4 +1,64 @@
...
+    public String getRel() {
+        return rel;
+    }
+
+    public String getHref() {
+        return href;
+    }
+
+    public void setRel(String rel) {
+        if(rel == null) {
```
jpgough-ms, line 35: Feel like this structure would be more idiomatic on these:

`this.rel = (rel == null) ? null : LINK_POLICY.sanitize(rel);`

A touch frustrating that the library does not gracefully handle nulls

jpgough-ms (changes requested):
I think there's a more idiomatic style to writing the null checks, suggested on one of the methods.

Now `Adr`, `Decision`, `Link`, `NewAdrRequest`, and `Option` have logic in there, I think there should be a test around the policy as it is applied on the setter. Testing this just with the resource ends up looking more like an integration test that maintaining a unit test style.

calm-hub/src/main/java/org/finos/calm/domain/adr/Link.java
grahampacker-ms: Unfortunately it changes nulls to empty strings which doesn't seem right.
I'll make this change

calm-hub/src/main/java/org/finos/calm/domain/adr/Decision.java
```diff
@@ -1,4 +1,55 @@
...
+    private final PolicyFactory DECISION_POLICY = Sanitizers.FORMATTING.and(Sanitizers.BLOCKS).and(Sanitizers.TABLES);
+    private Option chosenOption;
+    private String rationale;
+
+    public Decision() {
+
+    }
+
+    public Decision(Option chosenOption, String rationale) {
+        setChosenOption(chosenOption);
```
jpgough-ms, line 21: As this line doesn't do any validation (not sure if it should or not) this can just be 
`this.chosenOption = chosenOption`

calm-hub/src/main/java/org/finos/calm/domain/adr/NewAdrRequest.java
```diff
@@ -1,13 +1,106 @@
...
+            String contextAndProblemStatement,
+            List<String> decisionDrivers,
+            List<Option> consideredOptions,
+            Decision decisionOutcome,
+            List<Link> links
+    ) {
+        setTitle(title);
+        setContextAndProblemStatement(contextAndProblemStatement);
+        setDecisionDrivers(decisionDrivers);
+        setConsideredOptions(consideredOptions);
```
jpgough-ms, line 36: These last 3 sets, could just be a direct assignment. As a developer, if I was reviewing this code, I'd go look at the setter to see what was being done there. Nothing extra is being done there at the moment.

jpgough-ms (changes requested):
A couple of final minor changes.

calm-hub/src/main/java/org/finos/calm/domain/adr/Decision.java
grahampacker-ms: the Option constructor performs the sanitization for it's fields

calm-hub/src/main/java/org/finos/calm/domain/adr/NewAdrRequest.java
grahampacker-ms: that's fair. Was going for consistency but I can see how it would confuse

jpgough-ms (approved):

PR #945, opened 2025-02-28
Proposed migration to MSW from fetch-mock

rocketstack-matt:
Following several attempts to upgrade fetch-mock from v9 to v10 without success, I've looked at what alternatives exist that are more TypeScript friendly.

ChatGPT suggested the following options
Screenshot 2025-02-28 at 12 38 41

This PR adds a new version of `validate.spec.ts` called `validate.msw.spec.ts` in parallel to the fetch-mock tests as a proposed replacement.


shared/src/test/mocks/handlers.ts
```diff
@@ -0,0 +1,46 @@
+import { http, HttpResponse } from 'msw';
+import fs from 'fs';
+import path from 'path';
+
+// We'll export a function that creates handlers based on test needs
+export const createHandlers = (mocks: {
+    url: string;
+    response: string | number | object | { fixture: string };
+}[]) => {
```
aidanm3341, line 9: can we extract the type of the mocks array into a named type? e.g. 
```typescript
interface HandlerConfig {
    url:...,
    response:...
}
```

then create handlers becomes this (ideally using the function syntax)

```typescript
function createHandlers(mocks: HandlerConfig[]) { ... }
```

we can also then export that type and reuse it in the server file too

aidanm3341 (changes requested):

shared/src/test/mocks/handlers.ts
```diff
@@ -0,0 +1,46 @@
...
+    return mocks.map(mock => {
+        if (typeof mock.response === 'number') {
+            // Handle status code responses (e.g., 404)
+            return http.get(mock.url, () => {
```
aidanm3341, line 13: I know they will be small, but I think it would improve readability to extract these into functions. Mostly just because it's fairly nested and wrapped in a map function. 

E.g.

```typescript
function handleMockStatusCode(url: string, status: number) {...}
```

then this map function just becomes: 

```typescript
return mocks.map(mock => {
    if (typeof mock.response === 'number') {
        return handleMockStatusCode(mock.url, mock.response as number);
    } else if (typeof mock.response === 'string') {
        return handleMockStringResponse(mock.url, mock.response as string);
    } else if (...etc...
});
```

jpgough-ms (approved):

oliviajanejohns (approved):

aidanm3341 (approved):

Issue #715, opened 2024-12-31
CalmHub: Introduce OAuth2 and HTTPs Profile


jpgough-ms:
### Description of Problem:

CalmHub is purely in demo mode at the moment, and to deploy in an organisation will need to offer basic security. We need to ensure that we keep the product easy to develop with locally, so this will be a profile that can turn on or off. 

The idea will be to restrict the OAuth2 Client to be just used by the CLI. 

### Potential Solutions:

* We need an OAuth2 design to cover the use of Calm Hub, capturing the idea that we have the right number of scopes and a differentiation between write and read access. @Budlee is experienced here and can help add the design to this issue.
* Code modification to include support by any IDP
* Documentation of how to configure and setup CalmHub to use OAuth2 in a corporate deployment

Budlee:
This sounds something I can help with. Sounds like various avenues to explore. I will jump on the next office call to discuss these further.

challamani:
@jpgough-ms @Budlee I have a working example that I explored in our last Hackday using [keycloak](https://www.keycloak.org/guides), I'll demo it in our office hrs on Monday. 
- Authorization Code Flow with PKCE for CalmHub UI Access (using `oidc-client` dependency in React App) 
- Client Credentials Flow for service-to-service communication.

For granular authorization checks at CalmHub:
1. we have to define the roles & scopes based on CalmHub clients. 
 
2. Client Credential flow (for service-to-service communication): we may have to maintain one-to-many mapping, a single scope & corresponding endpoints mapping. 
example: `write:patterns -> [POST] /pattern`, `write:patterns -> [PUT] /pattern`, `read:patterns -> [GET] /patterns`

3. Authorization Code Flow with PKCE (for user access):
We may have to maintain the "roles to endpoints mapping (one-to-many mapping: a single role and associated list of endpoints that a user role is allowed to access)".

challamani-ms:
@jpgough-ms  @Budlee  
Scopes design proposal for fine-grained access control.

**Approach 1:** 
Namespace Resource Scopes:
- `namespace.read` - to read all namespaces
- `namespace.create` - to create a new namespace
- `namespace.manage` - to full control over namespaces (create, update, delete)

Pattern Resource Scopes: 
- `$namespaceId.read` 
- `$namespaceId.read` & `pattern.write` - to create new patterns.
- `$namespaceId.read` & `pattern.manage` - to full control of patterns associated with `$namespaceId`.
Cons: 
[...]

PR #978, opened 2025-03-06
finos#715 enable https profile and support coarse-grained access control using keycloak

challamani-ms:
Submitting this PR based on the discussion in #715

We can find the following changes in this PR
- Enable HTTPS profile in calm-hub using self-signed certificates.
- Docker compose script to launch KeyCloak Container with a sample realm.json
- JWT & Scope validation logic implemented in calm-hub.
- Used the oidc-client in CalmUI to enable Authorization Code flow (only when backend profile is https)
- Created a shell script to test the device-code flow.

Note: we will create two new issues to implement following
- Enable fine-grained access control (RBAC, user entitlement checks after a successful bearer token validation)
- Provide onboarding guide to integrate with different IdPs.


jpgough-ms:
Waiting to confirm things build and then will review

jpgough-ms (changes requested):
Build is failing

calm-hub/src/integration-test/java/integration/ScopesAllowedIntegration.java
```diff
@@ -0,0 +1,109 @@
...
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.empty;
+
+@QuarkusTest
+@TestProfile(IntegrationTestSecureProfile.class)
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+public class ScopesAllowedIntegration {
+
+    private static final Logger logger = LoggerFactory.getLogger(ScopesAllowedIntegration.class);
+    public static final String PATTERN = "{\"name\": \"demo-pattern\"}";
```
jpgough-ms, line 27: Can be private

calm-hub/src/integration-test/java/integration/ScopesAllowedIntegration.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @BeforeEach
+    public void setupPatterns() {
```
jpgough-ms, line 30: I think that this can be: 

`void setupPatterns()`

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
```diff
@@ -47,6 +48,7 @@ public ArchitectureResource(ArchitectureStore store) {
             summary = "Retrieve architectures in a given namespace",
             description = "Architecture stored in a given namespace"
     )
+    @ScopesAllowed({"architectures:all", "architectures:read"})
```
jpgough-ms, line 51: A comment for all these annotations in all files

It looks like this would be really easy to mess up. Should we create a `CalmHubScopes` class statically defining and fully documenting the scopes we want to support?

calm-hub/src/main/java/org/finos/calm/resources/NamespaceResource.java
```diff
@@ -20,6 +21,7 @@ public NamespaceResource(NamespaceStore store) {
             summary = "Available Namespaces",
             description = "The available namespaces available in this Calm Hub"
     )
+    @ScopesAllowed({"architectures:all", "architectures:read"})
```
jpgough-ms, line 24: I think this one needs checking, `/calm/namespaces` is used by both patterns and architectures. This might not be quite doing what you expect. I think a test around this would be good.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
```diff
@@ -39,6 +40,7 @@ public PatternResource(PatternStore store) {
             summary = "Retrieve patterns in a given namespace",
             description = "Patterns stored in a given namespace"
     )
+    @ScopesAllowed({"architectures:all", "architectures:read"})
```
jpgough-ms, line 43: Perhaps I am misunderstanding are patterns considered architectures? If so, that's probably not the right naming.

calm-hub/src/test/java/org/finos/calm/security/TestScopesAllowedFilterShould.java
```diff
@@ -0,0 +1,82 @@
...
+    private ScopesAllowedFilter scopesAllowedFilter;
+
+    @BeforeEach
+    public void setup() {
+        MockitoAnnotations.openMocks(this);
+        scopesAllowedFilter = new ScopesAllowedFilter(jwt, resourceInfo);
+    }
+
+    @Test
+    public void allow_the_request_when_scopes_not_defined_on_resource() throws NoSuchMethodException {
```
jpgough-ms, line 41: As we're using JUnit 5, I don't think this needs to be public

calm-hub/src/test/java/org/finos/calm/security/TestScopesAllowedFilterShould.java
```diff
@@ -0,0 +1,82 @@
...
+        Method method = TestNamespaceResource.class.getMethod("getNamespacesUnsecured");
+        when(resourceInfo.getResourceMethod()).thenReturn(method);
+
+        scopesAllowedFilter.filter(requestContext);
+        //verify(logger).warn("Unsecured endpoint accessed: "+ method);
```
jpgough-ms, line 46: Remove commented out code

jpgough-ms (changes requested):

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
challamani-ms: Make sense, I'll update accordingly, instead of hardcoding the scopes values, will use the static values.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
challamani-ms: Yes @jpgough-ms , we have created these scopes by considering `patterns` are part of `architectures`. 

For the coarse-grained access control we have defined two top-level scopes.
1. `architectures` scope for accessing the flows, patterns and architectures.
2. `adrs` scope for accessing the ADRs.

jpgough-ms: @rocketstack-matt interested in our opinion here, `patterns` and `architectures` end up getting grouped together under an `architectures` scope in this model.

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
jpgough-ms: @challamani-ms the static values could perhaps go in `@ScopesAllowed` - though thinking about it, perhaps it would read better as `PermittedScopes` i.e.

`@PermittedScopes({PermittedScopes.ALL, PermittedScopes.READ})

Another question is if you have ALL does READ need to be defined?

challamani-ms: Yeah, I think we need `READ` scope to allow only `GET` operations on resources. 
Example: CalmHub can reject any `POST/PUT` requests that contain only READ scopes, by having READ scope we can limit the client application access, in our case **CalmUI** is only allowed to request tokens with READ scopes, as it is not allowed to perform POST & PUT operations on any patterns.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
rocketstack-matt: My initial thought is that they're different, because I would think someone who can author a pattern isn't necessarily the same person using a pattern to generate an architecture or creating an architecture from scratch.

However, as I was typing that, isn't the separation really around the 'state' of a pattern or architecture. Is there any reason we would restrict anyone from being able to create a pattern or an architecture, but we perhaps need to have a different role that allows folks to approve architectures or patterns.

At this point, I think that role is actually split, in so far as someone who can approve an architecture isn't necessarily authorised to approve an entire pattern.

challamani-ms: Just an additional information @jpgough-ms @rocketstack-matt 
These scopes we have defined to control what a client application can do,  
example: 
-  Restrict the `CalmUI` only to read patterns, namespaces, and flow, by not granting the `architectures:all` scope to CalmUI client application in IdP.

Note: Once JWT validation is completed, CalmHub server validates the end-user level RBAC checks. 
We will create a separate PR, for the fine-grained user level access control, we are thinking to maintain a table in CalmHub backend to achieve that, we can find more details in #715.

challamani-ms: Probably we can introduce more scopes to control the client applications' access. 

Scope & associated actions:
- A client application with `architectures:all` scope, is allowed to perform all operations on flows, architectures, and patterns.
- A client application with `architectures:read` scope, is allowed to access only GET operations on flows, namespaces and patterns.
- A client application with `adrs:read` scope, is allowed to access only GET operations on adrs and namespaces. 
- A client application with `adrs:all` scope, is allowed to perform all operations on adrs.

Note: Additional user level RBAC checks in the Calm server are required, as OAuth2 cannot map the application scopes to User.

jpgough-ms:
@challamani-ms this branch will need a rebase, there are significant changes in the UI - which might need a bit of rework.

jpgough-ms (approved):

calm-hub-ui/src/ProtectedRoute.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }, []);
+
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  if (!user) {
+    return <div>Redirecting to login...</div>;
+  }
+  return <>{children}</>;
```
aidanm3341, line 37: One small thing is that the formatting is slightly inconsistent with the rest of the UI project. E.g. the indentation everywhere else is 4 spaces, this has 2. Also ideally we use the `function() {}` syntax for defining functions rather than the `const x = () => {}` syntax.

However these aren't really blockers for the PR I would say. The content of the PR looks great

jpgough-ms: I've created an issue to capture this @aidanm3341 and we can take this on separately

Issue #1026, opened 2025-03-17
Extend `calm template` to support dereferencing


LeighFinegold:
## **Description**
Currently, the generic template processor does not handle external URL-based references efficiently. This issue proposes an enhancement to:

1. Modify the template processor to support dereferencing URL links by first trying to hit the URL followed by local file system:
2. Extend the `calm template` CLI command and API endpoint to accept `--url-to-local-file-mapping` as an argument for local resolution when documents are yet to be published

  ```sh
  calm template --input model.json --bundle template-bundle --output output-dir --url-to-local-file-mapping url-to-file-directory.json
```

Proposed content of the file is as follows:-

```json
{ 
 "https://calm.finos.org/docuflow/flow/document-upload" : "flows/flow-document-upload.json"
}
```

LeighFinegold:
Closing. Discussed with @willosborne that we will bring this and SchemaDirectory together on a separate issue

PR #1027, opened 2025-03-17
Extend Calm Template to support de-referencing (#1026)

LeighFinegold:
## **Summary**
This PR implements **support for dereferencing external URL-based references** by:
1. Using `CompositeReferenceResolver` first to attempt URL resolution.
2. Falling back to a local file-based remediation strategy when URLs cannot be resolved.
3. Extending the `calm template` CLI command and API endpoint to accept `--url-to-local-file-mapping`.


cli/tsup.config.ts
```diff
@@ -7,7 +7,7 @@ export default defineConfig({
     format: ['cjs'],
     sourcemap: false,
     clean: true,
-    external: ['canvas', 'fsevents', /node_modules/],
+    external: ['canvas', 'fsevents', '@apidevtools/json-schema-ref-parser', /node_modules/],
```
LeighFinegold, line 10: If this is it not done. The whole index.js becomes corrupted! It took 3hrs to figure out that this needed to be done :(

shared/src/resolver/calm-reference-resolver.ts
```diff
@@ -0,0 +1,99 @@
...
+        if (!fs.existsSync(ref)) {
+            throw new Error(`File not found: ${ref}`);
+        }
+
+        return JSON.parse(fs.readFileSync(ref, 'utf-8'));
+    }
+}
+
+export class InMemoryResolver implements CalmReferenceResolver {
+    private mockData: Record<string, unknown>;
```
LeighFinegold, line 31: Minor: Forgot to change this variable. Originally this just resided in test but thought it could be useful still just to leave in main code.

shared/src/template/template-calm-file-dereferencer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    constructor(urlFileMapping: Map<string, string>, resolver: CalmReferenceResolver) {
+        this.urlFileMapping = urlFileMapping;
+        this.resolver = resolver;
+    }
+
+    private replaceUrlsWithRefs(data: unknown): unknown {
+        if (typeof data === 'string') {
+            const ref = this.urlFileMapping.get(data);
+            return ref ? { $ref: ref } : data;
```
LeighFinegold, line 19: For the moment we will not enforce people to leverage $ref when referring to an external URL in their implemented CALM docs, but to allow de-referencing using the json-schema-ref-parser, inject it in for them.

shared/src/template/template-calm-file-dereferencer.ts
```diff
@@ -0,0 +1,55 @@
+import $RefParser from '@apidevtools/json-schema-ref-parser';
+import {CalmReferenceResolver} from '../resolver/calm-reference-resolver';
```
LeighFinegold, line 2: Need to add .js

shared/src/template/template-engine.ts
```diff
@@ -85,14 +89,16 @@ export class TemplateEngine {
...
             for (const instance of dataSource) {
-                const filename = output.replace('{{id}}', instance.id); //TODO: Improve output naming for repeated use case.
+                const filename = output.replace('{{id}}', instance.id);
                 const outputPath = path.join(outputDir, filename);
+                fs.mkdirSync(path.dirname(outputPath), { recursive: true });
                 fs.writeFileSync(outputPath, this.templates[template](instance), 'utf8');
                 logger.info(`✅ Generated: ${outputPath}`);
             }
         } else if (outputType === 'single') {
-            const filename = output.replace('{{id}}', dataSource.id); //TODO: Improve output naming for repeated use case.
```
LeighFinegold, line 94: Shouldn't have removed this comment as still needs to be done.

shared/src/resolver/calm-reference-resolver.ts
```diff
@@ -0,0 +1,99 @@
...
+    async resolve(ref: string): Promise<unknown> {
+        if (!this.mockData[ref]) {
+            throw new Error(`Mocked reference not found: ${ref}`);
+        }
+        return this.mockData[ref];
+    }
+}
+
+
+export class HttpReferenceResolver implements CalmReferenceResolver {
```
willosborne, line 50: We probably need a way to turn off/override this for security purposes. Don't want it automatically reaching out to URLs which could end up going outside an organisation unintentionally.
I think as in other places the solution is to eventually replace this with a CALMHub focused resolver that uses the CALMHub protocol - but that work for schemadir isn't in yet (sorry)

shared/src/resolver/calm-reference-resolver.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.httpResolver = new HttpReferenceResolver();
+        this.fileResolver = new FileReferenceResolver();
+    }
+
+    canResolve(ref: string): boolean {
+        return this.httpResolver.canResolve(ref) || this.fileResolver.canResolve(ref);
+    }
+
+    async resolve(ref: string): Promise<unknown> {
+        if (this.httpResolver.canResolve(ref)) {
```
willosborne, line 87: Interesting that HTTP is at higher priority than local file system? Would we not want to be able to modify the local files when working locally?

willosborne (approved):

shared/src/resolver/calm-reference-resolver.ts
LeighFinegold: Very valid point! I think it makes sense to switch around the ordering

cli/src/index.ts
```diff
@@ -71,13 +72,14 @@ program
     .requiredOption('--input <path>', 'Path to the CALM model JSON file')
     .requiredOption('--bundle <path>', 'Path to the template bundle directory')
     .requiredOption('--output <path>', 'Path to output directory')
+    .option('--url-to-local-file-mapping <path>', 'Path to mapping file which maps URLs to local paths', '')
```
oliviajanejohns, line 75: Is the last parameter needed here? Is it not optional?

oliviajanejohns (dismissed):

willosborne (approved):

Issue #1046, opened 2025-03-19
Include Data Exchange Format On Interface


Adwoa-Konadu-Appiah:
### Description of Problem:
There is no way to describe the format when exchanging data between two nodes. 
Example: Binary, Json, XML, Text etc.

### Gap:
...describe why this feature doesn't fit into an existing project

### Potential Solutions:
We are proposing adding exchange formats to interfaces to allow users to specify the format of data that is exchanged between nodes using an interface.

` "data-exchange":{
      "$ref": "#/defs/interface-type",
      "type": "object",
      "properties": {
        "format": {
          "type": "array",
          "items": {
            "$ref": "#/defs/exchange-format"
          },
          "uniqueItems": true
        }
      },
      "required": [
        "format"
      ]
  },
  "exchange-format": {
    "enum": [
      "binary",
      "file-upload",
      "json",
      "text",
      "xml",
      "other"
    ]
  }`

### Onboarding Checklist
The below must be completed before a new project is accepted into the monorepo:
- [ ] `.github/CODEOWNERS` is updated to add your new module. You should use GitHub Teams for this rather than individuals.
- [ ] Ensure workflow actions are added to `.github/workflows` that build/verify the module for any PR's with it in its blast radius. For example if your module depends on shared it should trigger build/tests on shared changes. 
- [ ] `README.md` updated with new module and owners.  
- [ ] If it's a typescript module - add into the root level `package.json` under `workspaces:` so that we can hoist shared dependencies and allow root level installs/builds etc.
- [ ] New module should have a `README.md` documenting what the module does - how to develop for it and any other sort of contribution guidelines.
- [ ] If the new module requires a new label for PR clarity ( https://github.com/finos/architecture-as-code/issues/783 ) - add this into `.github/labeler.yml` and if this is a new label please manually create it on GitHub.
- [ ] Update the docs to include details of the new project for consumers.

rocketstack-matt:
Is there an existing standard we could model this around - or indeed should we look at integrating with an existing standard such that we don't end up needing to frequently update the exchange format enum?

jpgough-ms:
> Is there an existing standard we could model this around - or indeed should we look at integrating with an existing standard such that we don't end up needing to frequently update the exchange format enum?

@rocketstack-matt I wonder if we can use something like the `Content-Type` header approach from HTTP/REST. They have a pretty comprehensive Media Types list, e.g. this https://www.iana.org/assignments/media-types/media-types.xhtml

[...]

PR #1047, opened 2025-03-19
#1046 added interface for data-exchange format

Adwoa-Konadu-Appiah:
Related to #1046 to introduce data-exchange format for interfaces.
@rocketstack-matt @Budlee @jpgough-ms @dc-ms


rocketstack-matt:
See comments on issue for next steps

Issue #1051, opened 2025-03-19
Include Allowed Operations on Interface


Adwoa-Konadu-Appiah:
Description of Problem:
There is no way to describe allowed operations  between two nodes on an interface
Example: Read, Write

Potential Solutions:
We are proposing adding a read/write operation to interfaces to allow users to specify the operation that can be performed on a node using an interface.

The intent of the allowed operations is to show whether a read or write operation is possible on an interface.  Not to be conflated with roles or permissions.

```json
"allowed-operations-interface": {
      "$ref": "#/defs/interface-type",
      "type": "object",
      "properties": {
        "operation": "array",
        "items": {
          "$refs": "#/defs/read-write-operation"
        },
        "uniqueItems": true,
        "minItems": 0,
        "maxItems": 2
      },
      "required": [
        "operation"
      ]
  },
  "read-write-operation": {
    "enum": [
      "read",
      "write"
    ]
  }
```

Adwoa-Konadu-Appiah:
> Clarify the intent . . . is this about what a particular actor / persona can do (i.e. entitlements) or simply what operations are available, regardless of whether anyone has the entitlement to execute.

The issue description has been updated to reflect the intent of the allowed operations, i.e. read or write operations on an interface without being associated with roles or permissions.

rocketstack-matt:
See comment in https://github.com/finos/architecture-as-code/issues/1083

PR #1052, opened 2025-03-19
#1051 added allowed operations to interface

Adwoa-Konadu-Appiah:
Related to #1051 to introduce allowed operations for interfaces.
@rocketstack-matt @Budlee @jpgough-ms @dc-ms


jpgough-ms:
Please work with us to put this through the new schema changes proposal flow.

PR #1082, opened 2025-03-28
Remediate issue with calm docify on 0.6.0 release

LeighFinegold:
The following issue contains foundational changes to move towards ESM. 

It additionally fixes the current issue with the `calm docify` endpoint on the 0.6.0/0.6.1 release by bundling all the code needed for the DocusarusTransformer to work.


Thels (approved):
Spoken with @LeighFinegold  at length, as well as reviewing the changes locally in VSCode on Windows. These changes are built upon what @aidanm3341 and myself did on Wednesday.

I've asked @willosborne to include in the dev container or Flox setup for configuring ViTest extension so that it works natively out of the box. As a follow-up, but this is a great move away from CJS!

Issue #937, opened 2025-02-27
CLI Changes to Support Pattern Options


jpgough-ms:
### Description of Problem:

In #706 the schema has been updated to support pattern options. The goal of pattern options is to:

- Enable pattern authors to describe different options for a particular pattern
- Promote reuse of patterns by allowing controlled customisation
- Power the CLI to enable developer choices at the time of generation using the CLI

The first few points are addressed by the schema change, this issue will address integrating pattern options into the CLI.

#### CALM Validate

Based on `options-prototype.pattern.json` in [prototype](/finos/architecture-as-code/tree/main/calm/draft/2025-03/prototype) validation will fail with errors such as:

```
 calm validate -p calm/draft/2025-03/prototype/options-prototype.pattern.json -m calm/draft/2025-03/meta
 {
        "code": "connects-relationship-references-existing-nodes-in-pattern",
        "severity": "error",
        "message": "'application-b' does not refer to the unique-id of an existing node.",
        "path": "/properties/relationships/prefixItems/0/oneOf/1/properties/relationship-type/const/connects/source/node"
},

```
This is due to the validation rules not looking inside a `OneOf` or `AnyOf` block for nodes and relationships. We should fix the rule to look at these options and ensure that these nodes are included in the JSON path evaluation (assuming that is what is it). 

#### CALM Generate

The generate command should now support the options that are present in the pattern, using the `option-type` to drive what is presented. For example, each option should be presented to the user with the list of possible decisions. e.g.

```
calm generate -p options-prototype.pattern.json

Pattern has options. 

Option 1: The choice of nodes and relationships in the pattern
1) Application A connects to Application C
2) Application B connects to Application C

Choose an option: <wait for input>
```

Based on what is selected the `nodes` and `relationships` in the option relationship should be in the pattern. Patterns can also have an AnyOf series of options, meaning it is a one or more that can be selected.

PR #1112, opened 2025-04-02
Introduce Pattern options #937

aidanm3341:
This PR adds support for Pattern options in the calm generate command.
When a pattern has some options, the user will get an interactive prompt where they can select which option they want, and only the choice that was made will be included in the generated architecture.


aidanm3341:
I see there are some merge conflicts that I'll have to deal with later in the morning

aidanm3341:
I've also noticed a slight oversight on my part, which is that this currently only works for nodes and relationships. It shouldn't be difficult to add for metadata and controls, but I'll add that in as a followup PR.

cli/package.json
```diff
@@ -29,37 +29,37 @@
...
     "@types/xml2js": "^0.4.14",
     "@typescript-eslint/eslint-plugin": "^8.0.0",
     "@typescript-eslint/parser": "^8.15.0",
+    "axios": "^1.7.9",
     "chokidar": "^4.0.1",
     "eslint": "^9.13.0",
     "globals": "^16.0.0",
     "jest": "^29.7.0",
-    "supertest": "^7.0.0",
+    "jest-environment-node": "^29.7.0",
```
willosborne, line 56: We probably don't need this any more.

cli/src/command-helpers/file-input.ts
```diff
@@ -0,0 +1,23 @@
+import * as fs from 'node:fs';
+import { initLogger } from '@finos/calm-shared';
+
+export async function loadJsonFromFile(path: string, debug: boolean): Promise<object> {
+    const logger = initLogger(debug, 'file-input');
+    try {
+        logger.info('Loading json from file: ' + path);
+        const raw = fs.readFileSync(path, 'utf-8');
```
willosborne, line 8: This should probably be await fs.readFile since the method is async

willosborne (changes requested):

shared/src/commands/generate/components/options.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export interface CalmOption {
+    optionType: 'oneOf' | 'anyOf',
+    prompt: string,
+    choices: CalmChoice[],
+}
+
+function isOptionsRelationship(relationship: object): boolean {
+    return 'properties' in relationship && 'properties' in relationship['properties']['relationship-type'] && 'options' in relationship['properties']['relationship-type']['properties'];
```
willosborne, line 17: replace with null coalesing

shared/src/commands/generate/components/options.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function getItemsInOptionsRelationship(optionsRelationship: object): object[] {
+    return optionsRelationship['properties']['relationship-type']['properties']['options']['prefixItems'];
```
willosborne, line 21: can this fail/are we guaranteed to have a well formed options block by the time this is called?

shared/src/commands/generate/components/options.ts
```diff
@@ -0,0 +1,122 @@
...
+                    logger.debug(`Adding [${itemUniqueId}] to [${calmType}]`);
+                    calmItems.push(item);
+                }
+            });
+        });
+
+    logger.debug(`Removing "oneOf" and "anyOf" blocks from ${calmType}`);
+    calmItems
+        .filter((item: object) => 'oneOf' in item || 'anyOf' in item)
+        .forEach((item: object) => calmItems.splice(calmItems.indexOf(item), 1));
```
willosborne, line 77: Can we flip the fliter predicate to avoid having to mutate here

shared/src/commands/generate/components/options.ts
```diff
@@ -0,0 +1,122 @@
...
+function flattenCalmItems(pattern: object, calmType: 'nodes' | 'relationships', ids: string[], debug: boolean = false): void {
+    const logger = initLogger(debug, 'calm-generate-options');
+    logger.debug('Selecting ids [%O] in [%s]', ids, calmType);
+    const calmItems = pattern['properties'][calmType]['prefixItems'];
+
+    calmItems
+        .filter((item: object) => 'oneOf' in item || 'anyOf' in item)
+        .flatMap((item: object) => item['oneOf'] || item['anyOf'])
+        .forEach((item: object) => {
+            ids.forEach((id: string) => {
```
willosborne, line 65: Some points here:
- you are modifying an array while iterating through it. that means that this loop will potentially traverse the newly-added items at the end. if those items contain further anyOf/allOfs, it will keep executing them and could potentially be very confusing indeed. 
- when you hit an anyOf/oneOf, you add it to the end of the list. that means ordering is not preserved, which i think is important given we are representing nodes/relationships as arrays and automated tools may use positional indices to pull out specific nodes/relationships [they probably shouldn't, but json-schema syntax for matching a unique id is horrible.]

I suggest we could restructure this logic down the line - probably not on this PR though.
Here is my suggestion:

- make a function that takes a calm item and returns an array of calm items.
- if the item has no anyOf/oneOf, it just returns itself
- otherwise, it does the selection logic and returns an array containing all the selected items
- then map() over this function for all calm items
- then concat them together.

this is purely functional with no mutation and preserves ordering.

aidanm3341: yeah I like this idea. The mutation approach is a little sloppy in hindsight... We discussed in person that there's also an idea to refactor the logger to use a global one which Leigh has apparently been playing around with. I'll look into these ideas and do as a followup

prettier.config.js
```diff
@@ -0,0 +1,13 @@
+/**
```
oliviajanejohns, line 1: Do we not already have a prettier.config in calm-hub-ui?

oliviajanejohns (approved):

shared/src/commands/generate/components/options.ts
aidanm3341: yeah every time this is called we have already checked if it is an options relationship

willosborne (approved):

prettier.config.js
aidanm3341: yeah true, we can probably simplify to only have one. I'll have a look

Issue #1114, opened 2025-04-02
Changes to Descriptions on CALM Visualizer


oliviajanejohns:
1. Should rename to 'Relationship Descriptions' 
2. Should be turned on by default 

<img width="1141" alt="Image" src="https://github.com/user-attachments/assets/9bf1d76b-6e28-443a-85a3-2d94afd02977" />

Issue #1093, opened 2025-03-31
Add CALM logos to visualizer / CALM Hub


oliviajanejohns:
CALM Hub / CALM UI should make use new themes and logo: 

https://github.com/finos/architecture-as-code/tree/logo/brand/Icon

PR #1122, opened 2025-04-02
Add CALM logos to visualizer / CALM Hub #1093

oliviajanejohns:
https://github.com/finos/architecture-as-code/issues/1114[: Changes to Descriptions on CALM Visualizer](https://github.com/finos/architecture-as-code/commit/a0c684f3a45f53f66b074c897cde52b66babfe54)

Screenshot 2025-04-02 at 23 41 58


jpgough-ms:
@oliviajanejohns / @rocketstack-matt 

Just wondering if we need all the assets over in calm-hub-ui or if we should keep the brand folder?

oliviajanejohns:
Was trying to avoid duplicating these files! Moved to a public folder so can be read within the calm-ui [web](url) application...

aidanm3341:
> Was trying to avoid duplicating these files! Moved to a public folder so can be read within the calm-ui [web](url) application...

hmm I agree avoiding duplication would be good, but I do also think these should still be a top level brand folder for other tools to use. Maybe we could copy the files into public as part of the build or something?

oliviajanejohns:
> > Was trying to avoid duplicating these files! Moved to a public folder so can be read within the calm-ui [web](url) application...
> 
> hmm I agree avoiding duplication would be good, but I do also think these should still be a top level brand folder for other tools to use. Maybe we could copy the files into public as part of the build or something?

Have added script to copy the brand folder when starting calm hub ui FYI

aidanm3341 (approved):

PR #1138, opened 2025-04-03
finos#715 instead of buildtime property, use quarkus runtime property to on/off the OIDC integration

challamani-ms:
- `quarkus.oidc.enabled` is a build-time property. If we depend on this, we cannot override this property value at runtime, and we have to maintain two builds (two docker images), one with OIDC and one without OIDC.
- Instead of maintaining two images, we can turn on/off the ODIC by using a runtime property `quarkus.oidc.tenant-enabled`.


willosborne (approved):

willosborne:
Change looks good, but blocking merge until demos are complete just in case as this could break calmHub.

jpgough-ms (approved):

PR #1150, opened 2025-04-04
Expand nodes to fit description labels

aamanrebello:
Refer to conversation for latest updates.

UI at present:

How it looks on api-gateway

How it looks on traderX

### DIFF
--- a/calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
+++ b/calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
@@ -30,6 +30,8 @@ export type Node = {
         type: string;
         label: string;
         id: string;
+        _displayPlaceholderWithDesc: string;
+        _displayPlaceholderWithoutDesc: string;
         [idx: string]: string;
     };
 };
@@ -106,8 +108,14 @@ export const CytoscapeRenderer = ({
                 {
                     selector: 'node',
                     style: {
+                        label: isNodeDescActive ? 'data(_displayPlaceholderWithDesc)' : 'data(_displayPlaceholderWithoutDesc)',
+                        'text-valign': 'center',
+                        'text-halign': 'center',
+                        'text-wrap': 'wrap',
+                        'text-max-width': '180px',
+                        "font-family": 'Arial',
                         width: '200px',
-                        height: '100px',
+                        height: 'label',
                         shape: 'rectangle',
                     },
                 },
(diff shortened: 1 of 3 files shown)


calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -107,7 +107,7 @@ export const CytoscapeRenderer = ({
                     selector: 'node',
                     style: {
                         width: '200px',
-                        height: '100px',
+                        height: isNodeDescActive ? '250px': '100px',
```
aamanrebello, line 110: I set the height to 250px arbitrarily. This leaves some margin over the biggest description I have encountered in the examples I worked with

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscape.css
```diff
@@ -8,6 +8,11 @@
     background-color: white;
 }
 
+.expanded {
+    top: -4.75rem;
```
aidanm3341, line 12: what's this -4.75 value for?

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
aidanm3341: hmm, there's got to be a way to do this properly dynamically. Maybe along the lines something like this? 

```typescript
const el = document.getElementById("myDiv"); // ideally we could do this with some reference rather than just accessing the dom
if(el) {
  const height = elem.getBoundingClientRect().height;
}
```

aamanrebello: I'll look into it. I tried doing it with css and that was hard, but this way might do the trick.

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscape.css
aamanrebello: It's to make sure the label completely covers the node. As I increase the height of the node/label, I also have had to increase the value of `top` because the mis-alignment of the label (position absolute) and the node goes on increasing. So when I remove the top attribute, I get this:



I tried setting valign: 'top' on the node HTML labels in the code, but that also resulted in a misalignment:



I suppose I could set the adjustment dynamically based on the label height, using concepts discussed in your other comment.

aamanrebello: So I have made an attempt at allowing a dynamic nodel label sizing, and also removed the need for this top adjustment.

PR #1152, opened 2025-04-07
stop constant rerendering of graph when zoom changes

aidanm3341:


oliviajanejohns (approved):
approved - but agree not a fan of this solution! Needs refactoring

oliviajanejohns:
> @oliviajanejohns / @aidanm3341 should we decline this PR and do the proper fix? I'll work around it in my demo tomorrow.

if happy to work around this.. then yes I think lets decline

aidanm3341:
given we no longer need to put in this quick fix, we're closing this so we can fix it in a better way

PR #1156, opened 2025-04-07
1155: Node Description / Relationship description toggles not displaying correctly

oliviajanejohns:


jpgough-ms (approved):
Approving, but also thinking it would be nice to get some tests on these files. @aidanm3341 / @oliviajanejohns perhaps we can discuss at an office hours?

Issue #1088, opened 2025-03-31
Rendering Positions of Nodes on First Load


jpgough-ms:
### Steps to Reproduce:

Create an architecture which uses a contains relationship and load into the CalmHub

<img width="621" alt="Image" src="https://github.com/user-attachments/assets/d24d5b3f-194e-470c-bb69-1cc6614c4279" />

* Nodes that are not in the contains relationship are rendered in the contains relationship
* The box doesn't quite contain the nodes initially
* If I turn on descriptions, everything re-renders and doesn't resize for the new text.

### Expected Result:

* Nodes not in the relationship are outside the box
* Box should contain all nodes
* Descriptions should remain within the box after a node resize

<img width="767" alt="Image" src="https://github.com/user-attachments/assets/484687c8-3603-4c0c-b537-42d3401b9e2f" />

jpgough-ms:
Example of turning on descriptions and overlapping content containers.

<img width="761" alt="Image" src="https://github.com/user-attachments/assets/ec2bb282-1d41-40f0-8b0c-e892277e33d2" />

aamanrebello:
In reply to overlapping content containers, what do we think of this UI:

(when on)

![Image](https://github.com/user-attachments/assets/4b01fcfc-e2b8-4560-bfc1-0fcac6f8f673)

(when off)

![Image](https://github.com/user-attachments/assets/08d1d75e-a2d4-4683-80bf-26c62cbcbd43)

alternatively I can get the boxes to stay smaller when the descriptions are not turned on

aidanm3341:
looking good @aamanrebello! I think it would be better if the nodes shrank to be smaller when the descriptions are turned off. I imagine the main reason for people doing this would be to make the diagram smaller

aamanrebello:
Ok so part 1 is up where containers expand with the nodes: [https://github.com/finos/architecture-as-code/pull/1150](https://github.com/finos/architecture-as-code/pull/1150). I have tested it with traderX and api gateway, I don't know where I can access the example that James Gough is using (will that be used in the demo?)

For the second part, it will be important to identify which nodes are outside the "deployed in" or "composed of" relationship - a quick fix I can do is "force" them to initially be in some position which I know will be outside the container e.g. something like x=10, y=10 (or something better looking than that).

Will get it done hopefully by Saturday, but I am more than happy if other people think of something and get there first :)

aamanrebello:
[...]

PR #1157, opened 2025-04-08
Display external nodes outside the system

aamanrebello:
In response to issue https://github.com/finos/architecture-as-code/issues/1088. Useful context in comment: https://github.com/finos/architecture-as-code/issues/1088#issuecomment-2785005403

This arrangement will hopefully allow us to keep nodes external to a system (consisting of contains relationships) outside it in the visualizer. Consider the below example:

image

UPDATE: I have created a layout correction service that looks over the entire graph, identifies nodes that have to be moved and moves them such that:
- no node will be inside a non-parent
- nodes will render inside their parents
- a moved node will not intersect with another non-parent node
- The distance the node has to move is kept low while ensuring nodes are nicely spaced out

whichever layout we utilise down the line, we will be able to use this service to ensure that the above rules are respected. We can even extend it to enforce more rules.

image

Currently I have the traderX example to test with, as well as 6 unit tests. These do test the capabilities of the service, but it would be cool if we had more CALM schemas to test with.


calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -124,9 +125,25 @@ export function Drawer({ calmInstance, title, isConDescActive, isNodeDescActive
...
             return newData;
         });
 
+        if(groupsExist) {
+            nodes.push({
+                classes: 'group',
+                data: {
+                    label: 'External World',
+                    description: 'Area representing entities and actors external to the system',
+                    type: 'domain',
```
aamanrebello, line 142: Wondering if "domain" is the correct name to call it.

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -124,6 +126,14 @@ export const CytoscapeRenderer = ({
             maxZoom: 5,
         });
 
+        //Display external nodes outside the system
+        let yPosition = 0;
+        externalNodeIds.forEach((id: string) => {
+            const element = updatedCy.getElementById(id);
+            element.position({x: -1000, y: yPosition});
+            yPosition += 2*element.height();
```
aamanrebello, line 134: This is arbitrary to allow the external nodes to be spaced out on the left.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -124,10 +126,15 @@ export function Drawer({ calmInstance, title, isConDescActive, isNodeDescActive
             ) {
                 newData.data.parent = deployedInRelationships[node['unique-id']].parent!;
             }
+
+            if(newData.data.parent == null && newData.classes !== 'group' && groupsExist) {
```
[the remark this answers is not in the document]
aamanrebello: Will look into both points, hopefully today.

aamanrebello:
I have an idea on how to do this. I'm thinking of creating a layout correction service. Essentially, it looks over all the node positions and bounding boxes to check that "within parent" constraints are satisfied. If any node does not satisfy conditions, it can calculate which direction and how far to move the node in order to make the diagram "correct".

By default we would have the breadth first layout, and then apply correction on top of it to see if any nodes need to be moved. I'll try to get this done by Friday, I'm struggling to find time :D

calm-hub-ui/src/tests/Sidebar.test.tsx
```diff
@@ -11,7 +11,7 @@ describe('Sidebar Component', () => {
         label: 'Node 1',
         type: 'type-1',
         description: 'Mock Node',
-    };
+    } as Node['data'];
```
aamanrebello, line 14: This is just a type safety change.

calm-hub-ui/src/visualizer/contracts/contracts.tsx
```diff
@@ -0,0 +1,42 @@
+import cytoscape from "cytoscape";
```
aamanrebello, line 1: Moved all types to this file for better organisation.

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -167,7 +149,7 @@ export const CytoscapeRenderer = ({
                 tpl: getNodeLabelTemplateGenerator(true),
             },
         ]);
-
+        layoutCorrectionService.calculateAndUpdateNodePositions(updatedCy, nodes);
```
aamanrebello, line 152: This calls the layout correction service.

calm-hub-ui/src/visualizer/services/layout-correction-service.tsx
```diff
@@ -0,0 +1,352 @@
...
+    }
+    
+    /**
+     * Identifies nodes that are not rendered within parents (or grandparents and higher ancestors), as well
+     * as nodes rendered within non-ancestors.
+     * @param cyRef Reference to cytoscape graph.
+     * @param nodes Nodes to search over.
+     * @returns An object describing nodes not contained within ancestors, as well as nodes within non-ancestors, along with bounding boxes
+     */
+    private identifyNodesToBeMoved(cyRef: cytoscape.Core, nodes: Node[]): Map<string, NodeLayoutViolations> {
```
aamanrebello, line 141: This method goes through the graph and identifies nodes whose positions need to be updated.

calm-hub-ui/src/visualizer/services/layout-correction-service.tsx
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Based on node violations, suggests a corrected position for a violating node. It finds allowed
+     * positions for the node and suggests one of these allowed positions
+     * @param violations A on object containing the bounding box of the node and related nodes for which its rendered position
+     * violates the relationship 
+     * @returns The suggested position as an object.
+     */
+    private getCorrectedPosition(cyref: cytoscape.Core, nodeId: string, violations: NodeLayoutViolations): cytoscape.Position {
```
aamanrebello, line 319: The logic of this method is that it initially assumes the node can be anywhere on the grid i.e. range negative infinity to infinity. It then updates the ranges based on containers the node should be inside + the node cannot intersect with any other node.

Based on the allowed positions of the node, it returns a position selected from the allowed ranges.

calm-hub-ui/src/visualizer/helpers/set-functions.tsx
aidanm3341, line 1: this doesn't need to be `.tsx`, just `.ts`. Same with the test

calm-hub-ui/src/tests/set-functions.test.tsx
aidanm3341, line 1: I think we should follow the format of having tests co-located with the implementation. This is how most UI projects that I've seen work and I prefer it personally

calm-hub-ui/src/visualizer/services/layout-correction-service.tsx
```diff
@@ -0,0 +1,352 @@
...
+import { BoundingBox, IdAndBoundingBox, Node, NodeLayoutViolations } from '../contracts/contracts.js';
+import { difference, intersection, union } from "../helpers/set-functions.js";
+
+export class LayoutCorrectionService {
+    
+    //PUBLIC METHODS-------------------------
+
+    public calculateAndUpdateNodePositions(cyRef: cytoscape.Core, nodes: Node[]): void {
+        this.identifyNodesToBeMoved(cyRef, nodes);
+        //Make sure all parentless containers are correctly placed first, followed by child cotaimers of these parents
```
aidanm3341, line 11:
```suggestion
        //Make sure all parentless containers are correctly placed first, followed by child containers of these parents
```

calm-hub-ui/src/tests/layout-correction-service.test.tsx
```diff
@@ -0,0 +1,448 @@
...
+                data: {
+                    description: "Node 3 description",
+                    type: "type3",
+                    label: "Node 3",
+                    id: "node3",
+                    parent: "node1",
+                    _displayPlaceholderWithDesc: "Display 3 with desc",
+                    _displayPlaceholderWithoutDesc: "Display 3 without desc",
+                    extraField: "extraValue3",
+                },
```
aidanm3341, line 77: I reckon there's something we can do reduce duplication with these nodes... with this setup, it's pretty hard to see what's different between each test. What if we defined them globally at the top

```typescript
const node1 = {
classes: "class1",
    data: {
         description: "Node 1 description",
         type: "type1",
         label: "Node 1",
         id: "node1",
         _displayPlaceholderWithDesc: "Display 1 with desc",
        _displayPlaceholderWithoutDesc: "Display 1 without desc",
        extraField: "extraValue1",
    },
}
```
and then in the tests we can reference it like this to introduce the test specific changes:

```typescript
const nodes = [
    {
        ...node1,
        parent: "node2"
    },
    node2,
    node3
]
```

aamanrebello: Makes perfect sense, I'll see if I can simplify the bounding boxes as well.

calm-hub-ui/src/tests/set-functions.test.tsx
aamanrebello: Let's do this in another PR. This one is already quite complex.

calm-hub-ui/src/tests/layout-correction-service.test.tsx
aamanrebello: Made an attempt at simplifying the repetitions.

calm-hub-ui/src/visualizer/helpers/set-functions.tsx
aamanrebello: Changed.

calm-hub-ui/src/visualizer/helpers/set-functions.ts
oliviajanejohns, line 1: Unsure why these are needed. Can you explain?

aamanrebello: So my builds were failing earlier when I used the Typescript `Set` API. For example, over here: https://github.com/finos/architecture-as-code/actions/runs/14686442242/job/41215644940, the build was not accepting the `Set.difference` function. 

On looking into the issue online, the reason seems to be that the version of Node we are using does not support TypeScript sets: https://github.com/microsoft/TypeScript/issues/60012, https://www.webdevtutor.net/blog/typescript-set-difference-is-not-a-function, https://github.com/vitejs/vite/discussions/17750.

I need those set operations though, so took the liberty of defining them myself. If we are able to resolve the issue and recognise the Set API, I am happy to get rid of these.

aidanm3341: these set methods are part of the latest ECMAScript specification that's due to release this summer https://github.com/tc39/proposals/blob/main/finished-proposals.md 

So I think having these manual definitions is fine for now, we can hopefully remove them when the new version of ECMAScript is properly supported

calm-hub-ui/src/tests/Sidebar.test.tsx
oliviajanejohns: Should this not be "CalmNode"?

aamanrebello: Yeah, updating the tests as well.

oliviajanejohns: Sorry that was my doing. Basically I added additional properties and felt it made more sense to rename so we knew what a cystoscope Node was and our own to not confuse

aidanm3341 (approved):

Issue #275, opened 2024-06-05
Deprecation Warning on CLI


aidanm3341:
When installing the CLI from NPM, there are some warnings that are raised:

```shell
> npm install -g @finos/calm-cli
npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm WARN deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
npm WARN deprecated querystring@0.2.1: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead

changed 332 packages in 17s

81 packages are looking for funding
  run `npm fund` for details
```

We should track down the package that is causing these and try to update or find an alternative to get address the warnings

rocketstack-matt:
I was looking into the deprecation warnings . . . I think most are caused by the version of fetc-mock we're using; now I'm able to upgrade the version without too much effort but a bunch of our tests start failing . . . however I think this is actually highlighting that they're not working today.
 
For example, a bunch of the validation tests which pass today fail, but crucially if I run the validate command form the CLI against the same fixtures, it does indeed fail.

Thels:
Hope you don't mind - as one of the last tasks I'll pick up in 2024 I also had a peek at this. I've got a PR up which has some CVE issues but solves the fetch-mock problems you faced.

Thels:
Sat down to work on this task an hour or so ago - and are struggling to get a cloned fresh copy of the repository behaving. Until I have a working local copy this will be paused.

PR #1163, opened 2025-04-10
#275 - Attempt at upgrading versions to remove deprecation warnings.

Thels:


jpgough-ms:
@Thels looks like some updates I've been doing have caused a conflict on the `package-lock.json`

Issue #1159, opened 2025-04-09
Change the  Winston Logger  Library in the Shared Model to enable both Node.js and Browser capabiltities


Adwoa-Konadu-Appiah:
Issue
The current logging setup for the shared model is only compatible with Node.js. As a result, any model imported into the Visualizer fails, causing the entire Visualizer to break.

Suggested Solution:
Replace the current logging library with loglevel, a lightweight library that supports logging in Node.js and browser environments. 
This will ensure consistent functionality across both platforms and prevent the Visualizer from breaking.

Thels:
I don't think this is a bug - it's a change we want to make for a legitimate reason.

Adwoa-Konadu-Appiah:
[Change Logger for Node Js and Browser Compatibility #1186](https://github.com/finos/architecture-as-code/pull/1186)

PR reviewed and merged

PR #1186, opened 2025-04-16
Change Logger for Node Js and Browser Compatability

Adwoa-Konadu-Appiah:
In reference to #1159,  I have updated the logger to use the right logging model for Node.js and browser using the loglevel library.


rocketstack-matt:
@Adwoa-Konadu-Appiah please check you have the package-lock file synced.

shared/src/logger.ts
```diff
@@ -1,25 +1,112 @@
...
+/**
+ * Initializes a logger that works in both Node.js and browser environments.
+ * @param debug - Enables debug logging if true.
+ * @param label - Optional label to prefix Node.js logs.
+ * @returns Logger instance
+ */
+export async function initLogger(
+    debug: boolean,
+    label?: string
+): Promise<Logger> {
```
aidanm3341, line 10: where does the type for this `Logger` come from? I'm not sure I understand it

shared/src/parser/parser.ts
```diff
@@ -16,8 +16,8 @@ export class CalmParser {
             dereferencedData.controls = {}; // If this ends up being string documents then this will break CalmControl.fromJson
             return CalmCore.fromJson(dereferencedData);
         } catch (error) {
-            logger.error('Failed to parse calm.json:', error);
+            logger.then(l => (l.log(l.ERROR,`Failed to parse calm.json:' ${error}`)));
```
aidanm3341, line 19: hmm, I'm not a fan of how verbose this is. `logger.then(l => (l.log(l.ERROR,...` is a lot to have to type/read just to do a simple log. I suppose this is because this is a non-async context? Is there any way that we can make the initLogger function synchronous so we won't have this problem?

shared/src/schema-directory.spec.ts
```diff
@@ -54,55 +66,57 @@ describe('SchemaDirectory', () => {
...
         expect(interfaceDef.properties).toHaveProperty('host');
         expect(interfaceDef.properties).toHaveProperty('port');
         expect(interfaceDef.properties).toHaveProperty('unique-id');
     });
-    
+
     it('qualify relative references within same file to absolute IDs', async () => {
-        const schemaDir = new SchemaDirectory();
-        
+        // const schemaDir = new SchemaDirectory();
```
aidanm3341, line 95: why are these schemaDirs commented out?

shared/src/commands/validate/validate.ts
```diff
@@ -225,7 +224,7 @@ async function validatePatternOnly(jsonSchemaLocation: string, metaSchemaPath: s
  * @returns the validation outcome with the results of the spectral validation.
  */
 async function validateArchitectureOnly(architectureSchemaLocation: string): Promise<ValidationOutcome> {
-    logger.debug('Pattern was not provided, only the Architecture will be validated');
+    logger.log(logger.DEBUG,'Pattern was not provided, only the Architecture will be validated');
```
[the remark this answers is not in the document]
Adwoa-Konadu-Appiah: I have updated the code to call the logger the way it was before. The code is now ready for review

aidanm3341 (approved):

shared/src/logger.ts
```diff
@@ -1,25 +1,90 @@
...
+let loggerInstance: Logger;
+
+/**
+ * Initializes a logger that works in both Node.js and browser environments.
+ * @param debug - Enables debug logging if true.
+ * @param label - Optional label to prefix Node.js logs.
+ * @returns Logger instance
+ */
+export function initLogger(debug: boolean, label?: string): Logger {
+    if (loggerInstance) return loggerInstance;
```
LeighFinegold, line 23: Introducing state means the same format log line is used for all of the cli e.g. 0.7.8 of the cli it seems to be info [_TemplateEngine]:

Issue #1083, opened 2025-03-28
Change interfaces to be user extensible


rocketstack-matt:
### Target Schema:

- `core.json`
- `interface.json` (structure redefined)

### Description of Change:

Interfaces are currently statically defined within the CALM meta-schema (`interface.json`), requiring core schema changes to introduce new types. This limits extensibility and creates tight coupling between the core schema and specific interface definitions.

This proposal introduces a modular interface model, similar to `control-requirement`, where each interface instance includes:

- `interface-definition-url`: a URI pointing to the external JSON Schema that defines the interface-type.
- `configuration`: an inline object that conforms to the above schema.

To ensure backward compatibility, the new schema allows interface instances to be either a legacy `interface-type` or the new `interface-definition`, which bundles the more loosely coupled reference to the `interface-type` URL and the configuration which should match the linked type.

### Use Cases:

- Define and validate domain-specific interfaces (e.g., Kafka, MQTT, gRPC) without modifying the CALM core schema.
- Version interface types independently and host externally.
- Allow validation of inline interface configurations against extensible schemas.
- Reduce the need to change CALM’s core schema for new interface types.

### Current Limitations:

- Interface types are hardcoded in `interface.json`.
- No support for external schemas or domain-specific extension.
- Cannot reuse or version interface schemas independently.

### Proposed Schema Changes:

The new wrapper type `interface-definition` or the legacy `interface-type` can be referenced for backwards compatibility.
```json
"interfaces": {
  "type": "array",
  "items": {
    "anyOf": [
      { "$ref": "interface.json#/defs/interface-definition" },
      { "$ref": "interface.json#/defs/interface-type" }
    ]
  }
}
```

```json
{
  "defs": {
    "interface-definition": {
      "$ref": "https://calm.finos.org/draft/2025-03/meta/interface.json#interface-type",
      "type": "object",
      "description": "A modular interface definition referencing an external schema",
      "properties": {
        "unique-id": {
          "type": "string",
          "description": "Unique identifier for this interface instance"
        },
        "interface-definition-url": {
          "type": "string",
          "format": "uri",
          "description": "URI of the external schema this interface configuration conforms to"
        },
        "configuration": {
          "type": "object",
          "description": "Inline configuration conforming to the external interface schema"
        }
      },
      "required": ["unique-id", "interface-definition-url", "configuration"],
      "additionalProperties": false
    },

    "interface-type": {
      "type": "object",
      "properties": {
        "unique-id": { "type": "string" }
[...]

Issue #893, opened 2025-02-10
CALM Schema Governance Proposal


rocketstack-matt:
# **CALM JSON Meta Schema Governance Process**

## **1. Governance Overview**
The CALM JSON Meta Schema must be governed through a structured process to ensure:
- Stability in released versions.
- Transparency and traceability of changes.
- Compatibility with CALM tooling (CLI, CALM Hub UI, etc.).
- Community involvement and review.
- Controlled adoption of schema updates in minor and major releases.

---

### **a) Drafting New Schema Changes**
- Proposed schema changes are made in the `drafts/#issue-number` folder.
- This format enables schema changes to be proposed independently and in parallel.
- It also avoids confusion from draft proposals not being completed in the month they were created.
- Changes to draft schemas are allowed without restrictions to encourage iteration and innovation.
- Examples of using the change should be provided in the drafts/#issue-number/prototype folder.

### **b) Change Proposals & Discussion**
- Schema changes should be proposed via **GitHub Issues** and subsequent related **Pull Requests (PRs)** in the [FINOS Architecture as Code monorepo](https://github.com/finos/architecture-as-code).
- Schema changes should use the [Schema Change Proposal template](https://github.com/finos/architecture-as-code/issues/new?template=Schema_change_proposal.md).

### **c) Review & Approval**
- Each PR undergoes **community review** with at least one **calm-schema-governance** approver.
  - Once an approval is given the PR can be merged, because it will not impact any existing tools or inflight work, but being accepted as a draft is no guarantee that it will make it into the schema itself.
- The calm-schema-governance group may request:
  - Additional discussion via FINOS working group meetings.
  - Live demonstration of impact on tools (e.g., CLI, CALM Hub UI).

---

### **a) Minor & Major Version Adoption**
- Minor and major version releases can be proposed at any time by the **calm-schema-governance** group.
- The selected draft undergoes **full validation**:
  - Schema linting and validation.
  - Compatibility testing with CALM tools.
  - Backward-compatibility checks.

### **b) Release Candidate Process**
- Once validated, a **Release Candidate (RC)** (e.g., `v1.2-rc1`) is published in a dedicated `releases/` folder.
- The **community tests the RC** for **four weeks** before finalization.
- If no major issues are found, the **RC is promoted** to an official minor or major release.
- If tool maintainers do not reply in the period, it will be at the discretion of the **calm-schema-governance** group to determine if the tool is considered to support the change or to deprecate support of the tool.

[...]

PR #1195, opened 2025-04-18
Proposal to implement #1083

rocketstack-matt:
Note: This schema change is following the proposal for #893

### DIFF
--- a/calm/draft/1083/meta/interface.json
+++ b/calm/draft/1083/meta/interface.json
@@ -1,0 +1,202 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://calm.finos.org/draft/1083/meta/interface.json",
+  "title": "Common Architecture Language Model Interfaces",
+  "defs": {
+    "interface-definition": {
+      "type": "object",
+      "description": "A modular interface definition referencing an external schema",
+      "properties": {
+        "unique-id": {
+          "type": "string",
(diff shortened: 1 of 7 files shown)


jpgough-ms:
More a comment on creating schema proposals, it would be helpful to create the current structure in a first PR - so that the diffs of the change can be viewed more precisely.

calm/draft/1083/meta/interface.json
```diff
@@ -0,0 +1,202 @@
...
+          "items": {
+            "type": "string"
+          }
+        }
+      },
+      "required": [
+        "node"
+      ]
+    },
+    "host-port-interface": {
```
jpgough-ms, line 55: Although not backwards compatible, my preference is we remove the rest of this file from this point down. Folks using these already can bring them into their own projects until we have a more permanent location for them in calm hub.

jpgough-ms (changes requested):
Overall, think this looks good. I've asked a few questions that would be good to clarify before approving

rocketstack-matt:
> More a comment on creating schema proposals, it would be helpful to create the current structure in a first PR - so that the diffs of the change can be viewed more precisely.

Agreed, I realised this when trying to review it myself.

calm/draft/1083/meta/interface.json
rocketstack-matt: I'm good with this, have removed.

jpgough-ms (approved):

Issue #887, opened 2025-02-06
Retrieve documents from CALMHub in the CLI


willosborne:
### Description of Problem:
Currently the CLI can only work with the local filesystem, and doesn't talk to CALMHub.
There should be a mode where it's configured to point at a CALMHub instance and patterns, schemas, etc are retrieved with the REST API.

Note that this is only about **pulling** data - we should also consider how the CLI will be used to upload patterns, architectures, etc into CALMHub but this will come later.

### Potential Solutions:
Refactor the SchemaDirectory to have two implementations - a file system implementation that is the same as what we have today, and one that uses CALMHub.

Add an option to pass the CALMHub URL. Also potentially add the option to configure this in an init/rc file in the user's home directory. For example, `~/.calm.json`

willosborne:
Currently I have a working spike of this for the `generate` command. 
Some changes to the CALMHub REST API will need to be made, we are discussing these today on the Office Hours.

willosborne:
First step - resolve references from CALMHub.
i.e. you need to pass it a file to validate or generate from, but any refs are resolved against it.

Next 
- let you pass a calmhub url for generate
- resolve controls/interfaces

willosborne:
Locally I now have the ability to resolve refs and also generate from a calmhub pattern. 
Generate command only.

PR #1209, opened 2025-04-25
Refactor CLI and SchemaDirectory to use DocumentLoader to enable multiple loading strategies

willosborne:
Key step towards #887 . A second attempt at #897 which got hit by such severe merge conflicts I had to start again.
This should finally introduce a new abstraction called a DocumentLoader that is responsible for fetching the schemas requested by SchemaDirectory. DocumentLoaders have two lifecycle stages:
- initialisation - when the CLI starts. This is useful for up-front loading e.g. the recursive file system loader.
- load missing schemas - if something isn't currently in memory, this will load it. This will be the primary mode of operation for CalmHub down the line.

Initially there is only one implementation, FileSystemDocumentLoader, but on the hackday next week I will be looking at adding a CALMHub implementation. This implementation can only load things at startup, and will fail if you try to load something that's missing.

Most of the pain here was fixing tests. Also the CLI server had to change somewhat, since this now throws exceptions where previously it just returned {}.
Also, this makes a lot of things async that used to be simple operations since everything was assumed to be loaded already.


shared/src/document-loader/document-loader.ts
```diff
@@ -0,0 +1,55 @@
...
+export type CalmDocumentType = 'architecture' | 'pattern' | 'schema';
+
+export interface DocumentLoader {
+    initialise(schemaDirectory: SchemaDirectory): Promise<void>;
+    loadMissingDocument(documentId: string, type: CalmDocumentType): Promise<object>;
+}
+
+export type DocumentLoadMode = 'filesystem' | 'calmhub';
+
+export interface DocumentLoaderOptions {
```
aidanm3341, line 14: a nifty thing you can do here is a technique called Type Tagging. 

Currently you have these optional fields for `schemaDirectoryPath` and `calmHubUrl`, but I can see that we want a constraint that if the loadMode is 'filesystem' then we expect the `schemaDirectoryPath` to exist right? 

So you could instead define the types like this:

```typescript
type FileSystemLoaderOptions = {
    loadMode: 'filesystem',
    schemaDirectoryPath: string
}
type CalmHubLoaderOptions = {
    loadMode: 'calmhub',
    calmHubUrl: string
}
export type DocumentLoaderOptions = FileSystemLoaderOptions | CalmHubLoaderOptions;
```

This gives the caller of the function better type checking:

```typescript
buildDocumentLoader({
    loadMode: 'filesystem',
    schemaDirectoryPath: 'whatever', // <-- having this field was enforced by the type checker!
    calmHubUrl: 'can I do this?' // no! This is a type error
});
```

It also makes your function a little nicer to implement, because you don't need to do 'if (docLoaderOpts.schemaDirectoryPath)` anymore because the type checker _knows_ through type narrowing that if the loadMode is filesystem, then the schemaDirectoryPath must exist, so it could instead look like this:

```typescript
export function buildDocumentLoader(docLoaderOpts: DocumentLoaderOptions, debug: boolean): DocumentLoader {
    switch(docLoaderOpts.loadMode) {
        case 'filesystem': {
            const directoryPaths = [CALM_META_SCHEMA_DIRECTORY];
            directoryPaths.push(docLoaderOpts.schemaDirectoryPath);
            return new FileSystemDocumentLoader(directoryPaths, debug);
        } 
        case 'calmhub': {
            throw new Error('CalmHub loading isn\'t supported yet');
        }
    }
}
```

this is one of my favorite parts of TypeScript 😃

cli/src/cli.ts
```diff
@@ -100,3 +108,28 @@ export function setupCLI(program: Command) {
...
+        schemaDirectoryPath: undefined
+    };
+}
+
+async function buildSchemaDirectory(options: DocumentLoaderOptions, debug: boolean): Promise<SchemaDirectory> {
+    const docLoader = buildDocumentLoader(options, debug);
+    return new SchemaDirectory(docLoader, debug);
+}
+
+function buildDocumentLoader(options: DocumentLoaderOptions, debug: boolean) : DocumentLoader {
```
aidanm3341, line 131: I don't understand how this function type-checks... The return type is `DocumentLoader` but if `options.loadMode` is NOT `filesystem`, won't it return undefined?

cli/src/server/routes/validation-route.ts
```diff
@@ -43,12 +44,22 @@ export class ValidationRouter {
         if (!schema) {
             return res.status(400).type('json').send(new ErrorResponse('The "$schema" field is missing from the request body'));
         }
-        this.logger.info('Path loading schemas is ' + this.schemaDirectoryPath);
 
-        await this.schemaDirectory.loadSchemas(this.schemaDirectoryPath);
-        const foundSchema = this.schemaDirectory.getSchema(schema);
-        if (!foundSchema) {
-            return res.status(400).type('json').send(new ErrorResponse('The "$schema" field referenced is not available to the server'));
+        console.log('loading schemas');
```
aidanm3341, line 48: console.log? Are these for when you were debugging? Or did you intend for these to be actual logs, in which case should be using Winston

shared/src/schema-directory.spec.ts
```diff
@@ -14,116 +15,137 @@ vi.mock('./logger', () => {
...
-            'https://calm.finos.org/draft/2024-10/meta/units.json'
-        ];
-
-        expectedFileNames.forEach((fileName, index) => {
-            expect(loadedSchemas[index]).toBe(fileName);
-        });
+        await schemaDir.loadSchemas();
+
+        const expectedValue = {'$id': 'abcd'};
+        mockDocLoader.loadMissingDocument.mockReturnValueOnce(new Promise(resolve => resolve(expectedValue)));
```
aidanm3341, line 46: `.mockResolvedValueOnce(...)` exists

shared/src/schema-directory.ts
```diff
@@ -1,77 +1,53 @@
...
+        await this.documentLoader.initialise(this);
     }
 
-    private lookupDefinition(schemaId: string, ref: string) {
-        const schema = this.getSchema(schemaId);
+    private async lookupDefinition(schemaId: string, ref: string): Promise<object> {
+        const schema = await this.getSchema(schemaId);
         if (!schema) {
-            return undefined;
+            this.logger.warn(`Schema with $id ${schemaId} not found. Returning empty object.`);
```
aidanm3341, line 44: log line says returning empty object, but we're returning a placeholder object with actual content right?

cli/src/server/routes/validation-route.ts
willosborne: missed testing line, will remove

shared/src/schema-directory.ts
willosborne: Yep. It's a warning that gets returned, will update

shared/src/document-loader/document-loader.ts
```diff
@@ -0,0 +1,62 @@
...
+    loadMode: 'calmhub';
+    calmHubUrl: string;
+}
+export type DocumentLoaderOptions = FileSystemLoaderOptions | CalmHubLoaderOptions;
+
+export function buildDocumentLoader(docLoaderOpts: DocumentLoaderOptions, debug: boolean): DocumentLoader {
+    switch(docLoaderOpts.loadMode) {
+    case 'filesystem': {
+        const directoryPaths = [CALM_META_SCHEMA_DIRECTORY];
+        if (docLoaderOpts.schemaDirectoryPath) {
```
aidanm3341, line 28: you don't need this if statement because of the changes to the types. But not worth blocking the PR for...

aidanm3341 (approved):

shared/src/document-loader/document-loader.ts
willosborne: Technically I do need this right now because the user can omit this option. But i think we should make the types indicate this. Will do later because it isn't an issue right now and breaks a lot of things - will need it for the stricty mode change

Issue #713, opened 2024-12-31
CalmHub: Introduce Controls 


jpgough-ms:
### Description of Problem:

Calm Hub currently supports architectures, patterns, and schemas. #711 proposes introducing flows. 

To support the use of controls, this feature will add controls to CalmHub. The functionality will be a little different to the other functionality, as this should store both `control-requirement` and `control-configuration`. There is still some work going on in this space, with issue #687. This issue will need to remain blocked until then.

### Potential Solutions:

* Share the proposed URL structure on this issue for agreement for the best way to store controls requirements and configurations
* Implementation with unit tests
* Addition of integration tests

Assigning to @jpgough-ms to keep track of dependencies

rocketstack-matt:
CC: @LeighFinegold

jpgough-ms:
### Provisional Design

#### API Structure

Domains

```rest
GET     calm/controls/domains - List all the domains supported by the CALM Hub
POST    calm/controls/domains - Create a new Domain (body)
```

Common schemas used by this domain and potentially shared with other domains

```rest
GET     calm/controls/domains/{domain}/schemas - returns a list of SchemaResponse
GET     calm/controls/domains/{domain}/schemas/{id} - returns a JSON Schema
POST     calm/controls/domains/{domain}/schemas - create a Schema, return a 201 with id and SchemaResponse
```

Managing Controls

```rest
GET     calm/controls/domains/{domain}/controls - return a list of Control
GET     calm/controls/domains/{domain}/controls/{id} - return a Control JSON
GET     calm/controls/domains/{domain}/controls/{id}/requirements - return a list of ControlRequirement
GET     calm/controls/domains/{domain}/controls/{id}/requirements{id} - return a list of ControlRequirement JSON

POST     calm/controls/domains/{domain}/controls?name=&description= - create a new control, body json
POST     calm/controls/domains/{domain}/controls/{id}/requirements?name=&description= - create a new ControlRequirement
```

#### Model Objects

`Domain` - name, description, external links
`Schema` - id, name
`Control` - id, name, description
`ControlRequirement` - id, name, description

jpgough-ms:
Whilst reviewing the schemas section of this issue with @rocketstack-matt, we think that this is at the wrong level and there is perhaps a better way that we could model this.

This was the previous proposal:

```
GET     calm/controls/domains/{domain}/schemas - returns a list of SchemaResponse
[...]

PR #1226, opened 2025-04-29
Introduce domains for controls

jpgough-ms:


jpgough-ms:
Begins to add features to #713

package-lock.json
grahampacker-ms, line 1: I don't think this needed to be changed?

jpgough-ms:
@grahampacker-ms need a re-review on this when you have a second please.

grahampacker-ms (approved):

PR #1227, opened 2025-04-29
Add CLI support for loading documents from CalmHub for generate

willosborne:
- Adds calmhub document loader
- Adds ability to configure calmhub on the cli
- Adds user config file in ~/.calm.json
- Adds tests for document loaders


cli/src/cli-config.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface CLIConfig {
+    calmHubUrl?: string
+}
+
+function getUserConfigLocation(): string {
+    const homeDir = homedir();
+    return join(homeDir, '.calm.json');
+}
+
+export async function loadCliConfig(): Promise<CLIConfig | null> {
```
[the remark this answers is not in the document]
willosborne: updated to always return null

shared/src/document-loader/calmhub-document-loader.ts
```diff
@@ -0,0 +1,62 @@
+import axios, { Axios } from 'axios';
+import { SchemaDirectory } from '../schema-directory';
+import { CalmDocumentType, DocumentLoader } from './document-loader';
+import { Logger } from 'winston';
+import { initLogger } from '../logger';
+
+export class CalmHubDocumentLoader implements DocumentLoader {
+    private readonly ax: Axios;
```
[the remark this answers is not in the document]
willosborne: Spoke to aidan about this at the time and his conclusion is we should be using it over fetch in other places.
Fetch is very minimal and much harder to configure and mock than axios, since you aren't constructing an instance. It's all in-memory. There are also potential compatibility issues between node and the browser, whereas axios should be compatible between the two.

cli/src/cli.ts
```diff
@@ -109,20 +116,47 @@ export function setupCLI(program: Command) {
...
     };
 }
 
-async function buildSchemaDirectory(options: DocumentLoaderOptions, debug: boolean): Promise<SchemaDirectory> {
-    const docLoader = buildDocumentLoader(options, debug);
+async function buildSchemaDirectory(docLoader: DocumentLoader, debug: boolean): Promise<SchemaDirectory> {
     return new SchemaDirectory(docLoader, debug);
+}
+
+async function loadPatternJson(patternAccessor: string, docLoader: DocumentLoader, debug: boolean): Promise<object> {
```
LeighFinegold, line 154: Are these not still command helper methods? Technically the DocLoader here could still be FileSystemDocLoader

shared/src/commands/generate/generate.ts
```diff
@@ -17,15 +17,13 @@ export async function runGenerate(pattern: object, outputPath: string, debug: bo
...
         mkdirp.sync(dirname);
         fs.writeFileSync(outputPath, output);
         logger.info(`Successfully generated architecture to [${outputPath}]`);
     } catch (err) {
-        logger.debug('Error while generating architecture from pattern: ' + err.message);
-        if (debug) {
-            logger.debug(err.stack);
-        }
+        logger.error('Error while generating architecture from pattern: ' + err.message);
+        logger.debug(err.stack);
```
LeighFinegold, line 27: Altough not necessary your code, wondering if we should be have a general policy that command methods use Promise.reject and they should always bubble up to the cli

cli/src/cli.ts
```diff
@@ -109,20 +116,47 @@ export function setupCLI(program: Command) {
...
+    try {
+        const url = new URL(patternAccessor);
+        return await loadPatternFromCalmHub(url.href, docLoader, debug);
+    } catch (_) {
+        // If the pattern is not a URL, it must be a file path
+        return await loadJsonFromFile(patternAccessor, debug);
```
LeighFinegold, line 160: Have we thought this falling back to the FileSystemDocLoader?

LeighFinegold:
Reviewed PR. Some design thoughts/code consolidation if we extend beyond generate function:-

- I like the concept of the calm-hub config and how that can be used in conjunction with the end points to control configuration to calmhub. I could see this be important to contain other information around connectivity to calmhub. 
- The code will create a DocLoader from CalmHub or FileSystem depending on the url. If the document can't be in CalmHub then it falls back to loading via file not using the DocLoader. Have we thought about a CompositePattern similar to what was done on https://github.com/finos/architecture-as-code/blob/main/shared/src/resolver/calm-reference-resolver.ts#L72
 - On a related note, there is a little bit of overlap between the Loaders and the CalmReferenceResolver although the CalmReferenceResolver specifically was targeting a public html endpoint without any knowledge of calmhub. Isn't the CalmHubDocLoader just another specialised resolver? I guess the difference here would be if URLs from calmhub be in a diff form calmhub:
 - The FileSystemDocLoader stores the loaded files into the SchemaDirectory but the CalmHubDocLoader does not. 
 - Although in this pattern example you are probably only dealing with a single pattern document, the code won't work for other endpoints that need to do parse and load multiple documents as part of parsing the document.

willosborne:
Addressing comments above:
    
> The code will create a DocLoader from CalmHub or FileSystem depending on the url. If the document can't be in CalmHub then it falls back to loading via file not using the DocLoader. Have we thought about a CompositePattern similar to what was done on https://github.com/finos/architecture-as-code/blob/main/shared/src/resolver/calm-reference-resolver.ts#L72

So you mean that even when CalmHub is configured, we still fall back to the filesystem rather than either-or? That is a great idea - I will raise an issue for that. Then if you set both schemaDirectory and a calmHub url it will configure that.

> On a related note, there is a little bit of overlap between the Loaders and the CalmReferenceResolver although the CalmReferenceResolver specifically was targeting a public html endpoint without any knowledge of calmhub. Isn't the CalmHubDocLoader just another specialised resolver? I guess the difference here would be if URLs from calmhub be in a diff form calmhub:

They are definitely similar. I think the main difference is that the doc loader is CalmHub-aware - right now that doesn't make much difference, but there may be some consequences if we add a system for human-readable URLs in later requiring some kind of lookup for an 'alias'. As such I want to leave this in for now until we come to a conclusion on how we're doing that.
    
> The FileSystemDocLoader stores the loaded files into the SchemaDirectory but the CalmHubDocLoader does not.

Not explicitly; it returns the document directly to the SchemaDirectory which is then responsible for storing it. I couldn't do that for the loadSchemas() method because it's loading multiple schemas in one go, but that's necessary for the recursive file system setup. Maybe that could be changed to hit a callback in the same manner - but it is much, much easier to mock this way!
    
> Although in this pattern example you are probably only dealing with a single pattern document, the code won't work for other endpoints that need to do parse and load multiple documents as part of parsing the document.

Could you provide an example? Do you mean e.g. docify having to load a bunch of template files? This setup will load any references it sees against calmhub - although it only supports core schemas right now by design.

cli/src/cli.ts
willosborne: your point being that actually loading the file that we use should be handled by the command itself, rather than injecting a fully-loaded file for validation/generation etc?
My logic here is that I'm trying to keep logic related to loading things in cli, so that if we ever want to put this into the browser it's a bit isolated. I know we aren't there yet - filesystemdocloader is in shared - but this way of doing things felt more natural than wiring more logic in.
The other thing is that each command would need to implement this logic separately rather than just doing it once here.

What do you think?

cli/src/cli.ts
willosborne: I guess moving this code to the fs doc loader would make sense. I can't re-use the native methods there because this is just loading one file, not a whole directory. Let me take. a look at that

willosborne: OK so the problem here: docloader takes schema IDs, not file paths. we would have to use the loadMissingDocument function in a different way. I'm going to add another method that just loads any file by URL/file path and returns it

willosborne: Argh, even more problems. We can't just call the loaded document loader, because what if that's a calmhub url? We always need access to load from the file system.

I'm going to say let's not do this now, and i will add this logic in as part of the fall-back logic for the file system.
Otherwise extra logic is needed and i need to make filesystem loaders just to load the pattern.

shared/src/commands/generate/generate.ts
willosborne: Throwing in an async method is equivalent to Promise.reject, right? It's an interesting point - have been hesitant to throw errors and complicate teh CLI error handling, but maybe this would be a good change.
Agreed not for this PR though.

cli/src/cli.ts
```diff
@@ -23,17 +26,20 @@ export function setupCLI(program: Command) {
...
             const debug = !!options.verbose;
             const docLoaderOpts = await parseDocumentLoaderConfig(options);
-            const schemaDirectory = await buildSchemaDirectory(docLoaderOpts, debug);
-            const pattern: object = await loadJsonFromFile(options.pattern, options.verbose);
+            const docLoader = buildDocumentLoader(docLoaderOpts, debug);
+            const schemaDirectory = await buildSchemaDirectory(docLoader, debug);
+            const pattern: object = await loadPatternJson(options.pattern, docLoader, debug);
             const choices: CalmChoice[] = await promptUserForOptions(pattern, options.verbose);
             await runGenerate(pattern, options.output, debug, schemaDirectory, choices);
+            await runGenerate(pattern, options.output, debug, schemaDirectory, choices);
```
aidanm3341, line 42: am I missing something or is this line duplicated? What are we running generate twice?

willosborne: reeks of a bad merge from yours truly. will remove

aidanm3341 (approved):

PR #1254, opened 2025-05-05
Add inline control configuration support to CALM schema and example f…

rocketstack-matt:
…iles


jpgough-ms (approved):
I like the approach taken here.

PR #1258, opened 2025-05-05
Remove authentication property from relationships in favor of control…

rocketstack-matt:
… requirements


jpgough-ms (approved):
This looks good to me

PR #1293, opened 2025-05-16
Remove unused cytoscape dependencies

oliviajanejohns:
Removed these dependencies as they are unused and was seeing errors in our console previously
Screenshot 2025-05-16 at 16 19 09

### DIFF
--- a/calm-hub-ui/package.json
+++ b/calm-hub-ui/package.json
@@ -23,7 +23,6 @@
         "cytoscape": "^3.30.3",
         "cytoscape-cola": "^2.5.1",
         "cytoscape-dagre": "^2.5.0",
-        "cytoscape-expand-collapse": "^4.1.1",
         "cytoscape-node-edge-html-label": "^1.0.6",
         "oidc-client": "^1.11.5",
         "react": "^18.2.0",
--- a/calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
+++ b/calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
@@ -2,7 +2,6 @@ import './cytoscape.css';
 import { useContext, useEffect, useRef, useState } from 'react';
 import cytoscape, { Core, EdgeSingular, NodeSingular } from 'cytoscape';
 import nodeEdgeHtmlLabel from 'cytoscape-node-edge-html-label';
-import expandCollapse from 'cytoscape-expand-collapse';
 import { Sidebar } from '../sidebar/Sidebar.js';
 import { ZoomContext } from '../zoom-context.provider.js';
 import { Edge, CalmNode } from '../../contracts/contracts.js';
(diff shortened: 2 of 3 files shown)


calm-hub-ui/src/visualizer/components/sidebar/Sidebar.tsx
```diff
@@ -49,107 +48,244 @@ export function Sidebar({ selectedData, closeSidebar }: SidebarProps) {
...
+                    <span className="font-light">target: </span>
+                    <span className="font-semibold">{selectedData.target}</span>
+                </p>
+            </div>
+        </>
+    );
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function ControlDisplay({ control }: { control: any }) {
```
aidanm3341, line 217: given that the data here is just `any`, is this really a 'ControlDisplay'? Or is it just a generic JSON renderer? And if that's the case, can't we just use this for rendering everything rather than have specific components for each thing?

calm-hub-ui/src/visualizer/components/sidebar/Sidebar.tsx
```diff
@@ -49,107 +48,244 @@ export function Sidebar({ selectedData, closeSidebar }: SidebarProps) {
...
+                <p>
+                    <span className="font-light">target: </span>
+                    <span className="font-semibold">{selectedData.target}</span>
+                </p>
+            </div>
+        </>
+    );
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
aidanm3341, line 216: also can we change the type so that we don't need to disable this linting rule? If we genuinely accept any type here, can we use `unknown` instead?

calm-hub-ui/src/visualizer/components/sidebar/Sidebar.tsx
oliviajanejohns: Good point. Its generic JSON renderer so will utilise for interfaces too

calm-hub-ui/src/visualizer/components/sidebar/Sidebar.tsx
```diff
@@ -49,107 +48,244 @@ export function Sidebar({ selectedData, closeSidebar }: SidebarProps) {
...
+
+                {!isCALMEdge && !isCALMNode && (
+                    <div className="text-xl font-bold mb-2">Unknown Selected Entity</div>
+                )}
+            </div>
+        </div>
+    );
+}
+
+function NodeDisplay({
```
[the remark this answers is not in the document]
oliviajanejohns: done

calm-hub-ui/src/visualizer/components/sidebar/Sidebar.tsx
oliviajanejohns: done

aidanm3341:
yeah, this looks good to me. I personally like the simplicity of this approach @oliviajanejohns, and it's still human readable enough to be worth it IMO. I'm not certain from your screenshot, but are you just using this for controls specifically here? Can we just use this approach to render the entire node?

PR #1300, opened 2025-05-19
Adding support for interfaces 

andreeaniculcea:
Adding the ability to
- create an interface for a given namespace
- retrieve a list of interfaces for a given namespaces


jpgough-ms:
Converting to draft until CLA is sorted.

jpgough-ms:
@andreeaniculcea build is failing for this, looks like something could be missing from the PR?

PR #1301, opened 2025-05-19
Adds input constraints and XSS sanitization to Architecture and Pattern endpoints

grahampacker-ms:
Adds input constraints and XSS sanitization to Architecture and Pattern endpoints (#761)

There will be a follow up PR for flows.

### DIFF
--- a/calm-hub/pom.xml
+++ b/calm-hub/pom.xml
@@ -62,6 +62,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-validator</artifactId>
+        </dependency>
 
         <!-- Testing -->
         <dependency>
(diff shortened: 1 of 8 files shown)


calm-hub/pom.xml
```diff
@@ -185,6 +189,7 @@
         <configuration>
         <excludes>
           <exclude>**/*Builder.*</exclude>
+          <exclude>**/*Constants.*</exclude>
```
grahampacker-ms, line 192: Constants used in annotations are not detected as covered by Jacoco

calm-hub/pom.xml
```diff
@@ -61,6 +61,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-security</artifactId>
     </dependency>
+    <dependency>
```
jpgough-ms, line 64: I don't understand why this dependency is needed

jpgough-ms (changes requested):

calm-hub/pom.xml
grahampacker-ms: This is needed to enable constraint validation per: https://quarkus.io/guides/validation

I also did some reading and this is the reference implementation for bean validation: https://hibernate.org/validator/
`Hibernate Validator is the reference implementation for the [Bean Validation](https://beanvalidation.org/) specification. Indeed, the original specification was inspired by an early version of Hibernate Validator.`

Issue #1239, opened 2025-05-01
Hosted CALM-HUB


rocketstack-matt:
### Target Project:
calm-hub and calm-hub-ui - The server component and web interface of CALM that host and visualize architectural diagrams and schema definitions

### Description of Feature:
Deploy and host an official instance of calm-hub and its visual interface calm-hub-ui at calm-hub.finos.org under FINOS management. This centrally hosted instance would serve as the reference implementation of the CALM ecosystem, providing a persistent location for core schemas, shared controls, interfaces, and architectural patterns, with an intuitive visual interface for creating and exploring architectures. It would allow users to test and experience the full capabilities of CALM without having to deploy their own infrastructure.

### User Stories:
- As a potential CALM user, I want to access a live demo environment so that I can explore CALM's capabilities before investing in my own deployment.
- As an architect, I want to reference canonical schemas and controls from a trusted source so that I can ensure my architectures adhere to community standards.
- As a developer, I want to test my CALM CLI integrations against a stable API endpoint so that I can validate my implementations.
- As a community contributor, I want to share architectural patterns and controls in a central repository so that others can benefit from and build upon my work.
- As an enterprise adopter, I want to access a reference implementation so that I can understand how to deploy CALM in my own environment.
- As a solution designer, I want to use the visual interface to create and modify architectures so that I can collaborate with stakeholders without requiring them to understand JSON.

### Current Limitations:
1. No official hosted version of calm-hub exists, requiring users to deploy their own instances for testing
2. Lack of a centralized repository for canonical schemas, controls, and interfaces
3. Difficulty in demonstrating CALM capabilities to potential adopters without significant setup
4. No persistent environment for community members to share and collaborate on architectural patterns
5. Limited ability to experience the full CALM ecosystem without personal infrastructure investment
6. Testing new CALM features requires local deployments, complicating the feedback process

#### Infrastructure Requirements:
- Dedicated domain (calm-hub.finos.org) with proper DNS configuration
- Cloud-hosted infrastructure appropriate for running the calm-hub application
- Continuous deployment pipeline from the main repository
- Backup and recovery procedures
- Monitoring and observability tools

#### Functional Components:
1. **Public Access Layer**
   - Anonymous read access to public schemas, controls, and patterns
   - Authentication system for contributors and registered users
   - Role-based access control for different permission levels

2. **Schema Repository**
   - Hosting of all official CALM schema versions
   - Versioning and release management
   - Schema validation services

3. **Control and Interface Library**
[...]

PR #1307, opened 2025-05-20
Add calm-hub standalone mode

rocketstack-matt:
* Add NitriteDB support for an in process mongo like DB
* Enable file backed persistence for running in standalone mode with persistence

Addresses [#1239](https://github.com/finos/architecture-as-code/issues/1239) making it easier for developers and companies to run their own instance of calm-hub without Docker or MongoDB

To test, once you have packaged the calm-hub jar you can simply run:

`java -Dcalm.database.mode=standalone -jar target/quarkus-app/quarkus-run.jar`

Adding something to the DB via the swagger UI will allow you to view the docs in the calm-hub UI and restarting the process should show that the added entities persist over restarts.


calm-hub/src/main/java/org/finos/calm/config/NitriteDBConfig.java
jpgough-ms, line 1: My main concern with this file is that if we change `init-mongo.js` this complex parsing logic is going to fail, do we need it to be like this? Should we reconsider using init-mongo?

calm-hub/src/main/java/org/finos/calm/store/producer/DomainStoreProducer.java
jpgough-ms, line 1: Thinking about these files, if they are generated/created, the prompt to create them is something we should document - otherwise all incoming changes are going to have a lot of work to get this to work.

calm-hub/src/test/java/org/finos/calm/config/NitriteDBConfigTest.java
jpgough-ms, line 1: Similar concerns with this file and the `NitriteDBConfig` that this is going to be brittle.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteCoreSchemaStoreShould.java
```diff
@@ -0,0 +1,130 @@
...
+        // Arrange
+        DocumentCursor cursor = mock(DocumentCursor.class);
+        when(cursor.iterator()).thenReturn(new EmptyDocumentIterator());
+        when(mockCollection.find()).thenReturn(cursor);
+
+        // Act
+        List<String> versions = schemaStore.getVersions();
+
+        // Assert
+        assertTrue(versions.isEmpty());
```
jpgough-ms, line 51: This doesn't follow the other tests where we use Hamcrest matchers to do

`assertThat`

jpgough-ms (commented):
A couple of comments. Overall I like that we would have an in memory solution, however I am concerned that this will make changing/adding new features more difficult and some of the tests/config introduce a fair bit of brittleness. 

I am wondering whether writing the SQL version of the store + then using H2 would be a better approach. It would also offer the ability then for teams/orgs to use Postgres or similar if they cannot use Mongo, whilst offering us the in memory solution.

calm-hub/src/main/java/org/finos/calm/config/NitriteDBConfig.java
rocketstack-matt: Agreed, I was trying to make sure we had parity with the mongo solution, but I could make this far simpler if we're happy to simply have specific data collections to load and remove the brittle json parsing?

Could also remove bootstrapping altogether, perhaps something to discuss regarding how we see folks using this as they first start to get hands on with calm.

rocketstack-matt:
> A couple of comments. Overall I like that we would have an in memory solution, however I am concerned that this will make changing/adding new features more difficult and some of the tests/config introduce a fair bit of brittleness.
> 
> I am wondering whether writing the SQL version of the store + then using H2 would be a better approach. It would also offer the ability then for teams/orgs to use Postgres or similar if they cannot use Mongo, whilst offering us the in memory solution.

Not sure why this would be less brittle - obviously we can add as many implementations for backing store as we like and this change introduces the necessary changes to make different stores pluggable, which would be needed no matter what options we introduce.

I did originally consider H2 and ended up going with Nitrite to deliberately stick to the most similar storage option to Mongo for the first alternative, and it had the convenience of also offering a file backed option for simple startup.

To me, the brittle code here is all around bootstrapping; if we're happy to say we either don't care about bootstrapping this version, or provide a more native way to bootstrap then I think this is no better or worse than any other tech choice.

One thing that we could do to make it store agnostic is to actually provide a bootstrap script which could be used post-startup to bootstrap directly via the REST endpoints?

calm-hub/src/main/java/org/finos/calm/store/producer/DomainStoreProducer.java
rocketstack-matt: - [ ] There were manually created - will add details to the README for devs adding future store implementations as to how to wire into the Producers.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteCoreSchemaStoreShould.java
rocketstack-matt: [ ] Switch Nitrite tests to use ham crest matchers

jpgough-ms (approved):

Issue #1234, opened 2025-05-01
Toggling on/off descriptions in calm-hub should not reset the layout


rocketstack-matt:
### Steps to Reproduce:
1. Open an architecture diagram in the calm-hub visualization tool
2. Manually position/move nodes to create a custom layout
3. Toggle on or off the node descriptions or relationship descriptions using the visibility controls

### Expected Result:
The visualization should maintain the user's custom node positions when toggling the visibility of node or relationship descriptions.

### Actual Result:
The entire layout is reset to its default state, discarding all custom node positioning that the user had set up. The visualization returns to its initial algorithmic layout, causing the user to lose their work arranging the diagram.

### Environment:
- CALM-Hub UI
- Occurs consistently across different operating systems

### Additional Context:
This issue creates a frustrating user experience when trying to refine architecture diagrams. Users often spend time positioning nodes to create meaningful layouts that show logical groupings or flow. The layout reset forces users to redo their positioning work whenever they need to toggle description visibility.

Recommendation: Preserve node positions in a state variable separate from the description visibility settings, ensuring that toggling descriptions doesn't trigger a full diagram re-layout.

PR #1316, opened 2025-05-22
Added the functionality to keep node and edge positions when moved 

Adwoa-Konadu-Appiah:
In relation to Issue #1234, I have added the functionality to keep node and edge positions when moved around when the description toggle is toggled on or off.

The current functionality stores the new position of elements in the browser when the user moves nodes around in the UI.

The saved positions are rendered after the user toggles the node or edge description button on or off.

The images below describe the functionality:
**_The default element positions when an architecture is rendered._**
element positions on intialization

**_Trader and Web GUI node positions change, and the Node description button is toggled off_**
no node description

**_The node description  and Relationship  description buttons are toggled off_**
no node and edge description


calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.test.tsx
aamanrebello, line 1: Thoughts on putting this under the services folder?

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.tsx
aamanrebello, line 1: Thoughts on putting this in the services folder? I feel like this could qualify as a service

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -121,6 +134,15 @@ export function CytoscapeRenderer({
             edgeClickedCallback(edge?.data());
         });
 
+
+        cy.on('dragfree', 'node', () => {
```
aamanrebello, line 138: This is awesome, you only update the position after we stop dragging, love the efficiency!

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.tsx
Adwoa-Konadu-Appiah: That's right I will make the changes thanks

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.tsx
```diff
@@ -0,0 +1,31 @@
...
+    id: string;
+    position: { x: number; y: number };
+  }
+  
+  const STORAGE_KEY = 'cytoscapeNodePositions';
+  
+  export const CytoscapeElementPositionStorage = {
+    save(positions: StoredNodePosition[]) {
+      try {
+        localStorage.setItem(STORAGE_KEY, JSON.stringify(positions));
```
aidanm3341, line 11: interesting!! Using local storage is cool because it means the positions will be saved even the next time you visit the page. 

One thing I am curious about, does this work as expected when you have different architectures which have similar nodes? If I had `architecture A` which contained `node-1` and `architecture B` which also has a node called `node-1`, I think the same position would be used in both which could lead to some very confusing results!

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.tsx
aidanm3341, line 1: the formatting is a little off in this file. Do you have the prettier extension setup in VSCode? I would recommend

calm-hub-ui/src/visualizer/components/cytoscape-renderer/cytoscapeElementPositionStorage.tsx
Adwoa-Konadu-Appiah: If the nodes have the same ID, then yes, that node will take the position stored in the local storage on load. At the moment we use breadthfirst to set the initial position on load.

rocketstack-matt: Could the storage key be changed to include the $id of the document as a prefix, thereby only clashing if people uploaded architectures with clashing IDs?

calm-hub-ui/src/visualizer/services/cytoscapeElementPositionStorageService.tsx
```diff
@@ -0,0 +1,34 @@
+export interface StoredNodePosition {
+    id: string;
+    position: { x: number; y: number };
+}
+
+const STORAGE_KEY = '_CytoscapeNodePositions';
+
+export const CytoscapeElementPositionStorage = {
```
aidanm3341, line 8: Functionally I'm happy with this, but one thing from a code style perspective is that I'm not sure about this const value which has the functions on it. It's sort of being used like a static class and is named like one by using PascalCase, but is actually just a const value. 

IMO a more idiomatic way of doing this is just to export two top level functions instead, so it could look something like this:

```typescript
export function saveNodePositions(title: string, positions: StoredNodePosition[]) { ... }
export function loadNodePositions(title: string) { ... }
function createKey() { ... } // this doesn't need to be exposed, right?
```

what do you think?

Adwoa-Konadu-Appiah: That's right. I will make the changes.

calm-hub-ui/src/visualizer/services/cytoscapeElementPositionStorageService.tsx
```diff
@@ -0,0 +1,34 @@
+export interface StoredNodePosition {
+    id: string;
+    position: { x: number; y: number };
+}
+
+const STORAGE_KEY = '_CytoscapeNodePositions';
```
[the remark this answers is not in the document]
Adwoa-Konadu-Appiah: I'll rename the  position service files as well.

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -111,6 +114,18 @@ export function CytoscapeRenderer({
             maxZoom: 5,
         });
 
+        const savedPositions = CytoscapeElementPositionStorage.load(title);
```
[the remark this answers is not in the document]
aidanm3341: I think using the title in general makes sense under the assumption that the title is unique. However with the way that it's currently set up, I don't think it actually IS unique. 

The assumption in this code is reasonable though, and one of the refactors that I want to do is to improve this, so I don't think it needs to be part of this PR.

Issue #1190, opened 2025-04-17
[Bug][Minor] Upload menu button does not collapse once file is upload in visualizer


pranitk:
## Bug Report
Upload menu button does not collapse once file is upload in visualizer

![Image](https://github.com/user-attachments/assets/e6b0878f-a09f-4b88-bbde-c784162ddf5e)

![Image](https://github.com/user-attachments/assets/e6caa878-f228-4e2c-940e-e284fc2561f5)

### Steps to Reproduce:
 1. Go to visualizer
 2. Click upload
 3. Click architecture
 4. Select a file

### Expected Result:
Upload button should collapse the menu

### Actual Result:
Upload menu stays open, adding extra click for user to collapse it so that it doesn't come in the middle of the right panel.

### Environment:
main branch code in local

### Additional Context:
Suggestion: If no other options to be introduced inside 'Upload', just replace it with a button directly.

![Image](https://github.com/user-attachments/assets/80c53d52-965e-40f0-bda9-2d9561a48530)

PR #1317, opened 2025-05-23
Issue 1190: Update upload button to avoid expand/collapse issue

pranitk:
In relation to https://github.com/finos/architecture-as-code/issues/1190
Before: After uploading file, user has to click outside to close the dropdown. It also interfered with the node/edge details section on the right.

Above change avoids that since we only have one option and hence uses a button directly.


pranitk:
image
Avoids extra click to hide this dropdown.

oliviajanejohns (approved):

PR #1330, opened 2025-05-30
WIP:  Use Architecture Id as Position Storage key

Adwoa-Konadu-Appiah:
With #1234, positions are currently being stored with Architecture titles, which are not unique. 
This implementation allows us to store architecture with a key: `"${namespace}/{$calmtype}/${id}/${version}`
Screenshot 2025-05-30 at 14 17 43

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-	"image": "mcr.microsoft.com/devcontainers/universal:2",
+	"image": "mcr.microsoft.com/devcontainers/universal:3",
 	"features": {
 		"ghcr.io/devcontainers-contrib/features/maven-sdkman:2": {
 			"version": "latest",
@@ -11,5 +11,12 @@
 			"jdkVersion": "latest",
 			"jdkDistro": "open"
 		}
+	},
+	"containerEnv": {
+		"JAVA_HOME": "/root/.sdkman/candidates/java/current"
+	},
+	"hostRequirements": {
+		"cpus": 4,
+		"memory": "16gb"
 	}
 }
--- a/calm-hub/pom.xml
+++ b/calm-hub/pom.xml
@@ -5,8 +5,8 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.finos.calm</groupId>
-        <artifactId>architecture-as-code</artifactId>
+        <groupId>org.finos.architecture-as-code</groupId>
+        <artifactId>parent</artifactId>
         <version>1.0.0-SNAPSHOT</version>
     </parent>
 
--- a/calm/pom.xml
+++ b/calm/pom.xml
@@ -4,8 +4,8 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.finos.calm</groupId>
-        <artifactId>architecture-as-code</artifactId>
+        <groupId>org.finos.architecture-as-code</groupId>
+        <artifactId>parent</artifactId>
         <version>1.0.0-SNAPSHOT</version>
     </parent>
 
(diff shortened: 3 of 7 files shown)


calm-hub-ui/src/model/calm.tsx
```diff
@@ -6,7 +6,15 @@ export type ArchitectureID = string;
 export type FlowID = string;
 export type Flow = string;
 export type Version = string;
+export enum CalmType {
```
aidanm3341: enums are generally bad practice in TS - can we instead use a union of strings? i.e. `type CalmType = 'architecture' | 'pattern' | 'flow'` or something similar.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -179,6 +187,13 @@ export function Drawer({ calmInstance, title, isConDescActive, isNodeDescActive
             .filter((edge): edge is Edge => edge !== undefined);
     }
 
+    function setStorageKey(title: string, data?: Data): string {
```
aidanm3341: I wouldn't use the word 'set' in this function name, it implies that we're changing some state. maybe `createStorageKey` or something similar?

calm-hub-ui/src/hub/components/json-renderer/JsonRenderer.tsx
```diff
@@ -1,16 +1,24 @@
 import { useNavigate } from 'react-router-dom';
 import { allExpanded, defaultStyles, JsonView } from 'react-json-view-lite';
 import 'react-json-view-lite/dist/index.css';
+import { Data } from '../../../model/calm.js';
 
 interface JsonRendererProps {
-    json?: object;
+    json?: Data;
```
aidanm3341: I don't think we should be using 'Data' here - this component should be as pure as possible in the sense that it has no reason to know what CALM is. It's just a pure json renderer, it should be able to accept any json object as input and be able to render it.

calm-hub-ui/src/hub/components/json-renderer/JsonRenderer.tsx
```diff
@@ -19,7 +27,11 @@ function JsonDisplay({ data, handleClick }: { data: object; handleClick: () => v
             >
                 Visualize
             </button>
-            <JsonView data={data} shouldExpandNode={allExpanded} style={defaultStyles} />
+            <JsonView
+                data={DisplayData(data)}
```
aidanm3341: I'm not sure I understand the purpose of this change. Does this change the structure of what's displayed in the side panels?

calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -117,7 +119,7 @@ export function CytoscapeRenderer({
             maxZoom: 5,
         });
 
-        const savedPositions = loadStoredNodePositions(title);
+        const savedPositions = loadStoredNodePositions(calmKey ? calmKey : title);
```
aidanm3341: I was having a look at how the 'title' is used in the component, and from what I can see it's _only_ used for the purposes of this key. 

In that case, it would be better to remove the title from the props of this component entirely, and instead just have the 'calmKey' which becomes a mandatory property. Then in the component which calls this component, you can choose whether to pass in the title, or the new calm-hub ID if it exists.

That way, you don't need these ternary statements and optionals. You can just use 'calmKey' on it's own everywhere

calm-hub-ui/src/hub/components/json-renderer/JsonRenderer.tsx
Adwoa-Konadu-Appiah: The current data would display the calm type ID and version.

PR #1336, opened 2025-05-30
Enable selection of node or edge on single click

aamanrebello:
When child node is selected:
image
When edge is selected:
image
When parent node is selected:
image


calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -175,8 +158,6 @@ export function CytoscapeRenderer({
         edges,
         isNodeDescActive,
         isRelationshipDescActive,
-        nodeClickedCallback,
```
aamanrebello, line 178: I removed these dependencies from the `useEffect` and now I can select nodes without having to click twice, I only need to click once. I don't know if removing these dependencies has any side effects, let me know.

aidanm3341: do you have the eslint extension installed? It will raise issues like this in the editor. You can also just run 'npm run lint'. I think will be an issue because these are external to the hook, but the hook makes use of them.

aamanrebello: So I am only getting the one warning, I actually added in an eslint disable comment before the hook dependencies, otherwise I was getting another warning. But for now I just have this:

```
C:\Users\user\Documents\GitHub\architecture-as-code\calm-hub-ui\src\index.tsx
  10:7  warning  Fast refresh only works when a file has exports. Move your component(s) to a separate file  react-refresh/only-export-components

✖ 1 problem (0 errors, 1 warning)
```

aidanm3341: I don't think we should disable that warning, it's valid in this case and disabling it might hide bugs. Maybe we need to take a step back and look at how we should be dealing with these callbacks in the first place, it could be that we've not designed it very well and that's why we're getting these issues.

That second warning is something to do with the auth work that Mani has been looking at. It's not currently in use so it isn't a big deal. It will be worth looking at eventually though, even just to stop getting the warnings on every PR 😆

aamanrebello: I think I solved it. I have added the callbacks in so linting is gone, and now I am memoizing the callbacks using `useCallback`. I'm not sure how that has solved the problem, but it seems to have done. The nodes and edges now highlight on a single click rather than needing 2 clicks. This seems like a great change to merge in before OSFF, the highlighting will look good (take a look though, and hope the build works).

aidanm3341 (dismissed):

calm-hub/nitrite/init-nitrite.sh
aidanm3341, line 1: what is this change? It says "100755 → 100644" - I'm not sure what that means? Either way the fact that this has been included means we need an approval from a calm hub maintainer. Can you see if you can revert this change?

aamanrebello: Reverted.

PR #1339, opened 2025-06-02
Standards

jpgough-ms:


jpgough-ms:
This is the first PR for #1339.

I think there might be a residual bug when it comes to updating the name and description of the Standard. However, I think we'll get this merged and I'll look at it as a follow up.

@grahampacker-ms is also going to kindly look at the validation work

PR #1342, opened 2025-06-05
Potential fix to stop nodes moving apart on toggling descriptions (Not Working Yet)

aamanrebello:
IGNORE - this does not seem to work


calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.tsx
```diff
@@ -125,8 +125,6 @@ export function CytoscapeRenderer({
                     node.position(match.position);
                 }
             });
-
-            cy.fit();
```
aamanrebello, line 129: So a) I'm hoping this does not affect anything UI/UX wise, this "Pan and zooms the graph to fit to a collection". Also b) Please check if it works, I think it does so far from what I've been trying.

PR #1347, opened 2025-06-06
Update standards to also persist the name and description into the to…

jpgough-ms:
…p level doc on creating a new version


jpgough-ms:
Fixes #1339

rocketstack-matt (approved):
As discussed with @jpgough-ms the standards endpoint is a different approach to the others, returning more detail from the base get rather than just the ids. This was agreed on a prior office hours to make it easier to provide richer ui and will be ported to the other endpoints. 

We also have some superfluous details on the request body that we should remove in a subsequent pr (id, version, namespace).

PR #1355, opened 2025-06-13
#1234 added persisting of pan and zoom between renders and fixed repositioning bug on rerender

aidanm3341:
- Added the ability for the pan and zoom to be maintained between rerenders (i.e. when you clicked on a node, or when you used the toggles in the tool bar etc.)
- Fixed the bug where the container-like relationships were being incorrectly rerendered on every rerender
- Added some basic unit tests for the CytoscopeRenderer component


calm-hub-ui/src/visualizer/components/cytoscape-renderer/CytoscapeRenderer.test.tsx
```diff
@@ -0,0 +1,165 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
oliviajanejohns, line 1: Nice! Glad to see tests added here

oliviajanejohns (approved):

Issue #1352, opened 2025-06-12
Docify Website Updates for OSFFLN-2025


LeighFinegold:
This issue tracks a set of enhancements to the Docify website experience for OSSF

## ✅ Goals

The following improvements are proposed to enhance usability, completeness, and visual polish of the generated site:

### 1. Support Relationships in the Side Menu
- Update the navigation sidebar to include `relationships` with a dedicated page
- Ensure consistent with nodes page

### 2. Fix Controls List on Home Page
- Currently, the controls list on the homepage omits `relationships` - see https://github.com/finos/architecture-as-code/issues/1126

### 3. Render Metadata Sections
- Display any `metadata` defined in the CALM document
- Gracefully degrade to `No Metadata defined.` if empty

### 4. Add Logo to Navbar
- Add OSFF or Docify branding logo to the navbar

PR #1356, opened 2025-06-15
Adding ADRs, Cleaning Up Flow Sequence Generation + Support for inline controls on docify (#1352)

LeighFinegold:
Additional enhancements for calm docify prior to OSFF-LN

- Fix for composed-of relationship rendering
- Sequence Diagram leverage architecture to render Node Names

Screenshot 2025-06-15 at 20 37 20
Screenshot 2025-06-15 at 20 37 25


shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+import { Architecture } from '../../model/core';
+import { CalmControl } from '../../model/control';
+
+/**
+ * A class that manages control requirements and configurations across architecture documents
+ */
+export class ControlRegistry {
+    private controlRequirements: Record<string, { id: string, content: string, domain: string, scope: string, appliedTo: string }[]> = {};
```
willosborne: This inline type seems to be duplicated each time - should this be pulled out to a separate type definition?

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+     */
+    private addControlRequirementToTable(controls: CalmControl[], scope: string, appliedTo: string): void {
+        controls.forEach(control => {
+            control.requirements.forEach(detail => {
+                const requirement = detail.controlRequirementUrl;
+                const id = this.getRequirementId(requirement);
+
+                if (!id) return;
+
+                const existing = this.controlRequirements[id] ?? [];
```
willosborne, line 80: is a control requirement ID not unique? this some() logic seems to imply you can have multiple control requirements with the same id, but as long as the scope/domain are different things are ok? apologies if misunderstanding here.

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+                    });
+                    this.controlRequirements[id] = existing;
+                }
+            });
+        });
+    }
+
+    private getRequirementId(requirement: unknown): string | undefined {
+        if (typeof requirement === 'string') {
+            return requirement.split('/').filter(Boolean).pop();
```
willosborne: What does this line do? Seems like it might be a nifty JS trick I haven't seen before, I don't understand 😆

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        if (
+            typeof requirement === 'object' &&
+            requirement !== null &&
```
willosborne, line 118: would this catch a missing requirement/undefined? null is a specific value that undefined won't catch, i think.
should we maybe just check that it's a truthy value instead?

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+            '$id' in requirement &&
+            typeof (requirement as Record<string, unknown>)['$id'] === 'string'
+        ) {
+            return (requirement as Record<string, string>)['$id'].split('/').filter(Boolean).pop();
```
willosborne: Same as before, I don't understand this line. Could be worth pulling out to a helper function with name/docstring?

shared/src/docify/graphing/control-registry.ts
willosborne: i'm also thinking maybe that typeof object implies it's not null? this is just JS, it's not typed

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+                this.groupedByDomainConfigurations[config.domain].push(config);
+            });
+        });
+    }
+
+    /**
+     * Gets the flattened list of controls
+     * @returns Array of control objects
+     */
+    public getControls(): { id: string, name: string, schema: string, description: string, domain: string, scope: string, appliedTo: string, content: string }[] {
```
willosborne: same comment on duplication of type

shared/src/docify/graphing/control-registry.ts
```diff
@@ -0,0 +1,218 @@
...
+                ...config
+            }))
+        );
+    }
+
+    /**
+     * Gets the flattened list of control requirements
+     * @returns Array of control requirement objects
+     */
+    public getControlRequirements(): { id: string, content: string, domain: string, scope: string, appliedTo: string }[] {
```
willosborne: type

shared/src/docify/graphing/flow-sequence-helper.ts
```diff
@@ -0,0 +1,74 @@
...
+            relationshipId: transition.relationshipUniqueId,
+            source: this.getSourceFromRelationship(transition.relationshipUniqueId, architecture),
+            target: this.getTargetFromRelationship(transition.relationshipUniqueId, architecture)
+        }));
+    }
+
+    public getSourceFromRelationship(relationshipId: string, architecture: Architecture): string {
+        const relationship = this.findRelationshipById(relationshipId, architecture);
+        if (!relationship) {
+            const nodeId = relationshipId.split('-uses-')[0];
```
willosborne: is this convention documented/enforced anywhere?

shared/src/docify/graphing/control-registry.ts
LeighFinegold: The same control requirement specification can be used multiple times across an architecture.

shared/src/docify/graphing/control-registry.ts
LeighFinegold: its taking the last part of the URL to give the requirement an id - I think an issue/discussion I wanted to bring us is currently the model does assume the requirement documents is de-referenced. We just have the id from the the linked url.

shared/src/docify/graphing/flow-sequence-helper.ts
LeighFinegold: I was tempted to take it out. It was what was used in the code and the workshop.

shared/src/docify/graphing/control-registry.ts
LeighFinegold: I think i want to have a discussion potentially changing the whole model. That is currently the referencing process replacing the url with the actual document, but we now have the config section for inline controls. I don't know if the above is a good idea or not.

shared/src/docify/graphing/control-registry.ts
LeighFinegold: Moved it out of DocusaurusTransformer to abstract the logic that existed in it currently. 

Can we define this as a type? Not a terrible idea, but wondering if the underlying model of control is right atm.

shared/src/docify/graphing/control-registry.ts
LeighFinegold: Basically its checking its still the requirementUrl in dereferneced form or now an object having been dereferenced. The purpose of the class is to group requirements across the architecture so this visual is formed.

Screenshot 2025-06-16 at 11 43 07

shared/src/docify/graphing/control-registry.ts
willosborne: even if we just put it as a type ahead of more changes here, it would be good to reduce the duplication. right now it's hard to read and very hard to refactor

shared/src/docify/graphing/control-registry.ts
willosborne: Right, I see. if this is due to be refactored that makes sense - but in meantime could we please pull this out to a function to make this a bit easier to read. might not be refactored immediately

shared/src/docify/graphing/flow-sequence-helper.ts
willosborne: i'm inclined to think that a convention like this should potentially just be a new relationship type instead?

shared/src/docify/graphing/control-registry.ts
LeighFinegold: I additionally pulled out the id extraction logic but the method getRequirementId I feel encapsulates the general responsibility from obtaining from document vs url

shared/src/docify/graphing/flow-sequence-helper.ts
LeighFinegold: yes probably. When I raised this PR I tried to not seem like I was doing a full rewrite, but with this commentary I will just go ahead and change :D

willosborne (approved):

PR #1358, opened 2025-06-15
Fix tests in 'shared' workspace to work on Windows (finos#1357)

markscott-ms:
Before:
```
 Test Files  3 failed | 31 passed (34)
      Tests  6 failed | 186 passed (192)
   Start at  22:39:16
   Duration  64.13s (transform 3.96s, setup 0ms, collect 23.15s, tests 9.55s, environment 22ms, prepare 11.47s)
```

After fixing:
```
 Test Files  34 passed (34)
      Tests  192 passed (192)
   Start at  22:36:47
   Duration  58.53s (transform 2.34s, setup 0ms, collect 20.86s, tests 6.61s, environment 19ms, prepare 10.42s)
```


shared/src/template/template-calm-file-dereferencer.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { TemplateCalmFileDereferencer } from './template-calm-file-dereferencer';
 import path from 'path';
 import {InMemoryResolver} from '../resolver/calm-reference-resolver';
+import {resolve as url_resolve, cwd as url_cwd} from '@apidevtools/json-schema-ref-parser/lib/util/url';
```
willosborne, line 4: out of interest, would the built-in url.resolve() work here too? or is this needed to understand json ref style urls?

willosborne (approved):

shared/src/template/template-calm-file-dereferencer.spec.ts
markscott-ms: The built in Node method is discouraged in place of the WHATWG URL API: https://nodejs.org/api/url.html#urlresolvefrom-to

However, I have used the url resolution method from @apidevtools as the resolution needs to exactly mirror what happens within the CALM template engine/processor - which converts paths such as `C:\User\Bob\calm\...` to `c:/User/Bob/calm/...` (note that the case of the drive letter has changed!)

PR #1361, opened 2025-06-16
Add CALM logo to docs - minor version upgrade

rocketstack-matt:


oliviajanejohns (dismissed):

docs/package.json
```diff
@@ -16,8 +16,8 @@
     "dependency-check": "dependency-check --project 'calm-docs' --scan . --out ./dependency-check-report --format ALL --suppression ../.github/node-cve-ignore-list.xml"
```
oliviajanejohns, line 16: Just wondering - can we copy across the logos into the static/img folder on start? Similarly to as we do on the UI:
        "copy-public": "copyfiles -u 1 ../brand/**/* public/". 
        
This would help avoid duplicating these images.

rocketstack-matt: They have to be in the static folder for the build command to work, otherwise Docusarus will bomb out to avoid building an incorrect site with missing assets . . . is it possible to change the build to accommodate that?

rocketstack-matt: As discussed, we will merge this in for now (so we can get the logo in before OSFF) and look at whether it is feasible with Docusarus to copy these in to avoid duplicates after.

Issue #1354, opened 2025-06-13
CalmHub UI Hub Side Panel Bug


aidanm3341:
## Bug Report

When viewing the main 'Hub' page in the CalmHub UI, if the viewport is small enough, then a lot of the side panel is cut off making it unaccessible. @jpgough-ms  encountered this issue while doing a demo at a conference and meant that the UI could not be properly demo'd.

In this screenshot, you can see that the 'Architectures' column is partially cut off, and the 'Versions' column isn't even visible.
<img width="1007" alt="Image" src="https://github.com/user-attachments/assets/393d8f48-932e-458a-a3cf-189110d175e1" />

As a fairly easy tactical fix to this, I'd suggest that we only display one column at a time, and replace the currently shown column with the new one every time you click on a row. Then to navigate backwards, we have a simple breadcrumb component at the top which allows you to go back to a previous column view. 

DaisyUI (the component library that we use) has support for making breadcrumbs easily (https://daisyui.com/components/breadcrumbs/), and hopefully a lot of the existing code can be reused to achieve the result we want.

PR #1363, opened 2025-06-17
[Issue #1354] - Implementing new sidebar navigation

aidanm3341:
Niamh has been refactoring how the explore panel of the main Hub view works so that it will scale better on smaller screens. There's only a little work left on it so I'm just raising it as a PR now which I can properly review and push changes to if needed.

You can see here that navigation is now done via breadcrumbs, rather than stacking the columns next to each other.
image


calm-hub-ui/src/hub/Hub.test.tsx
```diff
@@ -0,0 +1,235 @@
+import React from 'react';
```
oliviajanejohns, line 1: Nice addition of tests here!

oliviajanejohns (approved):

Issue #1369, opened 2025-06-18
Workshop control configurations in generated docify website not rendering now that workshop docs are published


LeighFinegold:
**Description:**  
In our workshop example, we’ve relied on the `--url-to-local-file-mapping` argument to CALM Template to resolve control documents locally when they weren’t available online ([issue #1026](https://github.com/finos/architecture-as-code/issues/1026)). The workshop scripts use this flag (see [walkthrough.sh#L183](https://github.com/finos/architecture-as-code/blob/main/calm/workshop/walkthrough.sh#L183)) to pull in control JSON files.  

With the official control examples merged in [PR #1365](https://github.com/finos/architecture-as-code/pull/1365), those workshop control files now are network addressable and reachable — but when removing the  with `--url-to-local-file-mapping`, **control-config URLs** (e.g., `micro-segmentation.config.json`) are no longer dereferenced and remain literal strings.

## Steps to Reproduce
1. Check out `architecture-as-code` on `main` post-PR #1365.
2. Run the calm docify from the workshop
`
calm docify --input architecture/conference-secure-signup-amended.arch.json --output website
`

3. View docify website and note the controls sections indicates no controls.

## Proposed Fix
1. Ensure all network-addressable URLs (flows, control-requirements, control-configs) go through the same replacer pipeline:
   - **Mapped** → inline file content  
   - **Reachable** → `{ "$ref": "<url>" }`  
   - **Unreachable** → leave as string + log notice 
2. Add an E2E test verifying `control-config-url` dereferencing.

## References
- Offline mapping: issue [#1026](https://github.com/finos/architecture-as-code/issues/1026)  
- Workshop script: [walkthrough.sh#L183](https://github.com/finos/architecture-as-code/blob/main/calm/workshop/walkthrough.sh#L183)  
- Control examples merged: [PR #1365](https://github.com/finos/architecture-as-code/pull/1365)  

## Alternative/Additional  fixes
1. Never dereference but provide handlebar functions for the user to choose. For provided template-bundles where we are in control of Transformers, we will make decision to deference as of when we want to. 
2. In addition to the above, fail calm template fast if there any json documents which are non reachable and yet not mapped to a local file.

LeighFinegold:
[...]

Issue #490, opened 2024-10-17
Programmatic Derivation of CALM File Type


Thels:
### Description of Problem:
It's programmatically a bit of a pain to determine what sort of file you're validating/visualizing. Primarily as there is no indicator of what type of content you're processing without referring to schemas.

For example for the initial PR of the VSCode extension, this led to logic like `if $schema contains /pattern/` then it's an instantization. This isn't ideal and can quickly fall apart if someone refers to a pattern without this handy indicator.

Programmatically determining the type of file being processed could simplify our validation/visualization flows from a CLI perspective too.

### Potential Solutions:

- Introduce into the CALM schema the concept of a type - so the JSON files self-describe what they are.

`"calm-type": ["pattern", "pattern-instantiation", ...others ]`

- Taken from https://github.com/finos/architecture-as-code/discussions/482#discussioncomment-10972300 - go theJS route and introduce file naming convention for the different types.

| Type | Suggested File Suffix |
|--------|--------|
| Pattern | `.pattern.json` |
| Instantization of Pattern | `.instance.json` |
| Control Requirement | `.requirement.json` |
| Control Evidence | `.evidence.json` |
| Layout | `.layout.json` | 
| Flow | `.flow.json` | 

The benefit of having the file names describe what they are documenting would also allow for extra validation to occur - for example if I have a file called `.instance.json` I know it should have certain elements present.

Names are obviously a discussion point, just wanted to propose what the initial draft could be ( thanks @willosborne for helping with the list! )

willosborne:
Biased, but I like the second approach more as we discussed during the office hours.

What do people think about the control requirement/evidence part? Do we even want an extension for these domains? I know this was discussed on the call

yt-ms:
Could I suggest somehow making it less likely to have a collision on the extensions against other things using the same extensions (e.g. `.flow.json`)?  Something like `blah.calm-pattern.json` seems like a lot of typing, but maybe a compromise is something like `blah.cpattern.json` or `'blah.c-pattern.json`.

LeighFinegold:
With the definition of models https://github.com/finos/architecture-as-code/pull/902, its possible we don't really need this anymore. Do we still want to promote this as a best practice?

rocketstack-matt:
Does anyone think we still need this? It has been discussed for a long time and no one has ever found it important enough to actually implement.

With the evolution of CALM Hub, VSCode Extension and now CALM Workspaces I would argue this is just added complexity and propose we close the issue.

[...]

PR #1370, opened 2025-06-18
Changing calm dereferencing so that it will dereference all urls that are network-addressable and considered json schema (#1369)

LeighFinegold:
Fixing bug as described in detail on #1369


shared/src/docify/graphing/control-registry.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable  @typescript-eslint/no-explicit-any */
```
LeighFinegold, line 1: Unrelated to change - just removing the warning from lint

shared/src/docify/graphing/flow-sequence-helper.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable  @typescript-eslint/no-explicit-any */
```
LeighFinegold, line 1: Unrelated to change - just removing the warning from lint

shared/src/resolver/network-addressable-validator.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        return results;
+    }
+
+    /**
+     * Detects a JSON Schema _definition_ by presence of schema keywords.
+     * These keywords are mandated by a JSON Schema draft and unlikely in arbitrary documents.
+     */
+    private isJsonSchemaDefinition(doc: unknown): boolean {
```
LeighFinegold, line 53: Can anyone think of a better idea to distinguish between a schema and an implementation of a schema to just a json document which is not related to any schema?

shared/src/docify/docifier.e2e.spec.ts
LeighFinegold, line 1: The added tests prove now that dereferencing via the published workshop urls works the same way as the manual mapping.

shared/src/template/template-processor.e2e.spec.ts
```diff
@@ -258,36 +257,7 @@ describe('TemplateProcessor E2E', () => {
...
-        // 🧪 Compare file lists
-        expect(actualRelPaths).toEqual(expectedRelPaths);
-
-        // 🧪 Compare contents
-        for (const relPath of expectedRelPaths) {
-            const actualContent = await fsPromises.readFile(path.join(OUTPUT_DIR, relPath), 'utf-8');
-            const expectedContent = await fsPromises.readFile(path.join(EXPECTED_OUTPUT_WORKSHOP_DIR, relPath), 'utf-8');
-            expect(actualContent).toEqual(expectedContent);
-        }
+        await expectDirectoryMatch(EXPECTED_OUTPUT_WORKSHOP_DIR,OUTPUT_DIR);
```
LeighFinegold, line 260: "Workshop Infrastructure As Code Example" in template-processor.e2e.spec.ts still continues to work as before giving confidence the current walkthrough will continue to work

shared/src/template/template-calm-file-dereferencer.ts
```diff
@@ -15,39 +15,52 @@ export class TemplateCalmFileDereferencer {
...
+        const results = await validator.validate(entries);
+
+        const replacer = (data: unknown): unknown => {
+            if (typeof data === 'string' && /^https?:\/\//.test(data)) {
+                const entry = results.find(r => r.entry.value === data);
+                const mappedRef = this.urlFileMapping.get(data);
+                if (mappedRef) {
+                    return { $ref: mappedRef };
+                }
+                if (entry?.reachable && entry?.isSchemaImplementation) {
```
LeighFinegold, line 34: This is the key part for de-referencing. Can we enforce that folks who use document urls to reference  control-configuration or flows have atleast $schema defined?

LeighFinegold: Otherwise do we reference on file types as mentioned in 
https://github.com/finos/architecture-as-code/issues/490

shared/src/resolver/network-addressable-validator.ts
LeighFinegold: Atleast for the JSON schema validation, maybe I can use ajv?

shared/src/resolver/network-addressable-extractor.ts
```diff
@@ -0,0 +1,54 @@
...
+                pushUrls(keyPath, key, value, results);
+            } else if (value) {
+                traverse(value, keyPath, results);
+            }
+        }
+    }
+}
+
+function extractKey(path: string): string {
+    const match = path.match(/(?:\.([^.[\]]+)|\[(\d+)])$/);
```
willosborne, line 47: could you either pull this out to a named const or add a docstring - not really sure what this does

shared/src/resolver/network-addressable-validator.ts
willosborne: ironically an LLM would be great at this lol.
i think long term solution is some kind of convention but this more or less works for now

shared/src/resolver/network-addressable-extractor.ts
LeighFinegold: WIll do on next commit

PR #1374, opened 2025-06-20
Add support for flows inside pattern using const to a url reference + confirm generated arch document renders flow in docify

LeighFinegold:
# Changes to calm generate Function

The `instantiateFromProperties` function in the `instantiate.ts` file has been modified with three key changes:

## 1. Added Support for Constant Values in Array Items
This allows support in the pattern like this

```json
    "flows": {
      "type": "array",
      "minItems": 1,
      "maxItems": 1,
      "prefixItems": [
        {
          "const": "https://calm.finos.org/workshop/flows/conference-signup.flow.json"
        }
      ]
    }
```

## 2. Updated example on workshop to show generate works as expected and that subsequent docify command will render flow diagram like below

image

### DIFF
--- a/shared/src/commands/generate/components/instantiate.ts
+++ b/shared/src/commands/generate/components/instantiate.ts
@@ -101,12 +101,17 @@ async function instantiateFromProperties(
         const resolvedDef = await resolveSchema(def as JsonSchema, schemaDir);
 
         if (resolvedDef.type === 'array' && resolvedDef.prefixItems) {
-            output[key] = await Promise.all(resolvedDef.prefixItems.map(async (itemDef, idx) => {
-                const resolvedItem = await resolveSchema(itemDef, schemaDir);
-                return instantiateObject(resolvedItem, schemaDir, [key, `${idx}`]);
-            }));
+            output[key] = await Promise.all(
+                resolvedDef.prefixItems.map(async (itemDef, idx) => {
+                    const resolvedItem = await resolveSchema(itemDef, schemaDir);
+                    if (resolvedItem.const !== undefined) {
+                        return resolvedItem.const;
+                    }
+                    return await instantiateObject(resolvedItem, schemaDir, [key, `${idx}`]);
+                })
+            );
         } else {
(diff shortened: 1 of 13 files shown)


shared/src/commands/generate/components/instantiate.ts
```diff
@@ -101,12 +101,17 @@ async function instantiateFromProperties(
...
 
         if (resolvedDef.type === 'array' && resolvedDef.prefixItems) {
-            output[key] = await Promise.all(resolvedDef.prefixItems.map(async (itemDef, idx) => {
-                const resolvedItem = await resolveSchema(itemDef, schemaDir);
-                return instantiateObject(resolvedItem, schemaDir, [key, `${idx}`]);
-            }));
+            output[key] = await Promise.all(
+                resolvedDef.prefixItems.map(async (itemDef, idx) => {
+                    const resolvedItem = await resolveSchema(itemDef, schemaDir);
+                    if (resolvedItem.const !== undefined) {
```
LeighFinegold, line 107: Doing 
```javascript
if (resolvedItem.const !== undefined) {
    return resolvedItem.const;
}
```
over 
```javascript
if (resolvedItem.const) {
    return resolvedItem.const;
}
```

Allows support for 
```json
{ const: null }
```

shared/src/docify/docifier.e2e.spec.ts
```diff
@@ -26,15 +26,22 @@ describe('Docifier E2E - Real Model and Template', () => {
 
     it('generates documentation from the conference-signup.arch.json model', async () => {
 
-        const docifier = new Docifier('WEBSITE', join(INPUT_DIR, 'conference-signup.arch.json'), NON_SECURE_VERSION_DOC_WEBSITE, new Map<string,string>());
+        //TODO:  This test needs manual mapping until flows folder is republished.
```
LeighFinegold: If folks are happy with this PR, then recommend we first just publish the new flows document to S3 and then this will not be needed

jpgough-ms (dismissed):

jpgough-ms:
@LeighFinegold conflicts here, going to move to draft until fixed

PR #1378, opened 2025-06-21
Update the getting-started guide for OSFF.

rocketstack-matt:
Note there are two copies one in the calm/getting-started/README and one in the docs project for the Docusarusus portal. This is because the have different requirements in terms of shell commands and image locations.

PR #1379, opened 2025-06-22
Prevent minifying removing constructor names as needed for relationship rendering via isInstanceOf template helper. Adding a full test suite to run against latest cli specification

LeighFinegold:
This resolves an issue where our Handlebars instanceOf(value, className) helper was failing in production because esbuild (via tsup) was minifying and mangling class and function names. By enabling keepNames: true in our tsup.config.ts, we ensure constructor.name remains stable, restoring correct behavior of template conditionals that branch on relationship types (e.g. CalmComposedOfType, CalmInteractsType, etc.)

To ensure getting-started guide works as we progress CALM, I have added an additional e2e suite for the cli related steps


shared/tsup.config.ts
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
     ],
     format: ['esm'],
     clean: true,
+    keepNames: true,
```
LeighFinegold, line 9: This was the line fix to ensure constructor names are maintained and does not break the handlebar templating for relatedNotes section

cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@finos/calm-cli",
-    "version": "0.7.7",
+    "version": "0.7.8",
```
LeighFinegold, line 3: Prepping another cli with the minify fix

LeighFinegold:
@rocketstack-matt  - updated this PR to take into account your simplified getting started README on https://github.com/finos/architecture-as-code/pull/1378. Removed the check-in website folder. 

Post this PR we just need to decide if we want to add flows directly to the project and publish or make it an additional step for people to local about the additional local directory option. That won't need another cli version though

rocketstack-matt (approved):

PR #1385, opened 2025-06-23
Update Getting Started Docs Prerequisites

aidanm3341:
more specific node version in getting started prereqs because node 22 doesn't work


aidanm3341:
seems to not be an issue for everyone, so I must be missing something else. Declining this for now.

PR #1397, opened 2025-07-01
Validate refactor part 1 - move loading of files out to CLI

willosborne:
This is the first stage before I refactor to inject schema directory properly

### DIFF
--- a/cli/src/cli.e2e.spec.ts
+++ b/cli/src/cli.e2e.spec.ts
@@ -369,7 +369,7 @@ describe('CLI Integration Tests', () => {
     }
 
     // Utility to recursively remove specific line/character fields from JSON
-    function removeLineNumbers(obj: any) {
+    function removeLineNumbers(obj: object) {
         const fieldsToRemove = ['line_start', 'line_end', 'character_start', 'character_end'];
         if (Array.isArray(obj)) {
             obj.forEach(removeLineNumbers);
--- a/cli/src/cli.e2e.spec.ts
+++ b/cli/src/cli.e2e.spec.ts
@@ -75,7 +75,10 @@ describe('CLI Integration Tests', () => {
         const { stdout } = await execPromise(cmd);
 
         const expected = JSON.parse(fs.readFileSync(path.join(__dirname, '../test_fixtures/validate_output.json'), 'utf8'));
-        expect(JSON.parse(stdout)).toEqual(expected);
+        const parsedOutput = JSON.parse(stdout);
+        removeLineNumbers(parsedOutput);
+        removeLineNumbers(expected);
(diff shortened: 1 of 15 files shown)


cli/src/cli.e2e.spec.ts
```diff
@@ -358,5 +368,21 @@ describe('CLI Integration Tests', () => {
         writeJson(filePath, obj);
     }
 
+    // Utility to recursively remove specific line/character fields from JSON
```
willosborne, line 371: This is a temporary workaround. The fundamental problem is that line numbers have never been correctly tracked - as the file is loaded a lot of information is lost since it's parsed to/from JSON. But this loading mechanism seems to fully discard the line number information.

We should instead enrich this after the fact using the JSONPath information we have available - won't do that on this PR

cli/src/server/routes/validation-route.ts
```diff
@@ -61,31 +57,31 @@ export class ValidationRouter {
...
-async function createTemporaryFile(): Promise<string> {
-    const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'calm-'));
-    const tempFilePath = path.join(tempDir, `calm-instantiation-${uuidv4()}.json`);
-    return tempFilePath;
-}
+// async function createTemporaryFile(): Promise<string> {
+//     const tempDir = await fs.mkdtemp(path.join(os.tmpdir(), 'calm-'));
+//     const tempFilePath = path.join(tempDir, `calm-instantiation-${uuidv4()}.json`);
+//     return tempFilePath;
+// }
```
aidanm3341: can we delete this commented out stuff?

willosborne: Done

shared/src/commands/validate/validate.ts
```diff
@@ -118,27 +120,27 @@ async function runSpectralValidations(
...
+ * @param pattern The pattern as a JS object
+ * @param metaSchemaPath File path to meta schema directory
+ * @param debug Whether to log at debug level
+ * @returns Validation report
  */
 export async function validate(
-    jsonSchemaArchitectureLocation: string,
-    jsonSchemaLocation: string,
+    architecture: object,
+    pattern: object,
```
aidanm3341, line 131: given that on Line 138 we are checking for these being truthy, is the intention that these should be optional params?

```typescript
architecture?: object,
pattern?: object,
```

aidanm3341 (approved):

shared/src/commands/validate/validate.ts
willosborne: Good shout, will do on next PR

Issue #1418, opened 2025-07-13
Bug: Support Recursive detailed-architecture dereferencing


LeighFinegold:
## Summary

Architectural nodes include a **details** block where you can link to a more detailed view:

```json
{
  "details": {
    "detailed-architecture": "http://calm-hub/architecture-5",
    "required-pattern":    "http://calm-hub/pattern-2"
  }
}
```

Today, our dereferencer within calm-template tooling inlines that first-level URL but stops if the referenced document itself uses `detailed-architecture`. We need **full recursion** to handle deeply nested systems.

## Problem

* **Current behavior**: Resolver pulls in exactly one `detailed-architecture` link, then leaves any further URLs intact.
* **Edge case**: In microservices or complex domains, each sub-architecture can point to another, e.g.:

  1. Service → Architecture A → Architecture B → ...
  2. Only A is inlined; B and beyond remain unresolved URLs.

## Expected Behavior

1. Fetch and inline **all** `detailed-architecture` URLs at any depth.
2. For each inlined document, resolve its own sub-links until no further pointers remain.
3. Return a single JSON tree with **no** dangling `detailed-architecture` URLs.

## Example

**Top-level input**:

```json
{
  "unique-id": "service-123",
  "details": { "detailed-architecture": "http://hub/arch-A" }
}
```

**`http://hub/arch-A`** contains:

```json
{ "details": { "detailed-architecture": "http://hub/arch-B" } }
```

**`http://hub/arch-B`** is the leaf architecture:

```json
{ "nodes": [ ... ] }
```

**Desired output**:

```json
{
  "unique-id": "service-123",
  "details": {
    "detailed-architecture": {
      "details": {
        "detailed-architecture": {
          "nodes": [ ... ]
        }
      }
    }
  }
}
```

## Proposed Fix

* Update `replaceUrls` (or equivalent) to **recurse** immediately after inlining each URL.
* Optionally, loop until no further URLs are found, rather than a fixed-depth pass count.

This ensures our templating pipeline can handle arbitrarily nested architectures without manual unrolling.

PR #1419, opened 2025-07-13
Add fix to allow for recursive url dereferencing (#1418)

LeighFinegold:
This will enable templates access to the full architecture when building one page documents

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,11 +4,11 @@
 
 /calm-hub/ @jpgough-ms @rocketstack-matt @grahampacker-ms @Thels
 
-/cli/ @aidanm3341 @lbulanti-ms @willosborne @grahampacker-ms @jpgough-ms @rocketstack-matt @Thels @LeighFinegold
+/cli/ @aidanm3341 @lbulanti-ms @willosborne @grahampacker-ms @jpgough-ms @rocketstack-matt @Thels @LeighFinegold @markscott-ms
 
 /calm-visualizer/ @oliviajanejohns @aidanm3341 @jpgough-ms @rocketstack-matt
 
-/shared/ @aidanm3341 @lbulanti-ms @willosborne @grahampacker-ms @jpgough-ms @rocketstack-matt @Thels @LeighFinegold
+/shared/ @aidanm3341 @lbulanti-ms @willosborne @grahampacker-ms @jpgough-ms @rocketstack-matt @Thels @LeighFinegold @markscott-ms
 
 # Ownership for dependency-related files at the root level
 /package.json @finos/architecture-as-code-maintainers


shared/src/template/template-calm-file-dereferencer.ts
```diff
@@ -1,71 +1,120 @@
...
+                        logger.debug(`parser.read: resolving ${file.url}`);
+                        return JSON.stringify(await this.resolver.resolve(file.url));
+                    }
+                }
+            }
+        });
 
-        const entries: AddressableEntry[] = extractNetworkAddressables(doc);
+        let final: unknown = dereferenced;
+        for (let pass = 1; pass <= 5; pass++) {
```
markscott-ms: Why 5 passes?

Is this instead of loop detection?

LeighFinegold: I guess at the time, I was thinking noone is going to build architecture references that are more than 3 deep so 5 was buffer :). That said, can work more on changing this to go infinately but then keeping track of urls that could not dereference on each pass so I don't repeat.

shared/src/template/template-calm-file-dereferencer.ts
```diff
@@ -1,71 +1,126 @@
...
-                const entry = results.find(r => r.entry.value === data);
-                const mappedRef = this.urlFileMapping.get(data);
-                if (mappedRef) {
-                    return { $ref: mappedRef };
+        const walk = async (node: unknown): Promise<unknown> => {
+            if (typeof node === 'string' && /^https?:\/\//.test(node)) {
+                logger.debug(`replaceUrls.walk: examining URL ${node}`);
+
+                if (seen.has(node)) {
+                    logger.debug(`replaceUrls.walk: skipping seen URL ${node}`);
```
markscott-ms, line 81: Does this new implementation handle two nodes with the same detailed architecture, or would it only inline the first?

For example this pseudo-arch:
```
doc1:
   node1 { detailed-arch: doc2.json }
   node2 { detailed-arch: doc2.json }
   node3 { load balancer }
```

LeighFinegold: I think the above scenario works because its still using $RefParser.dereference. However a scenario that would not work is whether a nested document also has a reference to something that has been deferenced already. 

I'm not sure that is a valid scenario or whether even the above example you give should be valid. Partly because the whole idea is the detailed-architecture expands the current node details. 

If we want to support that then will end up maintaining a cache. At that point, we should consider deprecating the use of the json-ref-parser library

markscott-ms: Ok. What you have is a good evolution of the current situation. We can refine it further in the future if usage requires it.

markscott-ms (dismissed):

aidanm3341 (approved):

PR #1420, opened 2025-07-14
Review of calm schema 1.0-rc2

markscott-ms:
# Top level element type

It is unclear whether the type of the top level element of a JSON needs defined.

Some sources suggest that the presence of `properties` implicitly means that the element is an `object`. Some suggest that a validator might accept a non-object.

# Required properties

Reviewed the required properties and identified three issues:

- ability to specify that a node has details, but not be required to specify those details
- ability to specify that there is a deployed-in relationship somewhere, but not be required to specify between which nodes.
- a property name incorrect in evidence.json

### DIFF
--- a/calm/release/1.0-rc2/meta/core.json
+++ b/calm/release/1.0-rc2/meta/core.json
@@ -62,10 +62,6 @@
               "type": "string"
             }
           },
-          "anyOf" : [
-            { "required" : [ "detailed-architecture" ] },
-            { "required" : [ "required-pattern" ] }
-          ],
           "additionalProperties": false
         },
         "interfaces": {


calm/release/1.0-rc2/meta/core.json
```diff
@@ -61,6 +62,7 @@
               "type": "string"
             }
           },
+          "required" : [ "detailed-architecture" ],
```
rocketstack-matt: I don't think we want to make this mandatory . . . nodes are at an arbitrary level of the architecture - for example they could represent an ecosystem, a system, a service, etc. At some point they cannot be further decomposed and therefore you cannot provide a more detailed version.

Also, in many architectures, you don't care about the detailed architecture or aren't able to see into it (e.g. external systems.

markscott-ms: Agreed, but if someone has specified that a node has “details”, we should expect that some details be provided.

“details” has two properties - another architecture and a pattern. Pattern is optional, but we should always require the detailed architecture.

* node.details is optional
* node.details.detailed-architecture is required, if and only if node.details has been specified.
* node.details.required-pattern is optional

Valid: `"node": { "unique-id": ..., "name": ..., "description": ..., "node-type": ... }`
Invalid: `"node": { "unique-id": ..., "name": ..., "description": ..., "node-type": ..., "details": {}}`
Valid: `"node": { "unique-id": ..., "name": ..., "description": ..., "node-type": ..., "details": {"detailed-architecture": ...} }`

LeighFinegold: If I understand correctly from [anyof
](https://json-schema.org/understanding-json-schema/reference/conditionals#implication) example on json-schema.org 

We would want


```json
"details": {
  "type": "object",
  "properties": {
    "detailed-architecture": { "type": "string" },
    "required-pattern": { "type": "string" }
  },
  "anyOf": [
    { "required": ["detailed-architecture"] },
    { "required": ["required-pattern"] }
  ],
  "additionalProperties": false
}
```

Or atleast this is how i've seen it modelled on https://github.com/rjsf-team/react-jsonschema-form/issues/3834

markscott-ms: @grahampacker-ms and I discussed this morning (at the same time you wrote that @LeighFinegold ) and concur.

rocketstack-matt: Is it anyOf or if a pattern is provided do we actually then want to mandate that detailed-architecture is provided, which works simply by having detailed-architecture as mandatory and pattern as optional . . . this allows you to only provide a detailed architecture but also enforces that if you supply a pattern you must also have the architecture to support that you're using that pattern.

Is there a use case where we would not want to see the architecture if a pattern were supplied?

markscott-ms: Given the uncertainty over cases where we may have a pattern and not an architecture, and that it makes the schema _more_ restrictive, I suggest I do not add this validation at this point. We can potentially add in a later version of the schema after real world observation?

Issue #1240, opened 2025-05-01
Clean up CALM schema project


rocketstack-matt:
# CALM Schema Directory Cleanup

## Overview
This issue proposes a cleanup of the `calm` directory structure to improve organization, remove deprecated examples, and ensure a more maintainable schema repository. The recent introduction of the `prototype` directories within schema proposals provides a more consistent approach for examples, making several older directories redundant.

## Goals
- Simplify the directory structure
- Remove deprecated examples and redundant content
- Create a dedicated project for multiple workshops/demos 
- Preserve and migrate valuable TraderX reference architecture
- Ensure proper handling of any dependencies

### 1. Create New Standardized Repositories

Create two new folders at the top level of the calm directory to serve as golden sources for FINOS-published standard components:

- `calm/controls/` - Official repository of standardized control definitions
  - Will house validated, approved control requirements
  - Organized by domains (security, performance, compliance, etc.)
  - Versioned appropriately
  - Published to calm-hub as reference controls

- `calm/interfaces/` - Official repository of standardized interface definitions
  - Will contain validated, approved interface types
  - Organized by technology domains and communication patterns
  - Versioned appropriately
  - Published to calm-hub as reference interfaces

These repositories will form the foundation for a library of reusable, standardized components that organizations can use directly or extend for their specific needs.

### 2. Delete Deprecated Example Directories

The following directories contain outdated examples that have been superseded by the `prototype` directories in draft schema proposals:

- `calm/control-example/` - Contains old control examples
- `calm/domains-example/` - Contains domain-specific examples
- `calm/samples/` - Contains miscellaneous samples, except for the TraderX content

These should be removed as they no longer represent current best practices. Most valuable examples have been migrated to the `prototype` directories in corresponding schema proposals.

### Special Handling for TraderX Content

The TraderX CALM definition currently exists in:
- `calm/samples/2024-10/traderx/`
- `calm/samples/2024-12/traderx/`
- `calm/draft/2025-03/samples/traderx/`

This content represents a valuable reference architecture and should be preserved. Instead of deletion, we should:

1. Enhance the TraderX CALM definition to showcase the latest schema features
2. Migrate the enhanced definition to the TraderX repository
3. Ensure the migrated content is well-documented and accessible
4. Update any references to the old locations

### 2. Analyze and Address Pattern Directory Dependencies

The `calm/pattern/` directory is currently referenced by CLI tests:

```
cli/src/server/routes/validation-route.spec.ts
[...]

PR #1427, opened 2025-07-17
Delete old sample folders per #1240

rocketstack-matt:
The issue says we will handle traders separately, but I think it is worth rebuilding that from scratch so have removed along with the other examples.


LeighFinegold (approved):

cli/test_fixtures/api-gateway/api-gateway.json
```diff
@@ -0,0 +1,244 @@
+{
```
LeighFinegold, line 1: Might move this later, but can do in seperate PR

PR #1430, opened 2025-07-17
Inject schemadirectory into validate and add fallback loading mechanism 

willosborne:
This brings validate in line with generate and also adds the ability to fall back across multiple types of loading
This means file loading is all done via the DocumentLoader rather than separately.
A new type, MultiStrategyDocumentLoader, will take a list of other loaders and try them in order.

This also greatly improves validation tests by properly abstracting the AJV module into a JsonSchemaValidator type.

### DIFF
--- a/cli/src/command-helpers/generate-options.ts
+++ b/cli/src/command-helpers/generate-options.ts
@@ -51,7 +51,7 @@ export async function promptUserForOptions(pattern: object, debug: boolean = fal
     const logger = initLogger(debug, 'calm-generate-options');
 
     const patternOptions: CalmOption[] = extractOptions(pattern);
-    logger.debug('Pattern options extracted from pattern: ' + patternOptions);
+    logger.debug('Pattern options extracted from pattern: ' + JSON.stringify(patternOptions));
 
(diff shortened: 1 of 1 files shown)


cli/src/command-helpers/generate-options.ts
```diff
@@ -51,8 +51,8 @@ export async function promptUserForOptions(pattern: object, debug: boolean = fal
     const logger = initLogger(debug, 'calm-generate-options');
 
     const patternOptions: CalmOption[] = extractOptions(pattern);
-    logger.debug('Pattern options extracted from pattern: [%O]', patternOptions);
```
aidanm3341, line 54: I seem to remember doing it this way for a reason. Can you double check what this output now looks like? I have a feeling with this change the output will just be something like `[Object object]` rather than actually printing the object. If so, a better approach might be doing `JSON.Stringify` though, rather than what I had originally done here

willosborne: You're right. Fixed

aidanm3341 (approved):

PR #1433, opened 2025-07-17
Adding draft schemas to CLI bundle to keep the CLI backwards compatible when new versions of the schema are promoted

grahampacker-ms:
Adding draft schemas to CLI bundle to keep the CLI backwards compatible when new versions of the schema are promoted


grahampacker-ms:
I've pulled in an extra commit, will reraise

PR #1434, opened 2025-07-17
Including older versions of schema in the CLI bundle to keep the CLI backwards compatible (#1423)

### LINKED ISSUE #1423: Existing Patterns break with every schema change
## Problem 
Every time we're ready to release a new version of the CALM schema, we update the 'copy-calm-schema' script to point to the new version https://github.com/finos/architecture-as-code/blob/main/cli/package.json#L15

When we do this, the old version is no longer bundled with the CLI. The effect of this, is that any patterns or architectures which reference that old version of the schema can no longer be generated or validated using the CLI, because the correct version doesn't exist in the SchemaDirectory anymore.

## Proposed Solution
Ideally, we should bundle previous versions of the schema too in order to maintain backwards compatibility. We don't really lose anything by doing this except for slightly increasing the size of the CLI package.

grahampacker-ms:
fixes #1423

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@finos/calm-cli",
-    "version": "0.7.9",
+    "version": "0.7.10",
     "description": "A set of tools for interacting with the Common Architecture Language Model (CALM)",
     "homepage": "https://calm.finos.org",
     "repository": {
@@ -17,7 +17,7 @@
     "scripts": {
         "build": "tsup && npm run copy-calm-schema && npm run copy-docify-templates",
         "watch": "node watch.mjs",
-        "copy-calm-schema": "copyfiles \"../calm/release/1.0-rc1/meta/*\" dist/calm/",
+        "copy-calm-schema": "copyfiles \"../calm/release/**/meta/*\" \"../calm/draft/**/meta/*\" dist/calm/",
         "copy-docify-templates": "copyfiles \"../shared/dist/template-bundles/**/*\" dist --up 3",
         "test": "vitest run",
         "lint": "eslint src",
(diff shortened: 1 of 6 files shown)


cli/package.json
```diff
@@ -17,7 +17,7 @@
     "scripts": {
         "build": "tsup && npm run copy-calm-schema && npm run copy-docify-templates",
         "watch": "node watch.mjs",
-        "copy-calm-schema": "copyfiles \"../calm/release/1.0-rc1/meta/*\" dist/calm/",
+        "copy-calm-schema": "copyfiles \"../calm/release/**/meta/*\" \"../calm/draft/**/meta/*\" dist/calm/",
```
rocketstack-matt, line 20: Do we want to implicitly include all drafts? All releases I get, but this could get very bulky and we make no guarantee that drafts are supported by the CLI.

Further more, this is going to one open to breaking anyway, as deleting a draft folder (which we will once a release including drafts is deployed) will remove it from versions of the cli without notice.

Whilst I accept it duplicates code, I suspect we should be physically copying supported schemas into the CLI directly.

rocketstack-matt (changes requested):

cli/package.json
rocketstack-matt: Thinking further on my second point - maybe it's ok (given we're a monorepo) that versions come out, although we should check that at least gets accurately reflected in release notes.

The other argument for copying supported versions in is that we can (and should) explicitly write tests against them. If we don't copy in, we should clearly document the procedure for ensuring the CLI is fully compliant with at least releases as part of the schema release and CLI release documentation.

grahampacker-ms: I would agree, but we're currently in a situation where there are uses of multiple drafts which is causing issues to real users today. I think once we've gave people time to move on to the actual release version, we can be more strict on only including schemas in the "release" folder.

rocketstack-matt: Approving as discussed offline, but with the understanding without a different process, this could still break without notice

Issue #1284, opened 2025-05-15
Validate & Approve 1.0 release for CLI


rocketstack-matt:
Approvers: @willosborne / @LeighFinegold

PR #1447, opened 2025-07-20
Update types, models and tooling to meet RC2 definition (#1284)

LeighFinegold:


calm/release/1.0-rc2/meta/core.json
```diff
@@ -67,7 +67,7 @@
         "interfaces": {
           "type": "array",
           "items": {
-            "oneOf": [
+            "anyOf": [
```
markscott-ms, line 70: This change is necessary because every `interface-definition` (`unique_id` + `definition_id` + `config` properties) would also validate against `interface-type` (`unique-id` plus any other properties), which would violate the `oneOf` constraint.

calm/release/1.0-rc2/meta/interface.json
```diff
@@ -32,7 +32,8 @@
       },
       "required": [
         "unique-id"
-      ]
+      ],
+      "additionalProperties": true
```
markscott-ms, line 36: Whilst `true` is the default, being explicit gives clarity to the reader and future developers.

rocketstack-matt (approved):

PR #1450, opened 2025-07-21
tests for model/interface for 1.0-rc2; clarify interface model construction   (finos#1284)

markscott-ms:
## Interfaces

* Tightened conditions on `CalmInterfaceDefinition` subclass of `CalmInterface` being constructed so it is only constructed when ONLY `unique-id`, `definition-url` and `config` are provided and nothing else, as the 1.0-rc2 schema has `additionalProperties: false`.
* If extra properties exist, then it is constructed as a `CalmInterfaceType` subclass of `CalmInterface`.

## Flows

* Corrected and added additional tests for flow model.

## Controls

* Additional test expectation for control model

## Nodes

* Add support for additional properties on the node model, and optional properties in a node detail.
* Additional tests for nodes model

## Relationships

* Fix type/model for the `options` relationship type to match schema (options is array of decision, not array of array of decision)
* Add support for additional properties on the relationship model.

### DIFF
--- a/shared/src/model/relationship.ts
+++ b/shared/src/model/relationship.ts
@@ -5,11 +5,10 @@ import {
     CalmConnectsRelationshipSchema, CalmDecisionSchema, CalmDeployedInRelationshipSchema,
     CalmInteractsRelationshipSchema,
     CalmOptionsRelationshipSchema,
-    CalmOptionTypeSchema,
     CalmRelationshipSchema,
     CalmRelationshipTypeSchema
 } from '../types/core-types.js';
-import {CalmNodeInterface} from './interface.js';
+import { CalmNodeInterface } from './interface.js';
 
 
 export class CalmRelationship {
@@ -18,18 +17,28 @@ export class CalmRelationship {
         public relationshipType: CalmRelationshipType,
         public metadata: CalmMetadata,
         public controls: CalmControl[],
+        public additionalProperties: Record<string, unknown>,
         public description?: string,
         public protocol?: string
-    ) {}
+    ) { }
 
     static fromJson(data: CalmRelationshipSchema): CalmRelationship {
+        const {
+            'unique-id': uniqueId,
+            'description': description,
+            'relationship-type': relationshipType,
+            'protocol': protocol,
+            'metadata': metadata,
+            'controls': controls,
+            ...additionalProperties } = data;
         return new CalmRelationship(
-            data['unique-id'],
-            CalmRelationship.deriveRelationshipType(data['relationship-type']),
-            data.metadata ? CalmMetadata.fromJson(data.metadata) : new CalmMetadata({}),
-            CalmControl.fromJson(data.controls),
-            data.description,
-            data.protocol
+            uniqueId,
+            CalmRelationship.deriveRelationshipType(relationshipType),
+            metadata ? CalmMetadata.fromJson(metadata) : new CalmMetadata({}),
+            CalmControl.fromJson(controls),
+            additionalProperties,
+            description,
+            protocol
         );
     }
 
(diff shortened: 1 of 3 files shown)


shared/src/model/interface.ts
```diff
@@ -2,13 +2,14 @@ import {
...
+import { CalmInterfaceSchema } from '../types/core-types.js';
 
 export class CalmInterface {
-    constructor(public uniqueId: string) {}
+    constructor(public uniqueId: string) { }
 
     static fromJson(data: CalmInterfaceSchema): CalmInterface {
-        if ('definition-url' in data && 'config' in data) {
+        if ('unique-id' in data && 'definition-url' in data && 'config' in data
+            && Object.keys(data).length === 3) {
```
Thels: This feels brittle - can we instead define the keys that we expect - and then do a length of that? So that if we add a new piece of data we don't need to go about modifying the magic numbers outside of where but the test.

shared/src/model/interface.spec.ts
```diff
@@ -59,6 +59,31 @@ describe('CalmInterface.fromJson', () => {
...
+                config: { alpha: true, threshold: 5 },
+                extraField: 'extraValue'
+            };
+
+            const iface = CalmInterface.fromJson(defData);
+            expect(iface).toBeInstanceOf(CalmInterfaceType);
+
+            // Narrow via instanceof so TypeScript recognizes the subclass
+            if (!(iface instanceof CalmInterfaceType)) {
+                throw new Error('Expected CalmInterfaceType');
```
Thels, line 76: Can this be reached given the assert above? Or is this literally to avoid linting issues and the IDE telling you that it isn't the correct type?

markscott-ms: It won’t be reached but existing tests in this file had this extra instanceof check that comments led me to believe way to make a linter/static analyser more happy

shared/src/model/interface.ts
markscott-ms: Okay, so define an array of the three keys and then test against that. Makes sense

shared/src/model/relationship.ts
```diff
@@ -18,18 +17,28 @@ export class CalmRelationship {
...
-            CalmRelationship.deriveRelationshipType(data['relationship-type']),
-            data.metadata ? CalmMetadata.fromJson(data.metadata) : new CalmMetadata({}),
-            CalmControl.fromJson(data.controls),
-            data.description,
-            data.protocol
+            uniqueId,
+            CalmRelationship.deriveRelationshipType(relationshipType),
+            metadata ? CalmMetadata.fromJson(metadata) : new CalmMetadata({}),
+            CalmControl.fromJson(controls),
+            additionalProperties,
```
LeighFinegold, line 39: It would be nicer if this could be last  argument, but I assume its practice to leave optionals towards the tail end of constructor.

shared/src/model/interface.ts
```diff
@@ -2,13 +2,22 @@ import {
     CalmNodeInterfaceSchema
 } from '../types/interface-types.js';
 import { CalmInterfaceDefinitionSchema, CalmInterfaceTypeSchema } from '../types/interface-types.js';
-import { CalmInterfaceSchema }           from '../types/core-types.js';
+import { CalmInterfaceSchema } from '../types/core-types.js';
+
+const calmInterfaceDefinitionRequiredProperties = [
```
LeighFinegold, line 7: Could we not just put the properties in the order :D

shared/src/model/relationship.ts
markscott-ms: It's not just practice, it's required to have optional arguments at the end.

I initially tried to put it at the end but was complained at by the IDE.

shared/src/model/interface.ts
markscott-ms: Yes, but I didn't want the test to be flaky in the future. I wanted to explicitly say this is a sorted list, as we need to sort the actual and desired keys in order to check that the two lists contain the same keys.

LeighFinegold (approved):

Issue #1453, opened 2025-07-24
Office Hours - 24/07/2025


Thels:
## 24/07/2025

Thursday 24th July 2025 - 15:30 BST

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

Office Hours is a meeting for maintainers to get together to discuss active development and upcoming priorities.

- [x] Confirm scope: https://github.com/finos/architecture-as-code/issues/1452 @LeighFinegold 
- [x] Document Dereferencing: https://github.com/finos/architecture-as-code/issues/1369 and related https://github.com/finos/architecture-as-code/issues/490
- [x] #1456 
- [x] Confirm Prioritisation: https://github.com/finos/architecture-as-code/issues/1116 @LeighFinegold 
- [x] Dependency PR management - PRs are still not being proactively reviewed. We need folks to take better ownership of proactive review and merge. Appreciate most open PRs atm have build failures, but most if not all of these are the Quarkus related CVEs whereas the upgrades are aimed at other projects.
- [x] Confirm approach for interface types referencing `1.0-rc1` in materials and tests that use `1.0-rc2` schema, such as https://github.com/finos/architecture-as-code/blob/717350bec736a7f931c7c09df6b0b0b56e51612f/calm/getting-started/conference-signup.pattern.json#L35

### Zoom info

Join Zoom Meeting
[...]

Issue #1238, opened 2025-05-01
CALM Schema v1.0 Release Candidate


rocketstack-matt:
### Proposal Type:
Release Candidate - Proposing to merge draft/2025-03 and draft/1083 into release version 1.0-rc1

### Description of Release:

This proposal aims to create the first Release Candidate (RC) for CALM version 1.0, which will combine the changes from draft/2025-03 (current primary draft) and draft/1083 (interface enhancements). This release candidate would be the first step toward a stable 1.0 release of the CALM schema, following the governance process defined in issue #893.

### Included Changes:

1. **All schema updates from draft/2025-03**, which include:
   - Core schema definitions for nodes, relationships, and other architecture elements
   - Control requirements and configuration mechanisms
   - Flow definitions for representing business processes

2. **Interface enhancements from draft/1083**:
   - Modular interface definition model
   - Ability to define interface configurations inline
   - Enhanced interface type system

### Migration Impact:

- This is the first official release candidate, so there is no migration from previous releases
- Users currently working with draft/2025-03 will need minimal changes
- Users leveraging draft/1083 interface enhancements will be fully compatible
- Future releases will need to maintain backward compatibility with this RC or provide clear migration paths

#### Schema Validation:
- All schema files must validate against JSON Schema Draft 2020-12
- Cross-references between schema files must be consistent
- No circular dependencies in schema definitions

#### Tool Compatibility:
- CALM CLI must be tested with the RC schemas for validation, conversion, and other operations
- CALM Hub must be tested for storing, retrieving, and visualizing architectures using the RC schemas
- Any other tools in the ecosystem should be tested and updated as needed

#### Example Verification:
- Existing example architectures must validate against the new schema
- New examples demonstrating the combined features should be created
- Edge cases should be documented and tested

### Testing Period:

As defined in the governance process (issue #893), this release candidate will undergo a 4-week testing period:

1. Week 1-2: Initial testing by tool maintainers
2. Week 3-4: Community testing and feedback
3. End of Week 4: Decision to promote to final release or create a new RC

### Implementation Plan:

1. Create a new folder structure in `calm/releases/1.0-rc1/`
2. Copy all schemas from draft/2025-03 to this folder
3. Integrate interface enhancements from draft/1083
4. Update all references and $id fields to point to the correct release paths
5. Create a changelog documenting all features included in this release
6. Update documentation to reflect the RC status and testing requirements

### Documentation Updates:

- Create release notes detailing all features in 1.0-rc1
[...]

PR #1464, opened 2025-07-24
Add example interface type definitions for CALM examples to refer to

markscott-ms:
Need to add before referring to them so that `calm generate` can reference. #1238

This is work required to remove 1.0-rc1 references from examples such as the getting-started guide, which has the `conference-signup.pattern.json` use 1.0-rc2 schema, apart from its interfaces.

Our example guides should always use the latest and greatest.

Was discussed in #1453 with @rocketstack-matt that we should add some interface types to the `calm/interfaces` folder to be used in examples.

When/If we consider these interface definitions more than fit for examples we could change the folder name.

### DIFF
--- a/calm/interfaces/example/README.md
+++ b/calm/interfaces/example/README.md
@@ -1,0 +1,5 @@
+# Example interfaces
+
+This directory contains interface types that are considered examples and may only be partially well-formed.
+
+References to them should only be made from within the `finos/architecture-as-code` repository examples as they may change without warning.
--- a/calm/interfaces/example/container-image.json
+++ b/calm/interfaces/example/container-image.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://calm.finos.org/interfaces/example/container-image",
+  "title": "Container Image Interface",
+  "type": "object",
+  "properties": {
+    "image": {
+      "type": "string",
+      "description": "Location of the container image, for example in the format 'registry/repository:tag'."
+    }
+  },
+  "required": ["image"],
+  "examples": [
+    {
+      "image": "docker.io/library/nginx:latest"
+    },
+    {
+      "image": "calm.finos.org/images/repository:version"
+    }
+  ]
+}
--- a/calm/interfaces/example/tcp-host-port.json
+++ b/calm/interfaces/example/tcp-host-port.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://calm.finos.org/interfaces/example/tcp-host-port",
+  "title": "TCP Host / Port Interface",
+  "type": "object",
+  "properties": {
+    "host": {
+      "type": "string",
+      "format": "hostname",
+      "description": "The hostname or IP"
+    },
+    "port": {
+      "type": "integer",
+      "description": "The port on which connections are accepted",
+      "minimum": 1,
+      "maximum": 65535
+    }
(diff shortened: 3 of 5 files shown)


calm/interfaces/example/README.md
```diff
@@ -0,0 +1,5 @@
+# Example interfaces
+
+This directory contains interface types that are considered examples and may only be partially well-formed.
+
+References to them should only be made from within the `finos/architecture-as-code` repository examples as they may change without warning.
```
LeighFinegold, line 5: I like this statement.

calm/interfaces/example/README.md
```diff
@@ -0,0 +1,5 @@
+# Example interfaces
+
+This directory contains interface types that are considered examples and may only be partially well-formed.
```
LeighFinegold: Do we want to make it that they are not partially well formed? what is missing?

markscott-ms: Clarified wording to describe them as maybe not as well defined as they could be.

LeighFinegold: Just thought - what is the intention of these specific interfaces? and how do they differ the prototypes which in the calm release folder?  https://github.com/finos/architecture-as-code/tree/main/calm/release/1.0-rc2/prototype/interfaces

I know this specific calm/interfaces and calm/controls were going to contain potentially CALM recommended interfaces/controls.

markscott-ms: those prototypes in the calm release folder are used by https://github.com/finos/architecture-as-code/tree/main/calm/release/1.0-rc2/prototype/custom-interface-example.json. For some reason I thought they were elsewhere and inaccessible.

The interfaces in this PR are for the getting-started conference signup pattern to refer to.

Do we need to clarify with @rocketstack-matt given the discussion during #1453 ?

.github/workflows/s3-sync.yml
```diff
@@ -8,6 +8,7 @@ on:
         paths:
             - 'calm/draft/**'
             - 'calm/getting-started/**'
+            - 'calm/interfaces/**'
```
rocketstack-matt, line 12: Can we add `calm/controls` in here while we're at it

calm/interfaces/example/README.md
rocketstack-matt: Prototypes are intended to show how a new feature / change to the schema would operate, they're not intended for use / reference by real architectures.

The interfaces and controls folders are intended for things we want to publish as referencable.

calm/interfaces/example/container-image.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://calm.finos.org/interfaces/example/container-image",
+  "title": "Container Image Interface",
+  "type": "object",
```
rocketstack-matt, line 5: All the requirements should have this above the properties block

```json
"allOf": [
    {
      "$ref": "https://calm.finos.org/release/1.0-rc2/meta/interface.json#interface-definition"
    }
  ]
```

rocketstack-matt (changes requested):
Need to add the control-requirement reference in each control

calm/interfaces/example/container-image.json
rocketstack-matt: Ignore me - not how interfaces work!

.github/workflows/s3-sync.yml
markscott-ms: done

rocketstack-matt (approved):

PR #1468, opened 2025-07-29
Issue#1442 - Refactoring Architectures POST and PUT endpoints in CALM Hub

lbulanti-ms:
In this PR I updated the request body for a PUT and a POST request to accept two additional fields: name and description


calm-hub/pom.xml
```diff
@@ -245,6 +245,7 @@
                         <exclude>**/*CalmResourceErrorResponses.*</exclude>
                         <exclude>**/*Constants.*</exclude>
                         <exclude>**/*NamespaceStandardSummary.*</exclude>
+                        <exclude>**/*ArchitectureRequest.*</exclude>
```
grahampacker-ms, line 248: Why are we excluding this?

lbulanti-ms: We are excluding this from the coverage because the object does not contain any logic, it is used to only to hold data. The same approach was taken for other objects

grahampacker-ms (dismissed):

Issue #1476, opened 2025-07-31
CALM Schema Change: Disallow Undeclared Top-Level Fields in CALM Schema


LeighFinegold:
### Target Schema:

core.json

### Description of Change:

The CALM schema currently does not restrict additional properties at the top level. This proposal adds `"additionalProperties": false` to the root object in `core.json` to prevent accidental usage of undeclared top-level fields.

Given that the schema already includes a flexible `metadata` field for extensibility, this restriction encourages better conformance and clarity.

### Use Cases:

- Preventing schema users from introducing typos or undocumented properties at the top level.
- Promoting proper use of the `metadata` field for custom extensions.
- Enabling stricter validation in automated tooling and pipelines.

### Current Limitations:

- Any unknown or misspelled top-level fields are currently accepted silently.
- Users may assume undocumented properties are supported, leading to inconsistent usage and tooling issues.

### Proposed Schema Changes:

```json
{
  "additionalProperties": false
}
```

### Backward Compatibility:

- ❌ **Breaking change**: Existing documents with unknown top-level fields will fail validation.
- ✅ Valid CALM documents that adhere to the schema will remain valid.
- 🛠 A migration step may be needed for consumers to move any custom properties into `metadata`.

### Validation Strategy:

- Add test documents with and without extra top-level fields to verify enforcement.
- Validate with the JSON Schema test suite.
- Review all existing example documents to ensure compatibility.

### Implementation Impact:

- **Tools**: Validators may now report errors on previously-accepted documents.
- **Documentation**: Update schema examples and guidance to emphasize using `metadata` for extensions.
- **Schemas**: No other schema changes are required unless they assume open-ended top-level properties.

### Version Strategy:

- Mark as a **minor version** bump since the shape of valid documents is more constrained.
- Include in the v1.00 release without the need to create another RC candidate `1.0-rc3`
- Communicate clearly in release notes about the stricter top-level validation.

### Implementation Checklist:

- [x] Schema changes drafted and validated
- [x] Example documents created that validate against the new schema
- [x] Tests written to verify schema behavior
- [x] Documentation updated to reflect changes
- [x] Migration guide created (if applicable)
- [x] Related schemas updated (if applicable)

### Additional Context:

This change aligns with the CALM philosophy of clarity and contract-first design. Users are still able to extend architecture metadata using the flexible `metadata` field, which supports both objects and arrays.

markscott-ms:
[...]

Issue #1224, opened 2025-04-29
Add ADRs to core.json


jpgough-ms:
### Target Schema:

`core.json`

### Description of Change:

One of the key benefits of CALM is the ability to work alongside existing tools. We're not trying to replace/reinvent anything, more augment and centralise architectural content. We have a good compatibility story with C4, which continues to improve. What we don't have yet is ADRs. The introduction of ADRs in `core.json` will enable us to link to ADRs stored in either calm-hub, or existing documentation repositories. At the document level

### Use Cases:

- Documenting ADRs associated with both patterns and architectures
- Linking to existing ADRs
- Connecting to ADRs stored in both Calm Hub or other stores

### Current Limitations:

At the architecture level, so if an ADR applied to a given node or relationship that could not be attributed to it directly. The discussion point here was that the result of an ADR is a non-function/functional requirement that can be modelled as either a control interface where necessary

### Proposed Schema Changes:

```json
"adrs": {
  "type": "array",
  "items": {
    "type": "string"
  }
}
```

### Backward Compatibility:

Fully backwards compatible

### Validation Strategy:

Need to verify in validate and generate, but should be supported without much change

### Implementation Impact:

Testing required, but changes are not likely to be significant or needed

### Version Strategy:

Introduction to 1.0

### Implementation Checklist:

- [x] Schema changes drafted and validated
- [ ] Example documents created that validate against the new schema
- [ ] Tests written to verify schema behavior
- [ ] Documentation updated to reflect changes
- [ ] Migration guide created (if applicable)
- [ ] Related schemas updated (if applicable)

LeighFinegold:
This was discussed in https://github.com/finos/architecture-as-code/issues/1474 whilst @aidanm3341 was demoing calm-hub-uis.

The code base defines an [ADR](https://github.com/finos/architecture-as-code/blob/main/shared/src/model/adr.ts)
And [adr.spec.ts](https://github.com/finos/architecture-as-code/blob/main/shared/src/model/adr.spec.ts) indicate that this could be derived from JSON.

Was there a decision to not define a calm json schema based on the MADR template? If we are to hold the ADRs and render a format on CALM Hub UI why not give folks the option 

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://calm.finos.org/schema/adr/calm-adr.json",
  "title": "CALM Architecture Decision Record (ADR)",
[...]

PR #1478, opened 2025-08-04
New model that allows for native dererencing + canonical model for calm template --template/template-dir options

LeighFinegold:
## 🎯 Purpose

This PR introduces a foundational upgrade to the CALM modeling and templating pipeline. It adds support for canonical flattened outputs, model-native dereferencing, and a formal adapter interface for model/schema conversion. It also removes deprecated parser logic and aligns all CLI, docify, and test fixtures to support the new canonical model flow.

---

### 1️⃣ Native Dereferencing via `Resolvable<T>`

- Introduced `Resolvable<T>` wrapper for all reference fields (`configUrl`, `requirementUrl`, `definitionUrl`, etc.)
- Replaces `json-schema-ref-parser` with in-model reference resolution
- Provides `.isResolved`, `.value`, and `.dereference()` interfaces
- Tests added to validate resolved vs unresolved cases

### 2️⃣ Schema Transformation via `CalmAdaptable<T>`

- All model types now implement `CalmAdaptable<T>` with `.fromSchema()` and `.toSchema()`
- Enables symmetry between CALM JSON and internal model structure
- Supports clean validation, regeneration, and templating

### 3️⃣ Combined `ResolvableAndAdaptable<T>`

- Used specifically on `CalmNode.detailed-architecture` to allow referencing and adapting embedded CALM models
- Powers sub-architecture nesting and reuse

### 4️⃣ Canonical Flattening with `toCanonicalSchema()`

- Every model type now exposes `toCanonicalSchema()`
- Outputs flattened, dereferenced, and render-ready model data
- Fields like `requirement-url` and `config` are merged
- Greatly simplifies templates and downstream rendering

### 5️⃣ Default Transformer Outputs Canonical Model

- `default-transformer` now uses `toCanonicalSchema()` by default
- Template context is now uniform, predictable, and reference-free
- Templates updated to use flattened fields (e.g., direct `requirements`, `config`, etc.)

---

## 🧪 Supporting Changes and Cleanups

- Introduced model visitors (`dereference`, `logging`) to replace legacy deference and show the power of the new model.
- Expanded tests across all model layers for both transformed and raw states
- Updated CLI & template fixtures to use inline flow definitions and canonical models
- Flow Resolution Helper added to support enriched transition processing
- ADR models moved out of model package as it is yet to be part of the core schema - see discussion on [1224](https://github.com/finos/architecture-as-code/issues/1224#issuecomment-3141243236)
- Introduced a kind discriminator to relationshipType to replace brittle instanceof checks, which broke due to transpilation name mangling (e.g., _InteractsType instead of InteractsType).

---

## 📝 Additional Notes

- ⚠️ **Docify Templates Still Use Legacy Structure**  
  Docify templates still access `controls.data` and `details.detailed-architecture`. These will be migrated in a follow-up PR as we refactor the existing code to the concept of widgets. 

- 📦 **Schema Scope Constraint (v1 Lockdown)**  
  While support for `flow-url` may be reintroduced post v1, this PR assumes flows are defined **inline only** to match current schema constraints. Referenced flow files have been removed.

- 🔒 **Schema Lock: No Additional Properties**  
  Code assumes top-level CALM schema will remove `additionalProperties`, enforcing stricter shape constraints and cleaner model guarantees. See [1476](https://github.com/finos/architecture-as-code/issues/1476)


shared/src/template/types.ts
```diff
@@ -14,6 +16,6 @@ export interface TemplateEntry {
 }
 
 export interface CalmTemplateTransformer {
-    getTransformedModel(calmJson: string): any;
+    getTransformedModel(architecture: CalmCore): any;
```
LeighFinegold, line 19: We now start to lockdown calm template to only work on a calm model. This will be adapted to support other concepts like timeline in the future.

cli/src/cli.e2e.spec.ts
```diff
@@ -50,7 +50,7 @@ describe('CLI Integration Tests', () => {
 
     afterAll(() => {
         if (tempDir) {
-            fs.rmSync(tempDir, { recursive: true, force: true });
+            //fs.rmSync(tempDir, { recursive: true, force: true });
```
markscott-ms: Should this code be commented out?

markscott-ms (approved):

PR #1500, opened 2025-08-07
chore(deps): fixes #1499

### LINKED ISSUE #1499: Action Required: Fix Renovate Configuration
There is an error with this repository's Renovate configuration that needs to be fixed. As a precaution, Renovate will stop PRs until it is resolved.

Error type: Cannot find preset's package (helpers:pinGitHubActionsByDigest)

Thels:


.github/workflows/validate-renovate.yml
```diff
@@ -0,0 +1,40 @@
+name: Validate Renovate Configuration
+
+on:
+  push:
+    paths:
+      - 'renovate.json'
+      - '.github/renovate.json5'
```
rocketstack-matt, line 7: Is this a typo?

.github/workflows/validate-renovate.yml
```diff
@@ -0,0 +1,40 @@
...
+      - '.renovaterc*'
```
rocketstack-matt, line 8: What sort of files is this catch for?

.github/workflows/validate-renovate.yml
Thels: It's not - but it's removed for now.

I was reading through all of the possible names folks have called their renovate configurations. Back in 2019 they added support for `renovate.json5` - I added capability to handle it but it felt nonsensical as we're not going to shift the configuration without reason.

.github/workflows/validate-renovate.yml
Thels: Answered mostly in the other comment - but was thinking of supporting the different options available ( most listed at https://docs.renovatebot.com/configuration-options/ )

markscott-ms (approved):

PR #1506, opened 2025-08-07
Minor CLI Release with documentation updates

LeighFinegold:


LeighFinegold:
I was surprised I could merged this as that would impact cli test.

Remediated on https://github.com/finos/architecture-as-code/pull/1507

Issue #1556, opened 2025-08-31
Enhance DotNotation Support & Widget Integration


LeighFinegold:
## Summary

Improve the **DotNotation** handling and the **Template Preprocessor** so authors can write more natural Handlebars, reduce boilerplate, and reliably address nested models. This issue focuses on:

1. **Optional `this` for helpers** (e.g. allow `{{flow-sequence flow-id="x"}}` to behave like `{{flow-sequence this flow-id="x"}}`) as discussed 
on https://github.com/finos/architecture-as-code/issues/1553  
3. **Richer path handling** for **nested architectures** (support `nodes['node-name'].details` and deeper chains in helpers and standalone paths).  

This enhancement does **not** change widget behavior; it just enables cleaner templates for the end user.

## Goals / User Value

- Authors can write **shorter, clearer** templates:
  - `{{flow-sequence flow-id="x"}}` just works (implicit `this`).
  - Complex paths are **automatically** routed through `convertFromDotNotation`.
- **Predictable** handling of block openers, subexpressions, and reserved paths.
- **Better support for nested structures** with bracket and dot chaining.

## Proposed Implementation
TBD 

## Tasks

- [ ] Update `template-preprocessor.ts` to implement/confirm the rules above.
- [ ] Add/extend unit tests for all public methods.
- [ ] Add combined end-to-end fixture tests using realistic templates.
- [ ] Update docs / README: show new implicit-`this` behavior and nested DotNotation examples.

## Checklist

- [ ] Product agreement on implicit-`this` behavior.
- [ ] Engineering sign-off on preprocessor rules.
- [ ] Test coverage for nested & complex paths.
- [ ] Documentation updated.
- [ ] No breaking changes to widget inputs.

LeighFinegold:
While rendering with **calm-docify** doing below

```hbs
{{json-viewer nodes['load-balancer']}}
```

I ran into this issue with `calm docify`

```console
error [_TemplateProcessor]:    ❌ Error generating template: Parse error on line 9:
...es}}{{json-viewer nodes['load-balancer
---------------------^
Expecting 'CLOSE_RAW_BLOCK', 'CLOSE', 'CLOSE_UNESCAPED', 'OPEN_SEXPR', 'CLOSE_SEXPR', 'ID', 'OPEN_BLOCK_PARAMS', 'STRING', 'NUMBER', 'BOOLEAN', 'UNDEFINED', 'NULL', 'DATA', 'SEP', got 'INVALID'
```

[...]

Issue #1554, opened 2025-08-28
Create Flow Sequence Widget


LeighFinegold:
### Background
To render a flow sequence diagram, the widget must use both the flow document definition and the underlying architecture. Each flow is defined as a set of transitions, where each transition references a relationship by unique ID. To visualize the sequence, the widget must resolve these relationships to their corresponding nodes in the architecture, as node information is abstracted behind the relationship.

```json
  {
  "nodes": [
    {
      "unique-id": "conference-website",
      "name": "Conference Website",
      "description": "Website to sign up for a conference",
      "node-type": "webclient"
    },
    {
      "unique-id": "load-balancer",
      "name": "Load Balancer",
      "description": "Handles requests to the attendees service",
      "node-type": "network"
    },
    {
      "unique-id": "attendees",
      "name": "Attendees Service",
      "description": "Service for managing attendees",
      "node-type": "service"
    }
  ],
  "relationships": [
    {
      "unique-id": "conference-website-load-balancer",
      "description": "Request attendee details",
      "relationship-type": {
        "connects": {
          "source": { "node": "conference-website" },
          "destination": { "node": "load-balancer" }
        }
      }
    },
    {
      "unique-id": "load-balancer-attendees",
      "description": "Forward request to attendees service",
      "relationship-type": {
        "connects": {
          "source": { "node": "load-balancer" },
          "destination": { "node": "attendees" }
        }
      }
    }
  ],
  "flows": [
    {
      "unique-id": "flow-conference-signup",
      "name": "Conference Signup Flow",
      "description": "Flow for registering a user through the conference website",
      "transitions": [
        {
          "relationship-unique-id": "conference-website-load-balancer",
          "sequence-number": 1,
          "description": "User submits sign-up form via Conference Website to Load Balancer"
        },
        {
          "relationship-unique-id": "load-balancer-attendees",
          "sequence-number": 2,
          "description": "Load Balancer forwards request to Attendees Service"
        }
      ]
    }
  ]
}


```

```mermaid
sequenceDiagram
    Conference Website ->> Load Balancer: User submits sign-up form via Conference Website to Load Balancer
    Load Balancer ->> Attendees Service: Load Balancer forwards request to Attendees Service
```

### Requirement
Implement a `flow-sequence` widget that can render a sequence diagram for a given flow, resolving all node and relationship references as defined in the CALM model.

### Implementation Details

[...]

PR #1557, opened 2025-08-31
Adding flow-sequence widget

LeighFinegold:
## Description
Adds new widget to allow folks to render a flow sequence diagram 
```hbs
{{flow-sequence this flow-id="flow-conference-signup"}}
```
Additionally contains a small refactor to ensure widgets are dealing with a flattened options object and not a nested options object parsed to the widget renderer from handlebars

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Shared (`shared/`)
- [x] CALM Widgets (`calm-widgets/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


.github/workflows/build-calm-widgets.yml
```diff
@@ -32,7 +32,7 @@ jobs:
               run: npm run lint --workspace=calm-widgets
```
rocketstack-matt, line 32: If we're changing the method we specify the workspace on build and test do we want to change it for lint too?

LeighFinegold: calm-widget requires calm-models and therefore needs it built prior to building. Linting on the other hand is just static analysis of the source files. 

A similar approach has been taken on https://github.com/finos/architecture-as-code/blob/main/.github/workflows/build-shared.yml
https://github.com/finos/architecture-as-code/blob/main/.github/workflows/build-cli.yml#L35

LeighFinegold:
> One comment though - for the flow-sequence I have to specify `this` whereas for json-viewer I can specify `flows` - I think this should be consistent.

The list, table and json widget are very generic. They can take in any context and treat it as a json document.

The flow-sequence is a calm knowledgable widget it needs to leverage a calm document which has knowledge not just of the flows, but the relationships and nodes as well. Therefore the expectation is that the context you bind to is that of the calm document i.e. for this case wouldn't expect the need to specify flows. See https://github.com/finos/architecture-as-code/issues/1554#issuecomment-3236606286. 
At the top level and this example, `this` is CalmCore document, but if your in inner context with handlebars, then the user might want to use @root etc. 

https://github.com/finos/architecture-as-code/issues/1556 once implemented should tackle the need to reference "this" or imply that if no context path is defined, we are using the top level CALM document.  The DotNotation should reduce friction but we still probably have to provide some guidance on basic syntax permitted (which perhaps we can do when a VSCode (live calm docify) extension is provided.

PR #1574, opened 2025-09-04
fix(shared): validation of architectures against patterns with options (finos#1555)

### LINKED ISSUE #1555: Pattern Options Validation Bug
## Bug Report

Validation of architectures that were generated with Pattern Options do not work as expected when the Pattern Option involves multiple items selected from an `anyOf` block.

### Steps to Reproduce:
Here is a minimal pattern containing a pattern option (You can choose between no nodes, and two nodes from an `anyOf` block): [pattern.json](https://github.com/user-attachments/files/22048465/pattern.json)

Here is an architecture that was originally generated with the "two nodes" option, but then edited so that `node-2` is completely invalid (by removing all fields except `unique-id`: [architecture.json](https://github.com/user-attachments/files/22048541/architecture.json)

In this case, I'd expect errors regarding these missing fields. Instead I get this:
```json
{
    "jsonSchemaValidationOutputs": [],
    "spectralSchemaValidationOutputs": [
        // spectral errors omitted
    ],
    "hasErrors": false,
    "hasWarnings": true
}
```

No json schema errors.

Similarly if you take this architecture, which is the same above except for the `name` field is missing from `node-1`: [architecture.json](https://github.com/user-attachments/files/22048572/architecture.json)

Then I suddenly get a whole host of errors, some of which don't make any sense:

```json
{
    "jsonSchemaValidationOutputs": [
        {
            "code": "json-schema",
            "severity": "error",
            "message": "must have required property 'name'",
            "path": "/nodes/0",
            "schemaPath": "#/required"
        },
        {
            "code": "json-schema",
            "severity": "error",
            "message": "must have required property 'name'",
            "path": "/nodes/0",
            "schemaPath": "#/required"
        },
        {
            "code": "json-schema",
            "severity": "error",
            "message": "must be equal to constant",
            "path": "/nodes/0/unique-id",
            "schemaPath": "#/properties/nodes/prefixItems/0/anyOf/1/properties/unique-id/const"
        },
        {
            "code": "json-schema",
            "severity": "error",
            "message": "must match a schema in anyOf",
            "path": "/nodes/0",
            "schemaPath": "#/properties/nodes/prefixItems/0/anyOf"
        }
    ],
    "spectralSchemaValidationOutputs": [
        // spectral errors omitted
    ],
    "hasErrors": true,
    "hasWarnings": true
}
```

### Environment:
CALM CLI v1.0.0

### Additional Context:
@lbulanti-ms and I looked into this and believe that AJV simply cannot support this use case. It doesn't appear to have enough detail to properly validate the schema in the way that we'd like.

A potential solution to this problem involves the same approach taken to implement Pattern Options in the generate command. When Pattern Options are used, the generate command works like this:

- Extract which pattern options are available from the input pattern
- Prompt the user to select one of those options
- "Flatten" out the `oneOf` and `anyOf` blocks from the pattern, removing any nodes/relationships which were NOT chosen through the Pattern Option, and leaving behind those that were chosen
- Passing the modified pattern into the generation logic

Currently the validate command doesn't do any preprocessing of the pattern, it uses it exactly as it is when the command is called. The proposal for fixing this issue is to add in the same pre-processing step to flatten the pattern in the same way that generate does. It would look like this:

- Extract which pattern options were chosen from the input _architecture_ (this is available through the "options" relationship that's in the output architecture)
- Using the same logic that's used in the generate command, "Flatten" out the `oneOf` and `anyOf` blocks from the pattern, removing any nodes/relationships which were NOT chosen through the Pattern Option, and leaving behind those that were chosen
- Passing the modified pattern into the validation logic along with the architecture

With this approach, there are zero `oneOf` or `anyOf` blocks for AJV to deal with

markscott-ms:
## Description
Resolve #1555 by preprocessing a pattern against the chosen options documented in an architecture that was generated from it. This turns the pattern into one that can be used for validating.

The same option handling code used in `calm generate` is now used in `calm validate`, so little new code is added in this PR.

Tests have been added, drawn from the examples in #1555 

AI was utilised in a limited manner in production of this change.

## Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📚 Documentation update
- [X] ✅ Test additions or updates

## Affected Components

- [ ] CLI (`cli/`)
- [X] Shared (`shared/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [X] All existing tests pass

## Checklist
- [X] I have updated documentation if necessary

### DIFF
--- a/shared/src/commands/validate/validate.spec.ts
+++ b/shared/src/commands/validate/validate.spec.ts
@@ -1,4 +1,4 @@
-import { validate, sortSpectralIssueBySeverity, convertSpectralDiagnosticToValidationOutputs, convertJsonSchemaIssuesToValidationOutputs, stripRefs, exitBasedOffOfValidationOutcome } from './validate';
+import { validate, sortSpectralIssueBySeverity, convertSpectralDiagnosticToValidationOutputs, convertJsonSchemaIssuesToValidationOutputs, stripRefs, exitBasedOffOfValidationOutcome, extractChoicesFromArchitecture, applyArchitectureOptionsToPattern } from './validate';
 import { readFileSync } from 'fs';
 import path from 'path';
 import { ISpectralDiagnostic } from '@stoplight/spectral-core';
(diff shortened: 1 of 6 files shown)


shared/src/commands/validate/validate.ts
```diff
@@ -136,6 +133,12 @@ async function validateArchitectureAgainstPattern(architecture: object, pattern:
     let errors = spectralResult.errors;
     const warnings = spectralResult.warnings;
 
+    const patternResolved = applyArchitectureOptionsToPattern(architecture, pattern);
+
+    // Use JsonSchemaValidator
```
LeighFinegold: I know this statement was just moved, but can we delete this unnecessary comment

markscott-ms: sure

shared/src/commands/validate/validate.spec.ts
```diff
@@ -256,6 +256,85 @@ describe('validation support functions', () => {
         });
 
     });
+
+    describe('applyArchitectureOptionsToPattern', () => {
+        it('works with one options relationship', async () => {
```
LeighFinegold: This is actually an e2e test now due to file use. Perhaps we should move this test to there as well.

LeighFinegold (approved):

Issue #1762, opened 2025-10-28
CALM Schema change - architecture timelines


markscott-ms:
### Target Schema:

A new top level schema file, `calm-timeline.json` will be added.

### Description of Change:

This schema change comes from discussions related to #1409, primarily between @Thels, @markscott-ms, @rocketstack-matt, and @LeighFinegold 

The change introduces a timeline schema to enable CALM users to define and track an architecture OR pattern evolution through multiple iterations (_moments_).

Considerations made during this proposed change:

1. A minimal schema is being proposed. Adding additional properties in the future is far easier than removing or refactoring properties that are not required. For example, the following are not currently supported:
2. That a timeline _moment_ may refer to multiple architectures. If this is needed, it would be a trivial schema extension for a moment to support a list of _architectures_ rather than a single _architecture_.
3. Architecture mergers or spinoffs.
4. ADRs describe a reason for change, so naturally fit on a timeline, more so than on an architecture.
5. Consistency with the core CALM schema `calm.json` where it makes sense.
6. #1409 proposes a number of attributes which have not been implemented.
   1. `phase` - there was no strong justification to include. If deemed necessary it can be added in a minor revision.
   2. `sequence-order` - the moments are in an ordered list so this is unnecessary.
   3. `affected-nodes` etc - these should be derivable by comparing a pair of architectures.

### Use Cases:

- As an **Enterprise Architect**, I want to author timelines with a clear sequence of phases so that stakeholders can understand the planned evolution at a glance.
- As a **Developer**, I want to view future architecture phases so that I can write code that anticipates upcoming changes and avoid creating technical debt.
- As an **Auditor**, I want to link each phase to its ADRs so that I can trace the rationale behind every major architectural shift.
- As a **Project Manager**, I want to see planned completion dates and upcoming phases so that I can allocate resources appropriately and keep teams aligned
- As a **Artificial Intelligence Copilot**, I want to see how an architecture can be evolved between two points, to help model other evolutions.
- As a **Pattern Author**, I also want to capture how an architectural pattern has evolved, and the decisions involved in its change.

### Current Limitations:

* **No Native Timeline Support**. The current schema could be used to define a timeline, using `nodes` to represent each _moment_ in the timeline with a proprietary node type. Whilst this is a valid representation, it does not cleanly indicate that the document is a timeline. A timeline definition would solve this issue, making it easier for tooling to determine that a timeline rather than an architecture is being described.

### Proposed Schema Changes:

Please see pull request #1763 

[...]

PR #1763, opened 2025-10-28
feat(schema): CALM timelines schema + CLI support finos#1762

markscott-ms:
## Description
Proposed schema for CALM timelines as outlined in #1762 

In addition, the following code has been implemented:
- CALM CLI now supports `calm validate --timeline` to validate a timeline document against the schema and Spectral Rules
- AI Chatmode has Timelines support
- internal models have been updated with timelines.

## NOT done (further PRs invited / planned):
- CALM Widgets - a timeline rendering widget
- CALM Hub & UI
- Documentation
- VS Code Extension

I will create issues for these if this PR is accepted.

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [x] Shared (`shared/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-models/src/model/moment.ts
+++ b/calm-models/src/model/moment.ts
@@ -52,7 +52,7 @@ export class CalmMoment extends CalmNode implements CalmAdaptable<CalmMomentSche
             description,
             validFrom,
             CalmNodeDetails.fromSchema(details),
-            controls? CalmControls.fromSchema(controls): undefined,
+            controls ? CalmControls.fromSchema(controls) : undefined,
             metadata ? CalmMetadata.fromSchema(metadata) : undefined,
             adrs,
             Object.keys(additional).length > 0 ? additional : undefined


rocketstack-matt:
In response to the question from office hours, about how we could use nodes to represent the timeline . . . here are my thoughts:

- We declare a def of evolution, which is an array of nodes, each node has a detailed-architecture ref which points to a partially realised architecture, documenting purely the changes that would be happening at that evolutionary step.
- Moment becomes a def of node and an evolution. The node is the fully realised next step in the architecture timeline. The evolution is the specific stepwise changes that move the architecture from the previous version to the version in the moment.
- A timeline declares an array of moments

I think this reuse of node may make the integration into existing tools which already understand nodes and how to follow detailed-architectures simpler.

shared/src/commands/validate/json-schema-validator.ts
```diff
@@ -26,6 +27,7 @@ export class JsonSchemaValidator {
                 }
             }
         });
+        addFormats(this.ajv);
```
markscott-ms, line 30: This is to support the `"format": "date"` in the schema, to restrict the `valid-from` attribute to a YYYY-MM-DD date.

shared/src/commands/validate/validate.ts
```diff
@@ -86,28 +87,36 @@ async function runSpectralValidations(
...
+    timeline: object,
     schemaDirectory?: SchemaDirectory,
     debug: boolean = false
 ): Promise<ValidationOutcome> {
     logger = initLogger(debug, 'calm-validate');
 
     try {
-        if (architecture && pattern) {
+        if (timeline) {
+            if (architecture) {
```
LeighFinegold, line 106: What about --timeline and --pattern in conjunction with each other? 

I assume that does not make sense by how we defined the schema. Would one just use --timeline and use required-pattern instead of detailed-architecture?

cli/src/server/routes/validation-route.ts
```diff
@@ -57,7 +57,7 @@ export class ValidationRouter {
         }
         try {
 
-            const outcome = await validate(architecture, foundSchema, this.schemaDirectory, true);
+            const outcome = await validate(architecture, foundSchema, undefined, this.schemaDirectory, true);
```
LeighFinegold, line 60: Does the cli server still need to be maintained? Otherwise I suspect we would need to add support here.

shared/src/commands/validate/json-schema-validator.ts
LeighFinegold: It's interesting they decided to create an extension library for ajv core

shared/src/spectral/functions/timeline/ids-are-unique.ts
```diff
@@ -0,0 +1,21 @@
+import { JSONPath } from 'jsonpath-plus';
```
LeighFinegold, line 1: Do we have tests specifically for the spectral rules?

shared/src/commands/validate/validate.ts
```diff
@@ -86,28 +87,36 @@ async function runSpectralValidations(
...
+                logger.debug('You cannot provide an architecture when validating a timeline');
+                throw new Error('You cannot provide an architecture when validating a timeline');
+            }
+            return await validateTimelineAgainstPattern(timeline, pattern, schemaDirectory, debug);
```
LeighFinegold: This is confusing name because we don't actually check for the presence of pattern at all and in the example instructions you just give --timeline

calm-ai/tools/timeline-creation.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Architecture Timeline Creation Checklist
+
+- [ ] Include required $schema reference to CALM v1.1
+- [ ] Name file with `.timeline.json` suffix
+- [ ] Add moments array (even if empty initially)
+- [ ] Include metadata array for operational info
+**🚨 MANDATORY VALIDATION (Do not skip):**
+- [ ] **Verify calm-cli is installed:** `which calm`
+- [ ] **Run CALM validation:** `calm validate --timeline <filename>.timeline.json`
```
LeighFinegold, line 37: See https://github.com/finos/architecture-as-code/pull/1763/files#r2540006942 for confusing usage

shared/src/commands/validate/validate.ts
rocketstack-matt: Isn't a pattern only usable at a given moment rather than for a timeline overall?

shared/src/commands/validate/json-schema-validator.ts
markscott-ms: Yes, it was a surprise when the validation test failed and I had to look into why AJV was choking on something in the JSON Schema spec.

shared/src/commands/validate/validate.ts
markscott-ms: This is the shared library, so we are not talking command line options here.

The CLI validate command will REJECT usage of --timeline in conjuction with --pattern or --architecture. See `checkValidateOptions` in `cli/src/command-helpers/validate.ts`: https://github.com/finos/architecture-as-code/blob/3a244a666567a6c6e69a7ed6c52d5db495a25427/cli/src/command-helpers/validate.ts#L141

In the shared module validate.ts we are dealing with the call from L56 in the CLI https://github.com/finos/architecture-as-code/blob/3a244a666567a6c6e69a7ed6c52d5db495a25427/cli/src/command-helpers/validate.ts#L56   

At this point we have:
* a user provided architecture **with no** identifiable pattern or schema
* a user provided architecture **with** an identifiable pattern or schema
* a user provided architecture and pattern
* a user provided timeline **with no** identifiable schema
* a user provided timeline **with** identifiable schema.

The call to `validateTimelineAgainstPattern` on L110 is being passed the CALM timeline schema as the pattern.

I agree on the confusion. I shall rename this method validateTimelineOnly and make it clear that 'pattern' = 'calm-timelines schema' (technically it's whatever is in the $schema of the timeline JSON)

cli/src/server/routes/validation-route.ts
markscott-ms: An open question...

shared/src/spectral/functions/timeline/ids-are-unique.ts
markscott-ms: None of the spectral rules (for architectures, patterns or timelines) have specific tests. They only get exercised implicitly (and poorly). I will add some (for architectures, patterns and timelines).

cli/src/server/routes/validation-route.ts
markscott-ms: Please see https://github.com/finos/architecture-as-code/issues/1820 - I have discussed something like this with @jpgough-ms and want to see what others think.

shared/src/spectral/functions/timeline/ids-are-unique.ts
markscott-ms: I've added tests for timeline spectral rules.

I can add ones for architectures/patterns as part of a PR targeting main rather than this one.

rocketstack-matt:
This looks good to me

rocketstack-matt (approved):

PR #1913, opened 2025-12-04
feat(advent): add Advent of Calm zoom background image

rocketstack-matt:


markscott-ms:
For the record, O365 Copilot was partially involved in the creation of this background.

Issue #1826, opened 2025-11-20
CALM UI Hackday 25/11/25


aidanm3341:
# Planning for Hackday

We're having a hackday to brainstorm and work on a unified CALM UI.
The main two existing projects in this space are:

- [CALMHub UI](https://github.com/finos/architecture-as-code/tree/main/calm-hub-ui)
- [CALM Explorer](https://github.com/finos/architecture-as-code/pull/1766)

Proposed Plan:
- Identify features we like from both UIs
- Identify gaps in both UIs
- Create an ordered list of key features for the future
- Figure out what can be done in one day and can be done in parallel and get to work!

Let's use the comments to start posting some initial ideas.

rocketstack-matt:
We should also discuss the visualisations in calm-widgets and by extension the VSCode Extension to agree how (if) they all marry up. We certainly need improved visualization in the docify output.

rocketstack-matt:
Ideas from Hack Day kick off:
* We prefer CALM Explorer visual
  * Needs ability to connect to CALM Hub + show what is there
  * How easy would it be to package up the React components from CALM Explorer and make them all individually usable? These could then be used by CALM Hub UI + CALM Widgets

Proposal:
* Refactor elements we like from CLAM Explorer + CALM Hub UI and package as reusable components in calm-widgets
* Update CALM Hub UI to use the calm-widgets
* Docify + VS Code Extension already use calm-widgets

Plan:
* Extract CALM Hub UI Tree view into calm-widgets (@aidanm3341)
* Refactor docify website mode to allow complete control over the website structure using calm-widgets (@LeighFinegold)
* Extract visualiser from CALM Explorer and embed in CALM Hub UI (@aamanrebello)
* Introduce UI e2e testing to CALM Hub UI (user simulation) (@YoofiTT96)

Issue #1563, opened 2025-09-03
Introduce a two-stage website generation flow for CALM Docify that integrates with Docusaurus


LeighFinegold:
### Target Project:
`@finos/calm-docify` (CLI + shared template processing)

### Description of Feature:
Introduce a **two-stage website generation flow** for CALM Docify that integrates with **Docusaurus**.  
- **Stage 1 (Scaffold):** Generate a Docusaurus skeleton website with MDX files containing *unrendered* widget placeholders (e.g., `{{flow-sequence flow-id="..."}}`).  
- **Stage 2 (Deploy):** Fully process those placeholders using Docify templates and widgets, producing a ready-to-deploy static site.  

This enables teams to:  
- Quickly bootstrap a documentation website with CALM-aware scaffolding.  
- Edit, extend, or review placeholder docs in MDX before rendering.  
- Consistently deploy rendered static sites via `calm docify --deploy`.  

### User Stories:
- *As an* architect, *I want* to first scaffold a website or documentation with calm widget placeholders so that I can review, edit, or add explanatory text before running a final deploy that renders all widgets into diagrams and tables.  

### Current Limitations:
- Docify Website Mode currently renders templates directly, producing static docs immediately.  
- No way exists to generate a Docusaurus skeleton with unprocessed widget calls for teams who want editable scaffolding.  
- Editing or reviewing docs before final widget rendering is difficult.  

### Proposed Implementation:

- Add two new CLI commands:  
  - `calm docify scaffold` → scaffolds a Docusaurus project with MDX files containing **unrendered widget placeholders**.  
    - **Extra argument**: `--scaffold` (shorthand mode to trigger scaffold flow).  
  - `calm docify deploy` → processes placeholders into final rendered content and runs through the chosen renderer (default: Docusaurus).  
    - **Extra argument**: `--deploy`.  

- **Front matter** in scaffolded pages will include minimal CALM metadata so the VS Code extension and Docify know what they’re looking at:  

```yaml
---
id: architecture
title: Architecture Overview
calm:
  type: architecture
  modelRef: ./models/main.calm.json
  urlToLocalFileMapping: true
---
```

- Pages generated during `--scaffold` will contain placeholders such as:  

```mdx
# Architecture

{{w:table key="nodes"}}
{{w:flow-sequence flow-id="flow-add-account"}}
```

- Running `--deploy` replaces those placeholders with actual rendered diagrams/tables while keeping the surrounding text intact.

[...]

Issue #1533, opened 2025-08-21
Migrate Docify Website onto Calm Widget Framework


LeighFinegold:
### Background
The Calm Docify MVP v1.0.0 (#1452) successfully delivered basic template rendering and simple widgets (table, list, json). However, several advanced widgets that were previously available in the Docusaurus template bundle (C4 diagrams, flow sequences, related nodes) could not be migrated due to a circular dependency issue between `calm-widgets` and `shared` modules.

The current state:
- `shared` module contains CALM model types and depends on `calm-widgets`
- `calm-widgets` needs access to CALM types but cannot import from `shared` due to circular dependency
- Advanced widgets remain in Docusaurus-specific implementations
- Default Docify website output lacks sophisticated visualization capabilities

### Goal
Restructure the module dependencies and migrate all advanced widgets from Docusaurus-specific implementations to the general Calm Widget Framework, making them available in the default Docify website output.

### Success Criteria
- [x] Circular dependency between `shared` and `calm-widgets` resolved
- [x] All CALM model types accessible to widget implementations
- [x] Flow sequence widget migrated and functional in default Docify output
- [x] C4 diagram widget migrated and functional in default Docify output
- [ ] Related nodes widget migrated and functional in default Docify output
- [x] Default Docify website includes all advanced visualization capabilities
- [x] Feature parity maintained with previous Docusaurus implementations
- [ ] All existing functionality continues to work without regression

### Impact
Upon completion, users will have access to the full suite of architectural visualization tools in the default `calm docify` options

LeighFinegold:
Can close and will do so.

PR #2002, opened 2026-01-04
Scaffold Mode, Widget Framework Migration & Front-Matter Consolidation

LeighFinegold:
## Background

This PR delivers the production-ready implementation of the two-stage website generation flow that was prototyped during the [CALM UI Hackday](https://github.com/finos/architecture-as-code/issues/1826). 

The core objective is to enable architects to customize the default website output and ensure that what they see in VSCode's live docify preview is exactly what gets deployed. This creates a seamless experience from content creation in the IDE through to final deployment.

### 🎯 Scaffold Mode

A two-stage documentation workflow:

1. Stage 1 (`--scaffold`): Generate a Docusaurus skeleton with MDX files containing unrendered widget placeholders and front-matter variables
2. Stage 2 (`--template-dir`): Process the scaffolded templates to produce a deployable static site

This allows architects to review, edit, or add explanatory content to the generated documentation before final rendering.

### ♻️ Widget Framework Migration

Website mode now uses the unified **CALM Widget Framework**, replacing the legacy Handlebars partials. Templates use widget syntax like `{{table ...}}`, `{{flow-sequence ...}}`, `{{block-architecture ...}}`, and `{{related-nodes ...}}`.

### 📝 Front-Matter Variables

New front-matter system with variable support:
- Variables like `node-id`, `flow-id`, `relationship-id` are defined in `index.json` and injected into generated files
- `replaceVariables()` function substitutes `{{variable}}` placeholders with actual values
- Enables scaffolded files to work correctly in VSCode live preview

### 🔧 VSCode Plugin Consolidation

Removed duplicate `front-matter.ts` from the VSCode plugin - now uses `@finos/calm-shared` directly. This ensures consistent behavior between CLI and extension, and enables the extension to leverage front-matter variable substitution for scaffolded files.

### Two-Stage Workflow

```bash
# Stage 1: Generate scaffold with widget placeholders
calm docify --scaffold -a architecture.json -o ./website

# Edit generated MDX files in ./website/docs/

# Stage 2: Render final website from scaffolded templates  
calm docify -a architecture.json --template-dir ./website -o ./final-site
```

### Single-Stage Workflow (unchanged)

```bash
calm docify -a architecture.json -o ./website
```

## Breaking Changes

None for CLI users. Internal module restructuring only affects direct imports from template-bundles.

## Closes

- #1533 - Migrate Docify Website onto CALM Widget Framework
- #1563 - Introduce a two-stage website generation flow for CALM Docify

### DIFF
--- a/calm-plugins/vscode/package.json
+++ b/calm-plugins/vscode/package.json
@@ -2,7 +2,7 @@
     "name": "calm-vscode-plugin",
     "displayName": "CALM Preview & Tools",
     "description": "Live-visualize CALM architecture models, validate, and generate docs.",
-    "version": "0.2.1",
+    "version": "0.3.0",
     "publisher": "FINOS",
     "homepage": "https://calm.finos.org",
     "repository": {
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@finos/calm-cli",
-    "version": "1.26.1",
+    "version": "1.26.2",
     "description": "A set of tools for interacting with the Common Architecture Language Model (CALM)",
     "homepage": "https://calm.finos.org",
     "repository": {


LeighFinegold:
There are a few minor bugs that I saw with the table widget which will fix in a separate PR e.g. the rendered output when binding to an array of 1 interface vs 2 interfaces doesn't look quite right. The PR is quite large already (although most just small tweaks to generated output based on widgets)

One additional piece of functionality that could be enabled is an additional command in VS Code plugin to generate the website from an arch file. I have left it out of this PR. 

https://github.com/user-attachments/assets/d1b759b9-cca8-4664-8f23-64df4f75b684

Discussed with @markscott-ms  on how hopefully post this PR the support to change the theme via front matter can be controlled via one place. 

I'm still debating whether we should enable a --deploy as per the issue https://github.com/finos/architecture-as-code/issues/1563 just as its more friendly. Perhaps the scaffold should only generate the docs folder and not the website sidebar etc..

calm-widgets/src/widgets/table/table-vertical.html
```diff
@@ -13,7 +13,7 @@
 {{#with (lookup (lookup ../rows "0") "data") as |rowData|}}
 {{#with (lookup rowData colName) as |cellValue|}}
 <tr>
-    <th style="text-align:left">{{kebabToTitleCase colName}}</th>
+    <th>{{kebabToTitleCase colName}}</th>
```
LeighFinegold, line 16: Removed styles as there was already a css rule for docify to align-left th. Allows it to be more flexible in the future to end user.

markscott-ms (approved):

markscott-ms:
@LeighFinegold a note that the Live Docify in CALM Extension can't handle the rendering of a scaffolded node document.

PR #2012, opened 2026-01-06
fix(shared): protect against infinite recursion if a pattern references a JSON schema

markscott-ms:
## Description
@ConfuddledPenguin reported an issue during their Advent of CALM, where Copilot generated incorrect schema in a pattern, causing `calm validate` to fail after many seconds with an out of memory exception.

Here is a minimal pattern that would fail:

```
{
    "$schema": "https://json-schema.org/draft/2019-09/schema"
}
```

`calm validate -p <filename>`.

I traced this to an infinite loop in schema resolution by AJV. If the CALM schema, or JSON Schema Draft 2020-12 is referenced, all is well, but any other JSON Schema Draft leads to an infinite loop.

This PR causes validation to fail quickly and with a meaningful error if AJV seeks to load a JSON Schema:

```
error [json-schema-validator]:    Error fetching schema from schema directory: Error: Attempted to load standard JSON Schema with ID https://json-schema.org/draft-06/schema. This is not supported.
info [calm-validate]:     Formatting output as json
{
    "jsonSchemaValidationOutputs": [
        {
            "code": "json-schema",
            "severity": "error",
            "message": "Cannot read properties of undefined (reading '$schema')",
            "path": "/",
            "source": "pattern"
        }
    ],
    "spectralSchemaValidationOutputs": [],
    "hasErrors": true,
    "hasWarnings": false
}
```

## Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📚 Documentation update

## Affected Components

- [X] CLI (`cli/`)
- [X] Shared (`shared/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [X] All existing tests pass

## Checklist


shared/src/schema-directory.ts
```diff
@@ -136,6 +136,10 @@ export class SchemaDirectory {
     public async getSchema(schemaId: string): Promise<object> {
         if (!this.schemas.has(schemaId)) {
             try {
+                if (schemaId.startsWith('https://json-schema.org') || schemaId.startsWith('http://json-schema.org')) {
```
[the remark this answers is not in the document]
markscott-ms: `//...` style URLs are unsupported by CALM.

shared/src/schema-directory.spec.ts
```diff
@@ -133,18 +133,28 @@ describe('SchemaDirectory', () => {
...
         const definition = await schemaDir.getDefinition(ref);
 
         // this should include top-level, but also recursively pull in inner-prop
         expect(definition['properties']).toHaveProperty('top-level');
         expect(definition['properties']).toHaveProperty('inner-prop');
     });
+
+    it('reject attempts to load standard JSON Schemas - HTTPS protocol', async () => {
+        const schemaDir = new SchemaDirectory(mockDocLoader);
+        expect(() => schemaDir.getSchema('https://json-schema.org/draft/2019-09/schema')).rejects.toThrow();
```
rocketstack-matt: You're testing against a promise so should use await otherwise the test could pass even if the promise doesn't reject.

Also, wrapping in () => is unnecessary for promises.

```suggestion
        await expect(schemaDir.getSchema('https://json-schema.org/draft/2019-09/schema')).rejects.toThrow();
```

shared/src/schema-directory.spec.ts
```diff
@@ -133,18 +133,28 @@ describe('SchemaDirectory', () => {
...
+    });
+
+    it('reject attempts to load standard JSON Schemas - HTTP protocol', async () => {
+        const schemaDir = new SchemaDirectory(mockDocLoader);
+        expect(() => schemaDir.getSchema('http://json-schema.org/draft/2019-09/schema')).rejects.toThrow();
```
rocketstack-matt: You're testing against a promise so should use await otherwise the test could pass even if the promise doesn't reject.

Also, wrapping in () => is unnecessary for promises.

```suggestion
        await expect(schemaDir.getSchema('http://json-schema.org/draft/2019-09/schema')).rejects.toThrow();
```

shared/src/schema-directory.ts
aamanrebello: Could use regex matching here, something like `http(s?)://...`, but this is not a big deal.

shared/src/schema-directory.spec.ts
markscott-ms: Thanks, my TypeScript is clearly still in "learning" phase. I *thought* I copied what I'd seen elsewhere in the codebase, but the `() =>` is only used in other situations.

PR #2137, opened 2026-02-15
chore(calm-hub-ui): migrate from fetch to axios

### LINKED ISSUE #979: Migrate from fetch to HTTP library/axios
# Feature Request
## Description of Problem:
Mocking 'fetch' within Jest is hard and has introduced some awkward issues with libraries such as fetch-mock. 
@rocketstack-matt  has proposed a migration here #945 but we may be able to totally avoid this by just removing raw use of 'fetch' from the CLI.

# Potential Solutions:
Suggested solution is to use `axios` which is a lot easier to mock, since it isn't a global 'magic' function and instead creates an instance that can be mocked out.

selwyntheo:
## Summary

- Converts `calm-service.tsx` from standalone `fetch`-based functions to a `CalmService` class with injectable `AxiosInstance`, following the existing `AdrService` pattern
- Replaces the `fetch` call in `authService.tsx` (`checkAuthorityService`) with `axios.head()`
- Updates `TreeNavigation.tsx` and its tests to consume the new class-based API
- Adds comprehensive tests for `CalmService` (20 tests) and `checkAuthorityService` (2 tests)

Closes #979

## Test plan

- [x] All 196 existing + new tests pass (`npm test` in `calm-hub-ui/`)
- [x] Build compiles without errors (`npm run build` in `calm-hub-ui/`)
- [x] No remaining `fetch(` calls in `calm-hub-ui/src/service/` or `calm-hub-ui/src/authService.tsx`


markscott-ms:
@selwyntheo thanks for the contribution! We'll take a look over this soon as we have some other PRs pending or coming up for calm-hub-ui. FYI @aamanrebello

calm-hub-ui/src/hub/components/tree-navigation/TreeNavigation.tsx
aamanrebello, line 1: Hi Selwyn, sorry for taking so long to review this. Is it possible to fix the linting error in this file (around ReachHook having a missing dependency)? I'd rather we did not let this stuff creep in. If not possible, please explain why before I approve.

calm-hub-ui/src/authService.tsx
```diff
@@ -87,9 +88,9 @@ export async function getToken(): Promise<string> {
     return '';
 }
 
-export async function getAuthHeaders(): Promise<HeadersInit> {
+export async function getAuthHeaders(): Promise<Record<string, string>> {
```
aamanrebello: Can you explain why the type has been changed here? The new type is less restrictive, offering less type safety.

calm-hub-ui/src/service/calm-service.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+import axios from 'axios';
+
+const ax = axios.create();
+const mock = new AxiosMockAdapter(ax as never);
+
+const namespace = 'test-namespace';
+const resourceId = '1';
+const version = '1.0.0';
+
+const errorStatusCodes = [400, 401, 403, 404, 500];
```
aamanrebello: Why specifically these codes alone?

calm-hub-ui/src/authService.test.tsx
aamanrebello, line 1: Thank you for creating these tests, is there any reason why you have only tested `checkAuthorityService` and not the other functions?

calm-hub-ui/src/service/calm-service.tsx
```diff
@@ -1,251 +1,166 @@
...
+
+    constructor(axiosInstance?: AxiosInstance) {
+        if (axiosInstance) {
+            this.ax = axiosInstance;
+        } else {
+            this.ax = axios.create();
+        }
+    }
+
+    async fetchNamespaces(): Promise<string[]> {
```
aamanrebello: please put the public access modifier on these methods.

calm-hub-ui/src/service/calm-service.tsx
```diff
@@ -1,251 +1,166 @@
...
-        const values = Array.isArray(data?.values) ? data.values : [];
-        const namespaces = values
-            .map((v: { name?: string }) => v?.name)
-            .filter((name): name is string => typeof name === 'string');
-        setNamespaces(namespaces);
-    } catch (error) {
-        console.error('Error fetching namespaces:', error);
+        return this.ax
+            .get('/calm/namespaces', { headers })
+            .then((res) => res.data.values)
```
aamanrebello: Concerning that our current solution does not do this, but in the future we would like to add type safety around these calls.

calm-hub-ui/src/service/calm-service.tsx
aamanrebello, line 1: I love the way you have separated the "fetch" operation from operations like "setNamespaces"

calm-hub-ui/src/service/calm-service.tsx
aamanrebello: No need to implement in this PR though.

calm-hub-ui/src/hub/components/tree-navigation/TreeNavigation.tsx
selwyntheo: @aamanrebello  Thanks for reviewing this. I will add the lint error and push the changes requested here.

calm-hub-ui/src/service/calm-service.test.tsx
selwyntheo: These cover the most common HTTP error scenarios: But we can generalize

calm-hub-ui/src/authService.tsx
selwyntheo: Record<string, string> is more type-safe, not less is my view. It's a specific subset of HeadersInit that precisely describes what the function returns (an object with string keys/values), rather than a union of three possible types. This provides better type safety for consumers. But let me know your thoughts.

aamanrebello: If this is the case, then could you create a type for the auth header response and for now set it to be `Record<string, string` e.g. `export type AuthHeaders = Record<string, string>`

calm-hub-ui/src/service/calm-service.test.tsx
aamanrebello: If we want to cover different HTTP error scenarios, we should either cover all of them or cover just one and say that it can be extended to any other error code. Using only the "most common" scenarios feels a little arbitrary.

Could you possibly enumerate all the error codes. Or just use one error code.

calm-hub-ui/src/hub/components/tree-navigation/TreeNavigation.tsx
selwyntheo: @aamanrebello — Addressed in 330cb5d. The CalmService and AdrService instantiations are now wrapped in useMemo to stabilize the references, which resolves the useEffect missing dependency warning without introducing unnecessary re-renders. The lint rule passes cleanly.

calm-hub-ui/src/authService.tsx
selwyntheo: @aamanrebello — Good suggestion, done in f8126c2. I've introduced an AuthHeaders type alias:
typescriptexport type AuthHeaders = Record<string, string>;
getAuthHeaders now returns Promise<AuthHeaders>. This gives us a single point of change if the header shape evolves, while keeping the return type precise for current consumers.

calm-hub-ui/src/service/calm-service.test.tsx
selwyntheo: @aamanrebello — Agreed the selective list was arbitrary. In f8126c2, I've simplified this to use a single representative error code per test case, with a comment noting the behavior is status-code-agnostic since axios throws on any non-2xx response by default. This keeps the test suite focused and avoids the arbitrary selection issue you raised. Let me know if you'd prefer an exhaustive enumeration instead.

calm-hub-ui/src/authService.test.tsx
selwyntheo: @aamanrebello — Good callout. The initial test scope was limited to checkAuthorityService since it was the only function in authService.tsx that had its HTTP call migrated from fetch to axios in this PR. That said, broader coverage is valuable — in 330cb5d I added tests for getToken, getAuthHeaders, and isAuthServiceEnabled, bringing the auth service to full test coverage for the exported functions.

calm-hub-ui/src/authService.tsx
markscott-ms: looks good, thanks

PR #2378, opened 2026-04-24
feat(cli): Initial `calm workspace` commands

willosborne:
## Description
Adds `calm workspace` features. 
Purpose of this is to provide a local editing workflow similar to git branches - a workspace with files that will be resolved locally when validating, etc.
It allows pulling and pushing of files in these workspaces to/from CalmHub.

Commands (all under `calm workspace xyz`):
- `init` - makes a new workspace in root of current repository.
- `add` - tracks a file in the workspace. 
- `push` - upload tracked files to calmhub. will bump versions if they already exist, unless `--fail-if-modified` is set.
- `check` - verify that current workspace does not modify any files without having appropriately bumped their versions
- `bump` - look at every tracked file, and if they have changed, bump their versions. Then update all references in the tracked files to point at the new version.
- `switch` - change workspace
- `tree` - print contents of current workspace in a tree-like format
- `clean` - delete contents of current workspace or even entire workspace folder with `--all`

See `cli/README.md` for more details.

## Type of Change

- [x] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


willosborne:
@jimthompson5802 I've reworked this to use your calm hub client, rather than my custom service. Would appreciate a review when you get a mo. :)

willosborne:
Coverage is a smidge low. Working on it

cli/src/command-helpers/workspace/commands.ts
```diff
@@ -0,0 +1,385 @@
...
+        });
+
+    // Add a file to the current workspace bundle
+    workspaceCmd
+        .command('add')
+        .description('Add a file to the current CALM workspace bundle. By default, this creates a reference to the file at its current location.')
+        .argument('<file>', 'Path to the file to add to the bundle')
+        .option('--id <id>', 'Document ID to register for this file (defaults to filename without extension)')
+        .option('--copy', 'Copy the file into the bundle instead of referencing it from its current location.')
+        .option('--type <type>', 'Document type: architecture, pattern, schema, interface, timeline (defaults to unknown)')
```
jimthompson5802: The enumerated values here are a subset of the values for CALM_DOCUMENT_TYPE_LIST.  Would it be better to have this command hint use the same enumerated values that used later in the `enforceOptionPresenceByPrompt()` call?

cli/src/command-helpers/workspace/commands.ts
```diff
@@ -0,0 +1,385 @@
...
+
+async function enforceOptionPresenceByPrompt(cliInput: string | undefined, prompt: string, choices?: string[]): Promise<string> {
+    if (cliInput) {
+        // if it was selected already, just use that
+        return cliInput;
+    }
+    if (choices) {
+        return await select({
+            message: prompt,
+            choices: choices.map((p: string) => ({ name: p, value: p }))
```
jimthompson5802, line 466: If `--type` is not specified and the user is prompted to make a selection, when the cursor up/down key is pressed, a second list of enumerated values appear. See screenshot.

Image

Other than this UI anomaly the `--type` selection creates the expected workspace bundle specification.  If needed this is on MacOS.

Compare this UI behavior with the `calm init-ai` command and see how UI behaves with selection for AI provider.

cli/src/command-helpers/workspace/commands.ts
```diff
@@ -0,0 +1,385 @@
...
+                    }
+                    if (!id) {
+                        id = await input({ message: 'Enter a name for this document:' });
+                    }
+                }
+
+                const { id: resolvedId, destPath: finalDestPath } = await addFileToBundle(bundlePath, srcPath, {
+                    id,
+                    copy: options.copy,
+                    type: type as WorkspaceDocumentType | undefined,
```
jimthompson5802: If the user specifies an invalid value for `--type` that value is used in creating the bundle
```
calm workspace add calm/architecture/calm-1.json --namespace finos --type xyz --id my-id
```
created bundle
```
{
  "my-id": {
    "path": "/Users/jim/Desktop/finos/calm-workspace-commands-jmt/calm/architecture/calm-1.json",
    "type": "xyz",
    "namespace": "finos"
  }
}
```

The `calm workspace push` fails, which is expected behavior, so the integrity of data stored in CALM Hub is maintained. 
```
$ calm workspace push
(node:56103) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
error [workspace]:    Failed to create 'my-id': Hub error 400 on POST /calm/namespaces/finos/my-id: Invalid resource type: XYZ
```

From a UX perspective, would it be better to generate an error message on the `calm workspace add` command with the invalid `--type` to provide the user immediate feedback

cli/src/command-helpers/workspace/push.ts
```diff
@@ -0,0 +1,81 @@
...
+        if (remoteJson === null) {
+            try {
+                const calmHubId = await client.createResource(namespace, id, entry.type as CalmDocumentType, localJson);
+                manifest[id] = { ...entry, calmHubId };
+                await saveManifest(bundlePath, manifest);
+                logger.info(`Created '${id}' -> ${calmHubId}`);
+            } catch (e) {
+                logger.error(`Failed to create '${id}': ${e instanceof Error ? e.message : String(e)}`);
+            }
+        } else if (minifyJson(localJson) !== minifyJson(remoteJson)) {
```
jimthompson5802: as I understand this it is comparing the "stringified" representation of the JSON.  Does it matter if the order of keys change but the values are the same?  In this scenario, the string representation will be different even though the JSON objects are semantically equivalent.

shared/src/document-loader/document-loader.ts
```diff
@@ -29,6 +30,8 @@ export type DocumentLoaderOptions = {
     basePath?: string;
     allowedRemoteHosts?: string[];
     debug?: boolean;
+    // If set, DocumentLoader will attempt to load documents from a workspace bundle at this path
+    workspaceBundlePath?: string;
```
jimthompson5802, line 40: Is this for a future function?  I looked for references of `workspaceBundlePath` but could not find any in the `shared/src` or `cli/src` modules where this attribute is used.

rocketstack-matt:
@willosborne / @jimthompson5802 is this PR still in progress?

willosborne:
@rocketstack-matt i'm working on rewriting the underlying implementation to use the new mapped resources. i.e. calmhub changes have broken this. 
i want to get a minimal set of useful commands in so this isn't just hangnig aroudn any more.

willosborne:
### Note on shared commits with #2760

This branch now includes a merge of **#2760** (`feat(cli): add --fail-if-modified to hub push`), which has since been **merged to `main`** via merge commit `7d45299b`.

Because that was a merge (not a squash), those commits — the shared comparison helper `shared/src/hub/canonical.ts` and the `hub push --fail-if-modified` changes in `cli/src/command-helpers/hub-commands.ts` / `cli/src/cli.ts` — are now ancestors of `main` **and** of this branch. Once GitHub recomputes the merge base, they should drop out of this PR's diff, leaving only the `calm workspace` command suite as the net change here.

Why they were shared in the first place: `calm workspace push` reuses the same key-order-insensitive `canonicalEqual` helper as `hub push`. It was introduced once in `shared` (in #2760) so there's a single implementation, and added byte-identically on both branches so the merge produces no conflict or duplicate diff.

No action needed — flagging it so the overlap with the (now-merged) #2760 is clear during review.

shared/src/document-loader/document-loader.ts
willosborne: yeah i think this is not working correctly. i need to fix this.

cli/src/command-helpers/workspace/push.ts
willosborne: yeah i have reworked this. very good spot.

cli/README.md
```diff
@@ -526,4 +526,250 @@ To configure your CLI to use an auth plugin, use `~/.calm.json` in the same fash
 {
   "calmHubUrl": "http://calmhub.com",
   "authPluginPath": "~/plugins/auth-plugin.js"
-}
\ No newline at end of file
+}```
```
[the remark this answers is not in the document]
willosborne: fixed

cli/src/command-helpers/workspace/commands.ts
willosborne: i've updated this to use the same document document typres from docyument-loader as this should be more authoritative

cli/src/command-helpers/workspace/commands.ts
willosborne: interesting. ill look into this once this is merged as a separate bug

cli/src/command-helpers/workspace/commands.ts
willosborne: added same validations as 'new'

willosborne:
@rocketstack-matt @jimthompson5802 @LeighFinegold @aidanm3341 This PR is ready for review. I've cleaned it up a lot and aligned with our strategy around IDs.
This should now 
- let users bump versions and check things were bumped
- push with a `--fail-if-modified` flag
- interactively construct $ids with sensible defaults. e.g. prompt for each part of the URI and construct it for them.

rocketstack-matt (commented):
Nice feature, and the bundle/push/bump specs are thorough. Two points below — one functional, one for maintainability.

cli/src/command-helpers/workspace/bundle.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Build a dependency graph for the bundle where edges point from a document id to other document ids it references.
+ * Only references that resolve to ids present in the bundle manifest are included as edges.
+ */
+export async function buildDependencyGraph(bundlePath: string): Promise<DependencyGraph> {
+    const manifest = await loadManifest(bundlePath);
+    const idToPath: Record<string, string> = {};
+    for (const [id, entry] of Object.entries(manifest)) {
+        idToPath[id] = path.join(bundlePath, entry.path);
```
rocketstack-matt: `buildDependencyGraph` joins unconditionally, but `push.ts`, `bump.ts`, and `ref-rewrite.ts` all guard with `path.isAbsolute`. The default `add` (by reference) stores an **absolute** path, so `path.join('/bundle', '/abs/file.json')` yields a bogus path → `readFile` fails → no edges are built for that doc. `calm workspace tree` then shows wrong paths and no dependencies for the default workflow; only `--copy`'d files resolve. No test covers the referenced/absolute case.
```suggestion
        idToPath[id] = path.isAbsolute(entry.path) ? entry.path : path.join(bundlePath, entry.path);
```

cli/src/command-helpers/workspace/bump.ts
```diff
@@ -0,0 +1,146 @@
...
+    currentVersion: string;
+    latestHubVersion: string;
+}
+
+export interface BumpResult {
+    bumped: Array<{ id: string; filePath: string; fromVersion: string; toVersion: string }>;
+    refUpdates: RefUpdateResult[];
+}
+
+function resolveFilePath(bundlePath: string, entryPath: string): string {
```
rocketstack-matt: This `resolveFilePath` is duplicated in `ref-rewrite.ts` and inlined in `push.ts`; `stripVersionSuffix` is duplicated between `ref-rewrite.ts` and `shared/.../workspace-document-loader.ts`; and `resolveNewRef`/`WorkspaceDocumentLoader.resolveRule` are parallel implementations of the same matching rules. `bundle.ts` lacking the shared helper is what caused the bug above. Worth pulling these into one shared module so the path and ref-matching logic can't drift.

shared/src/hub/document-id-utils.ts
```diff
@@ -61,6 +61,37 @@ function parseDocumentId(documentId: string): DocumentIdMetadata {
     throw new Error(`Invalid document ID format: ${documentId}`);
 }
 
+/**
+ * Returns true if the given string is a conformant CalmHub document `$id` — either a
+ * namespace-resource id or a control requirement/configuration id.
+ */
+export function isConformantDocumentId(id: string): boolean {
```
jpgough-ms, line 68: **Blocking — data loss.** `ResourceType` is only `patterns/architectures/standards/interfaces`, so `parseDocumentId` throws for flow/adr/timeline `$id`s and `isConformantDocumentId` returns false. `workspace add` (`commands.ts:95`) then rewrites the file's `$id` and exits — silently mutating the user's file. All three are valid `--type` choices, and `timeline` was valid pre-PR, so this is also a regression.

cli/src/command-helpers/workspace/bundle.ts
jpgough-ms: **Blocking.** Default `add`/`new` store absolute reference paths, but this is the only manifest consumer missing the `path.isAbsolute` guard (push/bump/ref-rewrite all have it). `path.join(bundlePath, absolutePath)` produces a bad path, the read fails, and `workspace tree` shows every referenced doc as a childless root.

shared/src/document-loader/workspace-document-loader.ts
```diff
@@ -0,0 +1,198 @@
...
+            // Exact `$id` match
+            if (rule.docId && baseRef === rule.docId) return rule.localPath;
+
+            if (rule.basePath) {
+                // Same CalmHub path at a different version
+                const stripped = stripVersionSuffix(baseRef);
+                if (stripped !== null && stripped === rule.basePath) return rule.localPath;
+
+                // Full URL form — compare only the path portion
+                if (baseRef.startsWith('http://') || baseRef.startsWith('https://')) {
```
jpgough-ms, line 130: A host-less `$id` makes this full-URL branch match on path alone, so a `$ref` to the same path on a *different* host resolves to the local copy. Compare the origin too.

shared/src/document-loader/workspace-document-loader.ts
```diff
@@ -0,0 +1,198 @@
...
+        const fragmentIdx = reference.indexOf('#');
+        const baseRef = fragmentIdx >= 0 ? reference.slice(0, fragmentIdx) : reference;
+
+        for (const rule of this.rules) {
+            // Bare id match
+            if (baseRef === rule.bareId) return rule.localPath;
+            // Exact `$id` match
+            if (rule.docId && baseRef === rule.docId) return rule.localPath;
+
+            if (rule.basePath) {
```
jpgough-ms, line 122: A `$ref` without a `/versions/<v>` segment isn't matched here, so it falls through to CalmHub — breaking the documented "override every reference form" guarantee for unversioned (latest) references.

cli/src/command-helpers/workspace/ref-rewrite.ts
```diff
@@ -0,0 +1,214 @@
...
+
+        if (rule.basePath) {
+            // Path form with a different version
+            const stripped = stripVersionSuffix(baseRef);
+            if (stripped !== null && stripped === rule.basePath) {
+                return rule.targetPath + fragment;
+            }
+
+            // Full URL form — keep the same origin, replace only the path
+            if (baseRef.startsWith('http://') || baseRef.startsWith('https://')) {
```
jpgough-ms, line 68: Dead branch in the common case: when `$id` is a full URL, this compares a host-less `pathname` against a host-qualified `basePath` and never matches, so `bump` leaves stale cross-host version pointers in place. (Mirror of the loader host-matching issue.)

cli/src/command-helpers/workspace/commands.ts
```diff
@@ -0,0 +1,476 @@
...
+                const documentId = await promptForDocumentId({ baseUrlDefault });
+
+                name = await enforceOptionPresenceByPrompt(name, `Enter the title for your new ${type} document:`);
+                if (!template) {
+                    template = templates.length > 1
+                        ? await enforceOptionPresenceByPrompt(undefined, 'Select a template:', templates)
+                        : (templates[0] ?? 'empty');
+                }
+
+                const bundlePath = findWorkspaceManifestPath(process.cwd());
```
jpgough-ms: `workspace new` prompts the full interactive flow before this bundle check, so a user answers every prompt and then loses it all to `exit(1)`. Move the check ahead of the prompts.

cli/src/command-helpers/workspace/new.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param type The CALM document type (e.g. "architecture", "pattern")
+ * @param slug A short, filename-safe slug for the document (e.g. the mapping id)
+ * @param templateName The template to use (e.g. "empty", "with-structure"); defaults to "empty"
+ * @returns The absolute path to the newly created file
+ */
+export async function createNewDocument(documentId: string, name: string, type: string, slug: string, templateName = 'empty'): Promise<string> {
+    const templatePath = path.join(TEMPLATES_DIR, type, `${templateName}.hbs`);
+    const source = await readFile(templatePath, 'utf8');
+    const content = Handlebars.compile(source)({ id: documentId, name });
+    const filename = `${slug}.${type}.json`;
```
jpgough-ms, line 43: No overwrite guard and weak slug validation: re-running `new` with the same slug silently overwrites the prior file, and on Windows a `.\` slug escapes cwd. Add an `existsSync` check and validate the slug against `\` and `..`.

jpgough-ms (changes requested):
**Needs changes — correctness.** Two blockers (one is data loss) plus five fixes worth resolving before merge. Details inline. Nice first cut overall: every new file has a spec and the loader integrates cleanly into the strategy chain.

cli/src/command-helpers/workspace/bump.ts
willosborne: fixed

cli/src/command-helpers/workspace/bundle.ts
willosborne: fixed

willosborne: fixed already - matt also caught this.

jpgough-ms (dismissed):

shared/src/hub/document-id-utils.ts
willosborne: fixed, it add now warns about a bad id but doesn't wipe it

cli/src/command-helpers/workspace/commands.ts
willosborne: done

cli/src/command-helpers/workspace/new.ts
willosborne: done

shared/src/document-loader/workspace-document-loader.ts
willosborne: fixed

cli/src/command-helpers/workspace/ref-rewrite.ts
willosborne: origin and path are now checked separately

shared/src/document-loader/document-loader.ts
```diff
@@ -5,9 +5,15 @@ import { FileSystemDocumentLoader } from './file-system-document-loader';
 import { DirectUrlDocumentLoader } from './direct-url-document-loader';
 import { MultiStrategyDocumentLoader } from './multi-strategy-document-loader';
 import { MappedDocumentLoader } from './mapped-document-loader';
+import { WorkspaceDocumentLoader } from './workspace-document-loader';
 import { AuthPlugin } from '..';
 
-export type CalmDocumentType = 'architecture' | 'pattern' | 'schema' | 'timeline';
+export type CalmDocumentType = 'architecture' | 'pattern' | 'schema' | 'timeline' | 'interface' | 'flow' | 'adr' | 'control';
+export const CALM_DOCUMENT_TYPES_LIST: CalmDocumentType[] = ['pattern', 'architecture', 'interface', 'flow', 'control', 'schema', 'timeline', 'adr'];
```
LeighFinegold, line 12: If we are finally introducing this concept, this should be official in calm-models in my opinion.

LeighFinegold, line 12: Potential bug? 

🐞 CALM_DOCUMENT_TYPES_LIST includes `'adr'` and this list is used as the .choices() for `calm workspace new` / `add`. But there is no `adr/` template directory?

cli/package.json
```diff
@@ -15,12 +15,13 @@
         "calm": "dist/index.js"
     },
     "scripts": {
-        "build": "tsup && npm run copy-calm-schema && npm run copy-docify-templates && npm run copy-widgets && npm run copy-ai-tools",
+        "build": "tsup && npm run copy-calm-schema && npm run copy-docify-templates && npm run copy-widgets && npm run copy-ai-tools && npm run copy-workspace-templates",
```
LeighFinegold, line 18: As soon as I saw the templates, I was looking out for this - so its good it is there otherwise those cli commands will fail at runtime. 

It be good maybe to have some e2e tests covering this apsect?

jpgough-ms (approved):

shared/src/document-loader/document-loader.ts
willosborne: might just cut adrs for now as no one is using this

willosborne: agreed this needs to be standard in one place

cli/package.json
willosborne: absolutely. i want to have a set of e2e/smoke tests for the whole CLI/calm hub integration. docker container started then a set of command invokations with inspections against the API at each step. 

IMO this is out of scope for this task and should be separate, but i will definitely do this.
@jpgough-ms is on board too.

shared/src/document-loader/document-loader.ts
willosborne: Will fix as #2770

cli/package.json
willosborne: Will do in #2771

PR #2426, opened 2026-05-06
Missing MCP Tools: Update Architecture

jpgough-ms:
## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [ ] All existing tests pass

## Checklist


jpgough-ms:
/easycla

Does nothing for me @bingenito

rocketstack-matt (dismissed):
### 1. `name` and `description` are silently nulled on every update (blocker)

`Architecture.ArchitectureBuilder` defaults `name`/`description` to `null` (`Architecture.java:111-112`), and the new tool builds the object without setting them (`ArchitectureTools.java:148-153`). Both stores then unconditionally `$set` whatever is on the object:

- `MongoArchitectureStore.writeArchitectureToMongo` lines 230-233
- `NitriteArchitectureStore.writeArchitectureToNitrite` lines 287-288

User-visible impact: `getArchitecturesForNamespace` falls back to `\"Architecture \" + archId` when name is null (`MongoArchitectureStore.java:88-91`), so an architecture's display name silently degrades from e.g. `MCP Test Arch` to `Architecture 7` after the first `updateArchitecture` call. Description goes to empty string.

The integration tests don't catch this because the only `listArchitectures` assertion runs at `@Order(7)` — *before* the update at `@Order(26)`. Suggested fixes:

- Add optional `name`/`description` `@ToolArg`s (matches the REST `PUT` shape in `ArchitectureResource.updateVersionedArchitecture`), or
- Look up the existing values and re-apply them when the caller doesn't pass them.

A regression test would help — either a post-update `listArchitectures` assertion in the integration tests, or an `ArgumentCaptor<Architecture>` in the unit test asserting `arch.getName()` is preserved.

### 2. Tool description says \"publish a new version\" but the implementation is an upsert

`ArchitectureStore` exposes both `createArchitectureForVersion` (throws `ArchitectureVersionExistsException` — true insert-only) and `updateArchitectureForVersion` (silent upsert — overwrites if the version already exists). The PR wires up the upsert, but the `@Tool` description (`ArchitectureTools.java:132`) reads as insert-only:

> Publish a new version of an existing architecture. Use this to add a new semantic version (e.g. '1.1.0') against an existing architecture ID without allocating a new identity.

For an LLM-driven tool this is risky: passing the wrong version string silently clobbers a published one with no error. Two reasonable resolutions:

- If the intent is genuinely \"publish a new version\", switch to `createArchitectureForVersion` and surface `ArchitectureVersionExistsException` as a tool error. (No other MCP tool exposes this method yet, so there's no overlap concern.)
- If overwrite is intended, update the description to say so explicitly so the LLM/operator can reason about it.

### 3. `allow.put.operations` config gate is bypassed

The REST `PUT` equivalent (`ArchitectureResource.updateVersionedArchitecture`, lines 218-248) is gated on `allow.put.operations`, which **defaults to `false`** (`ArchitectureResource.java:49`). Default deployments therefore have REST PUT disabled. With `calm.mcp.enabled` defaulting to `true`, this new tool exposes a mutation path the REST surface explicitly forbids by default — effectively flipping the default mutation posture for any deployment running MCP.

`PatternResource` and `FlowResource` apply the same flag, so the convention is intentional. Suggest either honouring the flag in the MCP tool or documenting clearly that `calm.mcp.enabled` is the only gate (and that enabling MCP overrides `allow.put.operations`).

### 4. `architectureJson` payload size is not bounded

`McpValidationHelper.MAX_JSON_PAYLOAD_LENGTH = 100_000` and `validateMaxLength` already exist (`McpValidationHelper.java:32, 167`) for exactly this case but aren't applied here. Pre-existing — `createArchitecture` has the same gap — so not a regression in this PR, but worth closing the unbounded-blob vector while you're in this file.

willosborne (approved):

Issue #1116, opened 2025-04-02
Expose CALM as a Library for Extension and Template Use


LeighFinegold:
## Context  
As part of the ongoing effort to make CALM more modular and developer-friendly, we want to expose CALM models and utilities in a reusable way so that consumers (e.g., template authors) can import and extend the architecture model and related tooling.

This issue aims to clarify:
- What parts of CALM should be exposed
- How we should structure and publish those parts

---

## ❓ What to Expose

We currently have two main directions:

1. **Just the Model**  
   - Expose `CalmArchitecture`, `CalmNode`, `CalmRelationship`, etc.
   - Enough to allow template bundles to build visitors or augment the model

2. **Full Utility Set (Shared)**  
   - Expose everything from `@finos/calm-shared`, including schema resolution, parsing, loading, error handling, etc.
   - This would empower more advanced extensions, such as validators or custom CLI plugins

---

## 📦 How to Expose

We have two main packaging options:

### Option 1: Expose via `@calm-cli`
Users would import like this:

```ts
import { CalmArchitecture } from '@calm-cli/model/CalmArchitecture'
```

**Pros:**
- Keeps everything in one package
- Easy for template developers already using the CLI

**Cons:**
- CLI-focused name might confuse library users
- Less modular

---

### Option 2: Separate Package like `@calm-model` or `@calm/core`

Users would import like this:

```ts
import { CalmArchitecture } from '@calm-model'
```

**Pros:**
- Clean separation of concerns
- Easier for external consumers to depend on CALM without pulling in CLI or non-core features
- Better long-term for maintaining semantic versioning

**Cons:**
- More packages to manage
- Slight overhead for releases

---

## ✅ Proposal (Open for Discussion)

- **Short-Term**: Expose core models via `@calm-cli/model/*` to unblock template development
- **Long-Term**: Consider creating a dedicated package like `@calm-model` (or `@finos/calm-model`) that houses all shared definitions and utilities

---

## 👀 Questions

- Do template authors need more than the model and calm doc parser (e.g., validator, schema resolution, graphing,c4 tooling)?
- Would a dedicated `@calm-model` package reduce friction for new users?

Thels:
Interface that needs to be included is `CalmTemplateTransformer.ts` - as that will be required as part of the custom transformers to adhere to the interface. 

Another option to consider - creating a `model` module - which `shared` depends on - which is published with the same version as the CLI.

LeighFinegold:
[...]

PR #2452, opened 2026-05-13
Add auth plugin to allow configurable auth strategy to CLI - #2453

willosborne:
## Description
( #2453 ) Auth to CalmHub will vary based on org/environment. As such CLI will offload to external TS auth plugin.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/cli/src/cli-config.spec.ts
+++ b/cli/src/cli-config.spec.ts
@@ -1,5 +1,8 @@
 import { fs, vol } from 'memfs';
-import { loadCliConfig } from './cli-config';
+import { loadCliConfig, loadAuthPlugin } from './cli-config';
+import { resolve } from 'path';
+import { homedir } from 'os';
+
 
 vi.mock('fs/promises', async () => {
     const memfs: { fs: typeof fs } = await vi.importActual('memfs');
@@ -7,8 +10,13 @@ vi.mock('fs/promises', async () => {
     return memfs.fs.promises;
 });
 
+vi.mock('fs', async () => {
+    const memfs: { fs: typeof fs } = await vi.importActual('memfs');
+    return memfs.fs;
+});
+
 vi.mock('os', () => ({
-    homedir: () => '/home/user'
+    homedir: vi.fn(() => '/home/user')
 }));
 
 const exampleConfig = {
@@ -17,6 +25,9 @@ const exampleConfig = {
     authPluginPath: './auth-plugin.js'
 };
 
+const FIXTURES_DIR = resolve(__dirname, '../test_fixtures');
+const JS_FIXTURE = resolve(FIXTURES_DIR, 'test-auth-plugin.js');
+const TS_FIXTURE = resolve(FIXTURES_DIR, 'test-auth-plugin.ts');
 
 describe('cli-config', () => {
     beforeEach(() => {
@@ -25,6 +36,7 @@ describe('cli-config', () => {
 
     afterEach(() => {
         vol.reset();
+        vi.mocked(homedir).mockReturnValue('/home/user');
     });
 
     it('loads user config from .calm.json in home dir', async () => {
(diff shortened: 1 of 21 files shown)


yt-ms:
Does the plugin mechanism have to be specific to auth, or could auth be an example of something more generic?  We talked about needing a plugin mechanism last year I think.

willosborne:
Hmm, I guess we could have a more general plugin that is invoked with different lifecycle hooks for different operations. But off the top of my head I can't think of any other things we'd want to use this for.

@yt-ms I've put this in for discussion in this afternoon's Office Hours.

EDIT: whoops! Closed my own PR by mistake. Wrong button.

cli/src/cli-config.spec.ts
```diff
@@ -13,7 +13,8 @@ vi.mock('os', () => ({
 
 const exampleConfig = {
     calmHubUrl: 'https://example.com/calmhub',
-    allowedRemoteHosts: ['schemas.example.com']
+    allowedRemoteHosts: ['schemas.example.com'],
+    authPluginPath: './auth-plugin.js'
```
markscott-ms, line 25: What directory is the auth plugin relative to? Should we give an example of ~/.calm/auth-plugin.js instead to indicate a non relative path is best?

cli/test_fixtures/getting-started/STEP-2/website/package.json
```diff
@@ -15,14 +15,14 @@
...
-    "@mermaid-js/layout-elk": "^0.2.0",
+    "@mermaid-js/layout-elk": "^0.1.9",
     "docusaurus-plugin-search-local": "^2.1.2",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "unist-util-visit": "^5.0.0"
   },
   "overrides": {
-    "@mermaid-js/layout-elk": "^0.2.0",
+    "@mermaid-js/layout-elk": "^0.1.9",
```
markscott-ms: Why is there changes to layout-elk ? Unrelated to auth plugin

shared/src/docify/template-bundles/docusaurus/package.spec.ts
```diff
@@ -24,7 +24,10 @@ describe('docusaurus template bundle dependencies', () => {
         const exitCode = await new Promise<number>((resolve, reject) => {
             const stderr: string[] = [];
             const proc = spawn('npm', ['install', '--package-lock-only', '--prefix', tmpDir]);
-            proc.stderr?.on('data', (chunk: Buffer) => stderr.push(chunk.toString()));
+            proc.stderr?.on('data', (chunk: Buffer) => {
+                stderr.push(chunk.toString());
+                console.error(chunk.toString());
```
markscott-ms, line 29: Was this added for additional debugging and left by accident?

cli/test_fixtures/getting-started/STEP-2/website/package.json
willosborne: i cant run an npm install without this change, even if i delete node modules. is that really just me? another example of npm just not working on my laptop for some reason maybe... thats soooo annoying

shared/src/docify/template-bundles/docusaurus/package.spec.ts
willosborne: yeah. Its impossible to diagnose any issues with this test (which fails if there are any dependency tree problems) without as all output is suppressed. i thought about deleting it but decided to leave it in case anyone else runs into this issue

cli/src/cli-config.spec.ts
willosborne: Good idea. I'll definitely add that, i think this is how most people will install (or an absolute path, i guess)

cli/src/cli-config.ts
```diff
@@ -38,3 +42,54 @@ export async function loadCliConfig(): Promise<CLIConfig | null> {
...
+
+    filename = resolveHomeDir(filename);
+    
+    if (!existsSync(filename)) {
+        logger.error(`❌ Auth plugin file not found: ${filename}`);
+        throw new Error(`❌ Auth plugin file not found: ${filename}`);
+    }
+    if (filename.endsWith('.ts')) {
+        logger.info(`🔍 Loading auth plugin as TypeScript: ${filename}`);
+        register({
```
LeighFinegold: I’m a bit uneasy about copying the docify transformer loading pattern here. In shared/src/template/template-processor.ts, custom transformers can already be provided as .ts or .js, and for .ts we call ts-node.register(...) in-process before dynamically importing the transformer. That means user-provided extension code is compiled against the current host runtime / loader behavior rather than as a prebuilt artifact, which can create module-resolution and runtime compatibility issues.

I don’t think that’s a great precedent to extend to auth plugins as well. From an end-user perspective, a plugin should ideally be a stable artifact we load, not something we transpile inside the CLI process.

Also, if we do want people to implement these extension points, I think we should make sure we expose the relevant definitions/types as part of the public library surface, rather than asking users to rely on implicit runtime conventions. That lines up with the earlier discussion in #1116 about exposing CALM models/shared utilities to support extension and template authors.

So my preference would be:
- keep the runtime contract narrow (.js / precompiled modules), and
- separately make sure the plugin interface/types are properly exported for consumers.

willosborne: right - that's why i export the authplugin from calm-shared. idea being that anything you import should be built against that and this is already published to npm.

i'd be OK with that to be honest - i guess though, we have no way of _knowing_ they're writing TS and compiling against the definitions. there's nothing to stop them just writing plain JS.

If this is what you'd prefer I'm happy to change that.

Do you think I should add a new top-level project with an example package.json that declares a dependency on @finos/calm-shared as a model? im not sure how to do this without it just auto-importing from the workspace. maybe I can do that as a followup

cli/src/cli-config.spec.ts
willosborne: added tests that use ~ and absolute paths and also a readme showing examples.

cli/src/cli-config.ts
willosborne: also follow up - why is this approach used for doc transformer if it causes module issues?

cli/test_fixtures/getting-started/STEP-2/website/package.json
willosborne: ok, strangely, by nuking everything it fixes it. i think deleting root level node modules wasnt enough, also needed to do the project specific oens. removed the overrides.

cli/src/cli-config.ts
willosborne: I have removed the TS precompiling.

LeighFinegold: > also follow up - why is this approach used for doc transformer if it causes module issues?

I think it should be - but I guess do need to check if anyone is providing it currently. This is a new plugin so let's not even let it be an option. 

Also the comment was to bring to the meeting office hours the need to consider exposing typescript defs as well from build

LeighFinegold (approved):
As discussed on the office hours meeting. We can approve this plugin and evolve it if there are other plugin points e.g. if we want to inject observability into cli

cli/src/cli-config.ts
willosborne: Ah yeah OK makes sense. Thanks, agreed we should probably remove this from doc transformer and publish that dependency too.

willosborne:
FYI @yt-ms we discussed your suggestion on office hours. Conclusion - we _might_ need this, e.g. for custom validation rules, but right now there's no demand. Importantly though we agreed we would want any such extension logic to be loaded in a separate way to the auth plugin, which is a install-level concern that you'll want on every request for every use case. So this change isn't affected by anything we do there.

PR #2464, opened 2026-05-18
feat(calm-hub): add new profile to run calmhub behind sidecar proxy

willosborne:
## Description
Add new profile to run calmhub behind sidecar proxy, with user identity passed via a header.
Defaults to Proxy-Remote-User but is configurable.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


rocketstack-matt (commented):
Solid, focused change overall — the `proxy` profile cleanly coexists with `secure`, the README is explicit about the spoofing risk, and the new filter has good unit-test coverage of its meaningful branches. A few things to address before merge — inline comments below. Highlights:

- The `SearchResource` proxy-header fallback (the whole reason the constructor signature changed) has no test coverage.
- Removing `@IfBuildProfile("secure")` from `UserAccessValidator` silently changes behaviour in the default (no-profile) dev mode — worth re-gating with `anyOf = {"secure", "proxy"}` unless that's intentional.
- PR description says the default header is `Proxy-Remote-User`, but the code default (and README) is `Remote-User` — minor, but worth correcting the description.
- A handful of typos / a stray pair of backticks in the README code fence.

Nothing here is blocking by itself; collectively I'd like the test gap and the profile-gating decision sorted before approval.

calm-hub/src/main/java/org/finos/calm/security/UserAccessValidator.java
```diff
@@ -17,10 +16,9 @@
  * Validates whether a user is authorized to access a particular resource based on
  * their assigned permissions and namespaces.
  * <p>
- * This validator is active only when the 'secure' profile is enabled.
+ * This validator has no effext unless the 'secure' or 'proxy' profile is enabled.
  */
 @ApplicationScoped
```
rocketstack-matt, line 21: Removing `@IfBuildProfile("secure")` makes this bean resolvable in the **default** (no-profile) dev mode too, not just under `secure` or `proxy`. That has a subtle knock-on effect in `SearchResource.resolveReadableNamespaces()`:

- Before: `userAccessValidatorInstance.isResolvable()` was `false` in the default profile → method always returned `Optional.empty()` (no namespace filtering).
- After: it's now `true`. JWT instance is unresolvable, so it falls through to `httpHeaders.getHeaderString("Remote-User")`. Any dev-mode caller sending that header will silently get their search results filtered by access grants.

Not a production security regression (default profile is dev-only), but it is a silent behaviour change. Suggest re-gating to keep the surface tight:

```java
@IfBuildProfile(anyOf = {"secure", "proxy"})
```

(or document the new default-profile behaviour explicitly.)

calm-hub/src/main/java/org/finos/calm/security/UserAccessValidator.java
```diff
@@ -17,10 +16,9 @@
  * Validates whether a user is authorized to access a particular resource based on
  * their assigned permissions and namespaces.
  * <p>
- * This validator is active only when the 'secure' profile is enabled.
+ * This validator has no effext unless the 'secure' or 'proxy' profile is enabled.
```
rocketstack-matt: Typo: `effext` → `effect`.

calm-hub/src/main/java/org/finos/calm/security/ProxyAccessControlFilter.java
```diff
@@ -0,0 +1,67 @@
...
+        this.usernameHeader = usernameHeader;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
+        PermittedScopes annotation = resourceInfo.getResourceMethod().getAnnotation(PermittedScopes.class);
+        if (Objects.isNull(annotation)) {
+            logger.warn("Unsecured endpoint accessed: {}", resourceInfo.getResourceMethod());
+            return;
+        }
```
rocketstack-matt, line 42: `AccessControlFilter` checks `requiredScopes` from the JWT before invoking the user-grant validator (`hasRequiredScope`). The proxy filter skips that entirely — which is arguably correct (there's no JWT to read scopes from), but it means the `PermittedScopes` annotation degrades to a sentinel "is this endpoint secured?" marker under proxy mode rather than an enforced scope check.

Worth a brief class-level Javadoc note so future maintainers don't assume scopes are enforced here. Otherwise someone will add a new scope-only requirement and be surprised when it doesn't gate proxy-mode traffic.

calm-hub/src/main/java/org/finos/calm/security/ProxyAccessControlFilter.java
```diff
@@ -0,0 +1,67 @@
...
+
+        String username = requestContext.getHeaderString(usernameHeader);
+        if (username == null || username.isBlank()) {
```
rocketstack-matt, line 45: `isBlank()` rejects whitespace-only, but a value like `"alice "` (trailing space from a misconfigured proxy) will pass through and then miss any DB row keyed on `"alice"`. Cheap fix:

```java
username = username.trim();
```

after the blank check, before using it for the lookup.

calm-hub/src/main/java/org/finos/calm/security/ProxyAccessControlFilter.java
```diff
@@ -0,0 +1,67 @@
...
+            logger.warn("Request rejected: {} header is missing or blank", usernameHeader);
```
rocketstack-matt, line 46: Minor: `username` comes from a raw HTTP header (not a JWT-validated claim), so log-forging via CR/LF is more trivial here than in `AccessControlFilter`. `SearchResource.sanitizeForLog` already exists in this codebase — consider applying it (or at least stripping control chars) before logging the username on lines 46 and 61. Not blocking, since the same pattern exists in `AccessControlFilter`, but the attack surface is wider here.

calm-hub/src/main/java/org/finos/calm/security/ProxyAccessControlFilter.java
```diff
@@ -0,0 +1,67 @@
...
+
+        UserRequestAttributes userRequestAttributes = new UserRequestAttributes(requestMethod, username, path, namespace);
+        logger.debug("User request attributes: {}", userRequestAttributes);
+
+        if (!userAccessValidator.isUserAuthorized(userRequestAttributes)) {
+            logger.warn("No access permissions assigned to the user: [{}]", username);
+            requestContext.abortWith(Response.status(Response.Status.FORBIDDEN)
+                    .entity("Forbidden: user does not have required access grants")
+                    .build());
+        }
```
rocketstack-matt, line 65: Lines 53–65 here are essentially identical to `AccessControlFilter.authorizeUserRequest` (lines 88–104 in that file): build `UserRequestAttributes`, call `isUserAuthorized`, return 403 on failure. Consider extracting a small helper (e.g. `AuthzHelper.authorize(requestContext, username, validator)`) so both filters stay in lockstep. Not blocking, but it will drift.

calm-hub/src/test/java/org/finos/calm/resources/TestSearchResourceFilteringShould.java
```diff
@@ -77,7 +77,7 @@ void pass_resolved_readable_namespaces_to_store_in_secure_mode() {
         );
         when(mockSearchStore.search(eq("test"), any())).thenReturn(storeResults);
 
-        SearchResource resource = new SearchResource(mockSearchStore, mockValidatorInstance, mockJwtInstance);
+        SearchResource resource = new SearchResource(mockSearchStore, mockValidatorInstance, mockJwtInstance, "Remote-User");
```
rocketstack-matt, line 80: The constructor signature was updated in all four tests, but no test exercises the new proxy-header fallback branch in `SearchResource.resolveReadableNamespaces()` (lines 99–101) — i.e. the whole reason the signature changed.

Please add a test where:
- `jwtInstance.isResolvable()` returns `false` (or JWT username is null),
- `httpHeaders.getHeaderString("Remote-User")` returns a username,
- assert that username is passed to `validator.getReadableNamespaces`.

Without this the proxy-mode search path is uncovered.

calm-hub/src/test/java/org/finos/calm/security/TestProxyAccessControlFilterShould.java
```diff
@@ -0,0 +1,126 @@
...
+import org.mockito.MockitoAnnotations;
+
+import java.lang.reflect.Method;
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.Mockito.*;
+
+@QuarkusTest
```
rocketstack-matt: `@QuarkusTest` boots the full Quarkus app context for every run, but everything here is mocked — no CDI, no config, no HTTP. Drop the annotation and this becomes a plain JUnit 5 + Mockito unit test that runs in a fraction of the time. (For comparison, `AccessControlFilter` tests, if any, would likely be the same shape.)

calm-hub/README.md
```diff
@@ -120,6 +120,37 @@ From the `calm-hub` directory
...
+  - User identities will be extracted from the provided JWT which will be validated against the Authorization Server's Json Web Key Set (JWKS.)
+- `proxy`: Assumes CalmHub will be deployed behind an additional proxy component, such as `nginx` or `apache`, that performs OAuth 2 (or other) authentication for you.
+  - User identity is expected to be passed via a header; by default this is `Remote-User`.
+
+#### Proxy profile
+
+To launch CalmHub with the proxy auth mode enabled:
+
+```bash
+`../mvnw quarkus:dev -Dquarkus.profile=proxy`
```
rocketstack-matt: Stray inner backticks inside the fenced code block — they'll render literally:

```
```bash
`../mvnw quarkus:dev -Dquarkus.profile=proxy`
```
```

Should be:

```
```bash
../mvnw quarkus:dev -Dquarkus.profile=proxy
```
```

File: calm-hub/README.md [hunk 10]
```diff
@@ -120,6 +120,37 @@ From the `calm-hub` directory
...
+- This is because if users can directly call CalmHub, they can simply set the header to trivially impersonate any identity.
+
+#### Secure profile
+
+The secure profile requires an Identity Provider (IdP) to authenticate users. 
+The IdP will most likely be managed by your organisation in an enterprise environment, or by your Cloud Service Provider if you're deploying on public cloud.
+
+However, for local testing and development purposes, CalmHub includes a simple pre-configured IdP, Keycloak, that you can spin up locally to simulate a real IdP.
+
+The following sections descibe how to start Keycloak, and how to configure CalmHub to use it correctly.
```
rocketstack-matt, line 152: Typo: `descibe` → `describe`.

calm-hub/README.md
```diff
@@ -120,6 +120,37 @@ From the `calm-hub` directory
...
+To launch CalmHub with the proxy auth mode enabled:
+
+```bash
+`../mvnw quarkus:dev -Dquarkus.profile=proxy`
+```
+
+**Important notes**:
+- It is strongly recommended that the sidecar runs in the same pod or container as CalmHub, and that **CalmHub should only be accessible via this proxy.**
+
+- This is because if users can directly call CalmHub, they can simply set the header to trivially impersonate any identity.
```
rocketstack-matt, line 143: The spoofing call-out here is the most important sentence in the doc and it's in the right place — nice. Optional suggestion: consider also adding a startup-time WARN log in `ProxyAccessControlFilter` reminding operators that direct exposure trivially allows identity spoofing. Operator error (CalmHub bound to `0.0.0.0` with `proxy` profile and no sidecar) is the realistic failure mode here, and a loud log on every boot is cheap insurance.

calm-hub/src/main/java/org/finos/calm/security/ProxyAccessControlFilter.java
willosborne: This is a great point. I actually think we need to rework this. My initial plan was to just add the filter and then rework the entitlements model to do RBAC under the hood, but I think this is fundamentally the wrong abstraction.

I'm going to write up my analysis of the current situation in the codebase + plan for how we rework this to use RBAC instead of scopes on an issue before I take this PR any further.

calm-hub/src/test/java/org/finos/calm/security/TestProxyAccessControlFilterShould.java
willosborne: good shout, but i guess this probably still need @ExtendWith(MockitoExtension.class) to make @Mock work.

jpgough-ms:
@willosborne a few questions form me

Is there a reason the proxy filter doesn't reuse AccessControlFilter's authorizeUserRequest logic via composition rather than duplication? Matt suggested extracting a helper; I'd go one step further and have ProxyAccessControlFilter extend a base class that owns everything except identity extraction.

Are there other resources beyond SearchResource that read jwt.getClaim("preferred_username") directly today? If so, they all need the same proxy-fallback treatment and the CurrentUser bean (#4) becomes mandatory rather than nice-to-have.

jpgough-ms:
It looks like an integration test would be missing in this PR

willosborne:
@jpgough-ms @rocketstack-matt I don't think this PR makes any sense merged as is until we've sorted out the underlying problem you pointed out, Matt - our roles/rbac system isn't currently up to scratch. It was a bit hasty on my part :)

I think I am going to close this PR and raise a new spike one based on my proposal for auth here: #2473 
Please let me know any concerns on that one - I'm going to implement a spike which should give us some extra material to think about .

Outline of the proposal there:
- replace both AccessControlFilter and ProxyAccessControlFilter with proper Quarkus `HttpAuthenticationMechanism` instances - we only need one for proxy as the JWT one already exists
- implement a Quarkus `PermissionsChecker` for each scope
- use of proper quarkus constructs lets us remove basically all the code in `UserAccessValidator` - I need to see how this looks, but I have a feeling it may be completely redundant outside of the search logic.
- improve underlying UserAccessResource to facilitate this.

This more or less covers your comments @jpgough-ms I think. 
Agreed, integration tests definitely needed for this too.

PR #2475, opened 2026-05-20
feat(cli): allow setting of CLI config via env vars

willosborne:
## Description
Allows setting config props from OS env variables as well as the config file.
If provided they will override the config file. 

Note that --calm-hub-url and so on will still override the env variables if user has provided them.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


cli/README.md
```diff
@@ -3,6 +3,19 @@
...
+
+The CLI will load user config from `~/.calm.json` to allow profile-level config of CalmHub and other details.
+These properties can also be loaded from environment variables; if set they will override the config file.
+
+Note that if they're set on the command line, e.g. `--calm-hub-url`, this will override both the file and env vars.
+
+| Config file property | Environment variable   | Description              |
+| -------------------- | ---------------------- | ---------- |
+| `calmHubUrl`         | `CALM_HUB_URL`         | CalmHub instance to use. Note that setting this property will automatically configure CalmHub as a loading mechanism for commands such as validate. |
+| `authPluginPath`     | `AUTH_PLUGIN_PATH`     | Path to authentication plugin (should be a JS file.) See [Authentication Plugins](#authentication-plugins). |
```
markscott-ms: This documentation is out of alignment with expected env var names... which have a `CALM` prefix.

Also, should have the ordering alphabetical, i.e. CALM_ALLOWED_REMOTE_HOSTS, CALM_AUTH_PLUGIN_PATH, CALM_HUB_URL?

markscott-ms (dismissed):

cli/README.md
willosborne: good point let me fix this

cli/src/cli-config.ts
```diff
@@ -30,7 +30,7 @@ export async function loadCliConfig(): Promise<CLIConfig | null> {
         const config = await readFile(configFilePath, 'utf8');
         const parsed = JSON.parse(config) as CLIConfig;
         logger.debug('Parsed user config: ' + config);
-        return parsed;
+        return mergeWithEnvVars(parsed);
```
rocketstack-matt, line 33: Env vars only take effect when `~/.calm.json` exists. The `ENOENT` branch below returns `null` before `mergeWithEnvVars` is ever reached, so a user who sets only env vars (no config file) gets no config at all — which contradicts the README ("These properties can also be loaded from environment variables"). The new test masks this by writing an empty `{}` file first.

Suggest returning `mergeWithEnvVars({})` on `ENOENT` instead of `null`, and adding a test for the no-config-file case.

cli/src/cli-config.spec.ts
```diff
@@ -100,5 +109,42 @@ describe('cli-config', () => {
         const authPlugin = await loadAuthPlugin(config!.authPluginPath!, false);
         expect(authPlugin.getAuthHeaders).toBeDefined();
     });
+    
+    it('loads config props from environment variables', async () => {
```
rocketstack-matt, line 106: These tests set `process.env.CALM_*` but never restore them — there's no `afterEach` cleanup. They run last so CI is green, but it's a pollution risk for any future test added after them. Suggest saving/deleting the vars in an `afterEach` (or snapshotting `process.env` in `beforeEach`).

cli/src/cli-config.spec.ts
```diff
@@ -16,7 +16,16 @@ vi.mock('fs', async () => {
 });
 
 vi.mock('os', () => ({
-    homedir: vi.fn(() => '/home/user')
+    homedir: vi.fn(() => '/home/user'),
+    
+}));
+
+vi.mock('process', () => ({
```
rocketstack-matt: This `vi.mock('process', ...)` has no effect: the code under test reads the global `process.env`, not an imported `process` module, so the mock is never consulted (the tests work because they mutate `process.env` directly). Can be removed. Minor: the `os` mock above also picked up a trailing-whitespace empty line.

rocketstack-matt (commented):
A few smaller follow-ups in addition to the env var naming point already raised. All non-blocking.

cli/src/cli-config.spec.ts
markscott-ms: It's probably worth using `vi.stubEnv()` and `vi.unstubAllEnvs()`

cli/src/cli-config.spec.ts
markscott-ms: use vi.stubEnv() and vi.unstubAllEnvs(). Don't access `process`

willosborne: replaced with vi.stubEnv and the equivalent to unset

cli/src/cli-config.spec.ts
willosborne: fixed

cli/src/cli-config.ts
willosborne: great point, fixed. also checked for anything that relies on it being null - it always checks the properties as well.

willosborne:
@rocketstack-matt have addressed your feedback

rocketstack-matt (approved):

Issue #1222, opened 2025-04-29
Adding support for ADRs in the CALM Hub UI


lbulanti-ms:
### Description of Feature:
Adding support for ADRs to the CALM Hub UI

### Implementation:
* Display the ADRs in Markdown format. (We might use this library: https://github.com/remarkjs/react-markdown?tab=readme-ov-file)
* Make sure that the ADRs can be edited while keeping the defined structure

niamhg-ms:
Displaying markdown without styling:

![Image](https://github.com/user-attachments/assets/71a169ef-6e43-4fec-88da-98dd76f2d184)

PR #2500, opened 2026-05-24
feat(calm-models,calm-shared,calm-hub-ui): Set up ADR Diff Engine

aamanrebello:
## Description

Partially addresses https://github.com/finos/architecture-as-code/issues/2526 - "As a developer, I want to run calm adr-diff between two architectures so that I can identify what ADR links have been added or removed."

We will make changes to the CLI/Calm Hub UI after consultations, but for now I am just adding ADR diff capability and typing to CALM models. Types have been updated in all affected modules.

New dependencies are introduced into the monorepo - diff and @types/diff

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-models/src/diff/diff.spec.ts
+++ b/calm-models/src/diff/diff.spec.ts
@@ -311,7 +311,7 @@ describe('diff', () => {
     });
 
     describe('diffArchitectures - ADR changes', () => {
-        it('records the unchanged and added ADR when an ADR is added', () => {
+        it('records unchanged and added ADRs when an ADR is added', () => {
             const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.adrAdditionArchitecture);
             expect(result.adrDiffItems).toEqual([
                 {
@@ -319,23 +319,31 @@ describe('diff', () => {
                     changeType: 'unchanged'
                 },
                 {
-                    content: testArchitectures.adrAdditionArchitecture.adrs[1],
+                    content: testArchitectures.baseArchitecture.adrs[1],
+                    changeType: 'unchanged'
+                },
+                {
+                    content: testArchitectures.adrAdditionArchitecture.adrs[2],
                     changeType: 'added'
                 }
             ]);
         });
 
-        it('records the removed ADR when an ADR is removed', () => {
+        it('records unchanged and removed ADRs when an ADR is removed', () => {
             const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.adrRemovalArchitecture);
             expect(result.adrDiffItems).toEqual([
                 {
                     content: testArchitectures.baseArchitecture.adrs[0],
                     changeType: 'removed'
+                },
+                {
+                    content: testArchitectures.baseArchitecture.adrs[1],
+                    changeType: 'unchanged'
                 }
             ]);
         });
 
-        it('records the removed then added ADR when an ADR is modified', () => {
+        it('records the removed, added and unchanged ADRs when an ADR is modified', () => {
             const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.adrModificationArchitecture);
             expect(result.adrDiffItems).toEqual([
                 {
@@ -345,6 +353,28 @@ describe('diff', () => {
                 {
                     content: testArchitectures.adrModificationArchitecture.adrs[0],
                     changeType: 'added'
+                },
+                {
+                    content: testArchitectures.baseArchitecture.adrs[1],
+                    changeType: 'unchanged'
+                }
+            ]);
+        });
+
+        it('records the removed, added and unchanged ADRs when the ADR array is reordered', () => {
+            const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.adrReorderArchitecture);
+            expect(result.adrDiffItems).toEqual([
+                {
+                    content: testArchitectures.baseArchitecture.adrs[0],
+                    changeType: 'removed'
+                },
+                {
+                    content: testArchitectures.baseArchitecture.adrs[1],
+                    changeType: 'unchanged'
+                },
+                {
+                    content: testArchitectures.adrReorderArchitecture.adrs[1],
+                    changeType: 'added'
                 }
             ]);
         });
@@ -352,10 +382,9 @@ describe('diff', () => {
 
     describe('diffArchitectures - comprehensive scenarios', () => {
         const empty = (): CalmArchitectureSchema => ({
-            $schema: 'https://calm.finos.org/release/1.2/meta/calm.json',
             nodes: [],
             relationships: [],
-        } as CalmArchitectureSchema);
+        });
 
         it('handles empty architectures', () => {
             const result = diffArchitectures(empty(), empty());
--- a/calm-models/src/diff/diff.spec.ts
+++ b/calm-models/src/diff/diff.spec.ts
@@ -26,6 +26,7 @@ describe('diff', () => {
                 'unique-id': 'payment-service',
                 'node-type': 'service',
                 name: 'Payment Service',
+                description: 'Handles payments',
                 metadata: { version: '1.0', details: { priority: 'high', owner: 'finance' } },
             };
             const node2: CalmNodeSchema = {
@@ -33,6 +34,7 @@ describe('diff', () => {
                 'node-type': 'service',
                 metadata: { details: { owner: 'finance', priority: 'high' }, version: '1.0' },
                 name: 'Payment Service',
+                description: 'Handles payments',
             };
             expect(nodeStructureMatches(node1, node2)).toBe(true);
         });
@@ -42,11 +44,13 @@ describe('diff', () => {
                 'unique-id': 'payment-service',
                 'node-type': 'service',
                 name: 'Payment Service',
+                description: 'Handles payments',
             };
             const node2: CalmNodeSchema = {
                 'unique-id': 'payment-processor',
                 'node-type': 'service',
                 name: 'Different Name',
+                description: 'Handles payments',
             };
             expect(nodeStructureMatches(node1, node2)).toBe(false);
         });
@@ -56,12 +60,14 @@ describe('diff', () => {
                 'unique-id': 'payment-service',
                 'node-type': 'service',
                 name: 'Payment Service',
+                description: 'Handles payments',
                 metadata: { tags: ['api', 'web'] },
             };
             const node2: CalmNodeSchema = {
                 'unique-id': 'payment-processor',
                 'node-type': 'service',
                 name: 'Payment Service',
+                description: 'Handles payments',
                 metadata: { tags: ['web', 'api'] },
             };
             expect(nodeStructureMatches(node1, node2)).toBe(false);
@@ -272,7 +278,7 @@ describe('diff', () => {
         });
 
         it('skips nodes missing unique-id but surfaces them via invalidItems', () => {
-            const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.invalidShapeArchitecture);
+            const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.invalidShapeArchitecture as CalmArchitectureSchema);
             expect(
                 result.nodesAdded.length +
                     result.nodesRemoved.length +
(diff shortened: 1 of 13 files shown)


rocketstack-matt:
Thanks for the PR @aamanrebello but I think we have a bit of a sequencing issue here.

As things stand, ADRs on architectures live outside the timeline, e.g.
Screenshot 2026-06-01 at 15 10 35

It's not entirely obvious, but hopefully you can see above that the ADRs at a given point in time are in a separate 'metadata' tab above the timeline.

As your PR shows we now have ADR changes showing in the timeline, but I would still need to use the metadata tab to access them - which in the case of the comparison view is not even visible, so I would now need to navigate to a specific version to access the ADR.

This brings us on to what I would consider the biggest issue, that ADRs I don't believe are renderable at all, certainly for me clicking one via either the metadata tab or from the ADRs item in the explorer view results in an entirely blank page.

Personally, I think we need to work through the flow and design of how we would like ADRs to be presented and agree that rendering ADRs is a priority before we move on to raising a PR for them. Indeed, as a group I think we need to get better at working on items that have been prioritized on the roadmap as I think a lot of us are getting excited about a feature, raising a PR and then working on it before as a group we've agreed it is a high priority.

Looking at ADRs standalone, I actually think that a different type of interface could work better, certainly from the explorer view I think perhaps a card style view would be better, not so sure on the architecture piece.

I hope I'm not sounding too negative, I think it would be worth bringing the issue to Office Hours, ideally with some proposal for how this could work in a target state and then a number of iterative PRs towards that would absolutely make sense.

rocketstack-matt:
While exercising the ADR/timeline functionality around this PR locally, I hit a separate, **pre-existing** bug in the CALM Hub UI worth flagging here for context: the **ADR detail page renders a completely blank page** for any ADR without a populated `decisionOutcome` — i.e. every `draft` ADR created via the REST API.

Root cause is an unguarded `decisionOutcome.chosenOption` dereference in `DisplayChosenOption`, with no error boundary around `AdrRenderer`, so the `TypeError` unmounts the whole view. It is **not introduced by this PR** — it dates back to the original ADR detail page (Issue #1222) — but it surfaced while testing ADRs in support of this work.

Logged as #2548 and fixed in #2549 (UI-only guard + regression tests).

aamanrebello:
> While exercising the ADR/timeline functionality around this PR locally, I hit a separate, **pre-existing** bug in the CALM Hub UI worth flagging here for context: the **ADR detail page renders a completely blank page** for any ADR without a populated `decisionOutcome` — i.e. every `draft` ADR created via the REST API.
> 
> Root cause is an unguarded `decisionOutcome.chosenOption` dereference in `DisplayChosenOption`, with no error boundary around `AdrRenderer`, so the `TypeError` unmounts the whole view. It is **not introduced by this PR** — it dates back to the original ADR detail page (Issue #1222) — but it surfaced while testing ADRs in support of this work.
> 
> Logged as #2548 and fixed in #2549 (UI-only guard + regression tests).

Yes, an ADR renderer does exist, it's good that the bug was identified. Will take a look at the PR as soon as I can.

aamanrebello:
> Thanks for the PR @aamanrebello but I think we have a bit of a sequencing issue here.
> 
> As things stand, ADRs on architectures live outside the timeline, e.g. Screenshot 2026-06-01 at 15 10 35
> 
> It's not entirely obvious, but hopefully you can see above that the ADRs at a given point in time are in a separate 'metadata' tab above the timeline.
> 
> As your PR shows we now have ADR changes showing in the timeline, but I would still need to use the metadata tab to access them - which in the case of the comparison view is not even visible, so I would now need to navigate to a specific version to access the ADR.
> 
> This brings us on to what I would consider the biggest issue, that ADRs I don't believe are renderable at all, certainly for me clicking one via either the metadata tab or from the ADRs item in the explorer view results in an entirely blank page.
> 
> Personally, I think we need to work through the flow and design of how we would like ADRs to be presented and agree that rendering ADRs is a priority before we move on to raising a PR for them. Indeed, as a group I think we need to get better at working on items that have been prioritized on the roadmap as I think a lot of us are getting excited about a feature, raising a PR and then working on it before as a group we've agreed it is a high priority.
> 
> Looking at ADRs standalone, I actually think that a different type of interface could work better, certainly from the explorer view I think perhaps a card style view would be better, not so sure on the architecture piece.
> 
> I hope I'm not sounding too negative, I think it would be worth bringing the issue to Office Hours, ideally with some proposal for how this could work in a target state and then a number of iterative PRs towards that would absolutely make sense.

To my understanding, the presentation of the ADR diff within the CalmHub UI is the issue. Fair enough, I was just trying to follow the format of your previous PR's where you made changes to the CALM models, CALM CLI and Calm Hub UI in a single PR. To be honest, the UI part of this PR isn't as important, I will get rid of those commits (LIFO). What's more important is the changes I have made in the CALM Models and CALM CLI.

I think if we have a CLI command/function to diff two CALM architectures, we cannot just stop at nodes and edges. The diff must extend to other parts of the schema as well i.e. ADRs, controls, metadata etc. That, in my opinion, is the true meaning of diffing two CALM files. Since the ADRs are string arrays, enabling diffing of these in the CLI is an easy low-hanging fruit (you can see from the code how simple it is) - and it is useful to have the functionality available. 

So my conclusion is that I will roll back changes to the UI and ask for reviews only on my changes to CALM models and CALM CLI. Related changes to Calm Hub UI can be tackled later based on our priorities/decisions.

I am about to track this in https://github.com/finos/architecture-as-code/issues/2526

rocketstack-matt:
Thanks @aamanrebello to be clear, not asking you to roll anything back on this, more that I think we need to do two things as a community (not specific to you, I'm just as guilty, as are others).
1. Raise issues and request feedback ahead of implementation (unless the PR is a POC to show the potential, not looking to stifle creativity). in this specific case I think that would have gotten us to look more at how the ADRs are currently presented, which seems a bit out of sync with the rest of things - that is in large part my fault because I implemented the timeline feature without ever looking at ADRs.
2. Ensure we're working on things prioritized on the roadmap . . . obviously the roadmap should and will evolve, we don't look at it very often, but a whole bunch of things we decided were important we haven't looked at and we all keep raising new features not on it.

I have put this discussion on Office Hours.

aamanrebello:
What I have decided is to not directly affect any of the tooling until we discuss it at office hours. I am just adding the capability into calm-models and updating affected downstreams. This incorporates nice-to-have functionality without affecting any tooling.

Until we make the decisions, I have updated `DiffResult` to `NodesAndRelationshipsDiffResult` (only scoped to nodes and relationships) in downstreams - eventually we expect to extend the types to incorporate ADRs and everything else.

shared/src/commands/diff/diff.ts
```diff
@@ -7,12 +7,12 @@ import {
     diffTimelineAdjacent,
     diffTimelineMoments,
     type ArchitectureResolver,
-    type DiffResult,
     type MomentDiff,
     type TimelineInput,
 } from '@finos/calm-models/diff';
 import type { CalmArchitectureSchema, CalmNodeSchema, CalmRelationshipSchema } from '@finos/calm-models/types';
 import { initLogger } from '../../logger.js';
+import { NodesAndRelationshipsDiffResult } from '../../../../calm-models/dist/diff/diff-types.js';
```
rocketstack-matt: **Blocking:** this reaches across the workspace boundary by relative path into another package's *build output* (`calm-models/dist/...`). That breaks on a clean checkout / wrong build order, defeats the workspace symlink, and bypasses the public entrypoint — and it drops the `type` modifier on a type-only import.

`@finos/calm-models/diff` already re-exports `NodesAndRelationshipsDiffResult` (see `calm-models/src/diff/index.ts` in this PR), so just fold it into the existing import block above and delete this line:

```ts
import {
    diffArchitectures,
    diffPatterns,
    diffTimelineAdjacent,
    diffTimelineMoments,
    type ArchitectureResolver,
    type NodesAndRelationshipsDiffResult,
    type MomentDiff,
    type TimelineInput,
} from '@finos/calm-models/diff';
```

shared/src/commands/diff/diff.spec.ts
```diff
@@ -3,7 +3,7 @@ import { tmpdir } from 'node:os';
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
 import path from 'node:path';
 import { runDiff, formatDiff, hasChanges, detectDocumentType } from './diff.js';
-import type { DiffResult } from '@finos/calm-models/diff';
+import { NodesAndRelationshipsDiffResult } from '@finos/calm-models/diff';
```
rocketstack-matt: `NodesAndRelationshipsDiffResult` is used only as a type here — keep `import type` (it was `import type` before the rename). Consistent with the rest of the file and safer under `isolatedModules` / `verbatimModuleSyntax`:

```ts
import type { NodesAndRelationshipsDiffResult } from '@finos/calm-models/diff';
```

calm-models/src/diff/diff.spec.ts
```diff
@@ -272,7 +278,7 @@ describe('diff', () => {
         });
 
         it('skips nodes missing unique-id but surfaces them via invalidItems', () => {
-            const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.invalidShapeArchitecture);
+            const result = diffArchitectures(testArchitectures.baseArchitecture, testArchitectures.invalidShapeArchitecture as CalmArchitectureSchema);
```
rocketstack-matt, line 285: This call didn't previously need a cast. Why is `as CalmArchitectureSchema` newly required here (and on the `empty()` literal below)? It likely points at the typed JSON-fixture import no longer matching the param type under strict mode — worth understanding rather than silencing, since the cast hides whatever shape changed. (Casting is an existing idiom elsewhere in this file, so not a blocker — just flagging the *new* ones.)

rocketstack-matt (changes requested):
Thanks for this — the ADR diff engine itself is clean and well-tested, and the `NodesAndRelationshipsDiffResult` / `ArchitectureDiffResult` type split is a sensible design. A couple of things to sort before this can merge; I've left inline comments on the specifics.

**Blocking**

1. Cross-package deep import into `dist/` in `shared/src/commands/diff/diff.ts` (inline comment) — please import from the package entrypoint instead.
2. `package-lock.json` looks like an unclean regeneration. Beyond the intended `diff` / `@types/diff` additions, the diff also: removes `"dev": true` from dozens of `@esbuild/*` / `@rollup/*` optional platform packages (flipping them dev→prod), bumps `cli` `1.43.0 → 1.43.1`, and rewrites the calmstudio `@finos/calm-models` specifiers from `"*"` to `"file:..."`. These are exactly the spurious changes the `validate-lockfile` workflow guards against. Could you regenerate cleanly on Node 22 with `rm -rf node_modules package-lock.json && npm install` so only the two new deps remain?

**Should fix**

3. `import type` was dropped for a type-only import in `shared/.../diff.spec.ts` (inline comment).
4. Title scope `calm-shared` isn't a recognised scope — the project uses `shared` (i.e. `feat(calm-models,shared,calm-hub-ui)`).

Non-blocking nits I'll leave out of the formal list: the `"Mofifies"` typo in the `adrModificationArchitecture` fixture description, and the empty `@param archA`/`@param archB` JSDoc tags on `diffArchitectures`.

calm-models/src/diff/diff.spec.ts
aamanrebello: Type errors in the code:

In the referenced case, I think the node JSON does not validly conform to the CALM schema.
```
Type '{ "node-type": string; name: string; description: string; "unique-id"?: undefined; }' is not assignable to type 'CalmNodeSchema'.
          Types of property ''unique-id'' are incompatible.
            Type 'undefined' is not assignable to type 'string'.
```
Here, I will just add a comment saying that casting is done here to avoid compiler errors.


In the other case:
```
Object literal may only specify known properties, and '$schema' does not exist in type 'CalmCoreSchema'.
```
I will just remove the `$schema` property here.

shared/src/commands/diff/diff.spec.ts
aamanrebello: Fixed.

shared/src/commands/diff/diff.ts
aamanrebello: Fixed. This was a particularly bad one, thanks for identifying.

rocketstack-matt (approved):

PR #2561, opened 2026-06-03
fix(calm-hub): guard against null versions document in resource stores

### LINKED ISSUE #2498: bug(calm-hub): null `versions` document causes NPE in Mongo/Nitrite resource stores
### Steps to Reproduce:
The normal create flow always populates `versions: { … }`, so this is hard to hit in production. It is reachable when a resource document exists without a `versions` sub-document — e.g. a hand-edited Nitrite store, a partial / aborted migration, or a future schema change. In that state, the following all NPE:

- `MongoArchitectureStore.getArchitectureVersions` (~line 137) — `versions.keySet()` with no null guard.
- `MongoArchitectureStore.getArchitectureForVersion` (~line 205) — `versions.get(...)` with no null guard.
- `MongoTimelineStore.getTimelineVersions` / `getTimelineForVersion` (lines ~120 / ~162) — identical pattern, introduced in #2496.
- `NitriteArchitectureStore.getArchitectureVersions` / `getArchitectureForVersion` (~lines 162 / 205) — identical pattern; the `versionObj` retrieved in `getForVersion` is also cast to `String` gated only by a null-check, not an `instanceof` check.
- `NitriteTimelineStore` equivalents (introduced in #2496) — same cast.
- `NitriteTimelineStore.writeTimelineToNitrite` (~line 283) — `versions.put(...)` with no null guard; same pattern in the equivalent architecture write path.
- Pattern stores have the same shape.

### Expected Result:
A typed exception (e.g. `ResourceVersionsCorruptException`, or just `*NotFoundException`) and a sensible HTTP 4xx response.

### Actual Result:
`NullPointerException` propagates out of the store and the resource layer returns HTTP 500.

### Environment:
`calm-hub` on `main` after PR #2496 lands.

### Additional Context:
Surfaced during the pre-PR review of #2496. The new timeline stores in that PR deliberately mirror the un-null-checked pattern in the existing `MongoArchitectureStore` / `NitriteArchitectureStore` to keep the new code idiomatic; fixing only the timeline stores would create drift. This issue tracks the project-wide hardening across **architecture / pattern / timeline** stores in one PR.

**Suggested fixes:**
- Add a defensive null-check on the `versions` document in each `get*Versions`, `get*ForVersion`, and `write*ToNitrite` method, throwing `*NotFoundException` (or a new `*Corrupt` exception) when null.
- Add an `instanceof String` guard before casting `versionObj` in the Nitrite stores.
- Cover both with unit tests that construct a resource document with no `versions` sub-document.

YoofiTT96:
## Description
When a resource document exists without a versions sub-document (e.g. hand-edited store, partial migration), all get* and write* methods in the Mongo and Nitrite architecture/pattern/timeline stores would NPE.

Add null checks on the versions document in every affected method, throwing the appropriate *NotFoundException or *VersionNotFoundException rather than propagating a NullPointerException as HTTP 500.

Also add instanceof String guards in the Nitrite stores before casting versionObj, covering the case where the stored value is not a String.

All store implementations now behave consistently: a null versions document throws *NotFoundException from getVersions and *VersionNotFoundException from getForVersion across both Mongo and Nitrite backends.

Inline null checks were preferred over a helper method because each method throws a different checked exception type, and Java checked exceptions do not compose cleanly with functional interfaces — a helper would require an awkward multi-throws signature or an unchecked wrapper that hides intent. The three-line pattern is idiomatic Java here.

NitriteTimelineStore.getTimelineVersions previously returned an empty list (200) when the versions sub-document was absent; it now throws TimelineNotFoundException (404) for consistency with all other store implementations. This is an intentional behavior change.

Fixes #2498.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
+++ b/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
@@ -136,6 +136,9 @@ public class MongoArchitectureStore implements ArchitectureStore {
             if (architecture.getId() == architectureDoc.getInteger("architectureId")) {
                 // Extract the versions map from the matching pattern
                 Document versions = (Document) architectureDoc.get("versions");
+                if (versions == null) {
+                    throw new ArchitectureNotFoundException();
+                }
                 Set<String> versionKeys = versions.keySet();
 
                 //Convert from Mongo representation
@@ -176,6 +179,9 @@ public class MongoArchitectureStore implements ArchitectureStore {
             if (architecture.getId() == architectureDoc.getInteger("architectureId")) {
                 // Retrieve the versions map from the matching pattern
                 Document versions = (Document) architectureDoc.get("versions");
+                if (versions == null) {
+                    throw new ArchitectureVersionNotFoundException();
+                }
 
                 // Return the pattern JSON blob for the specified version
                 Document versionDoc = (Document) versions.get(architecture.getMongoVersion());
--- a/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
+++ b/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
@@ -125,6 +125,9 @@ public class MongoPatternStore implements PatternStore {
             if (pattern.getId() == patternDoc.getInteger("patternId")) {
                 // Extract the versions map from the matching pattern
                 Document versions = (Document) patternDoc.get("versions");
+                if (versions == null) {
+                    throw new PatternNotFoundException();
+                }
                 Set<String> versionKeys = versions.keySet();
 
                 //Convert from Mongo representation
@@ -165,6 +168,9 @@ public class MongoPatternStore implements PatternStore {
             if (pattern.getId() == patternDoc.getInteger("patternId")) {
                 // Retrieve the versions map from the matching pattern
                 Document versions = (Document) patternDoc.get("versions");
+                if (versions == null) {
+                    throw new PatternVersionNotFoundException();
+                }
 
                 // Return the pattern JSON blob for the specified version
                 Document versionDoc = (Document) versions.get(pattern.getMongoVersion());
--- a/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoTimelineStore.java
+++ b/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoTimelineStore.java
@@ -117,6 +117,9 @@ public class MongoTimelineStore implements TimelineStore {
             if (timeline.getId() == timelineDoc.getInteger("timelineId")) {
                 // Extract the versions map from the matching timeline
                 Document versions = (Document) timelineDoc.get("versions");
+                if (versions == null) {
+                    throw new TimelineNotFoundException();
+                }
                 Set<String> versionKeys = versions.keySet();
 
                 // Convert from Mongo representation
@@ -157,6 +160,9 @@ public class MongoTimelineStore implements TimelineStore {
             if (timeline.getId() == timelineDoc.getInteger("timelineId")) {
                 // Retrieve the versions map from the matching timeline
                 Document versions = (Document) timelineDoc.get("versions");
+                if (versions == null) {
+                    throw new TimelineVersionNotFoundException();
+                }
 
                 // Return the timeline JSON blob for the specified version
                 Document versionDoc = (Document) versions.get(timeline.getMongoVersion());
--- a/calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteArchitectureStore.java
+++ b/calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteArchitectureStore.java
@@ -162,6 +162,9 @@ public class NitriteArchitectureStore implements ArchitectureStore {
             if (architecture.getId() == architectureDoc.get(ARCHITECTURE_ID_FIELD, Integer.class)) {
                 // Extract the versions map from the matching architecture
                 Document versions = architectureDoc.get(VERSIONS_FIELD, Document.class);
+                if (versions == null) {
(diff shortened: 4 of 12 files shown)


rocketstack-matt (commented):
Reviewed — this correctly hardens the architecture / pattern / timeline stores and fully closes the scope of #2498 (which deliberately limited itself to those three resource types). Nice call leaving the Mongo write paths untouched; they use `$set` operators and can't NPE, so guarding them would have been dead code. The `instanceof String` tightening is a genuine improvement over the old null-only check.

Two things before merge:

**1. Missing tests for three new write-path guards.** The new null-`versions` guards in `NitritePatternStore.createPatternForVersion`, `NitritePatternStore.updatePatternForVersion`, and `NitriteTimelineStore.writeTimelineToNitrite` have no covering tests — only `writeArchitectureToNitrite` is exercised (via `testUpdateArchitectureForVersion_whenVersionsDocumentIsNull`). Could you add three small tests mirroring that one? Per the repo's "all new code has tests" rule.

**2. Heads-up on a response-contract change (no action needed, just flagging for the description/changelog).** `NitriteTimelineStore.getTimelineVersions` previously returned an empty list (200) when `versions` was null; it now throws `TimelineNotFoundException` (404). That's the one path that didn't already NPE, so it's a real behavior change — intentional and correct for consistency, just worth a line in the PR description.

Scope note: the `interface` and `standard` stores carry the identical unguarded pattern, but those were **out of scope for #2498**, so this PR is right not to touch them. I've raised #2564 to track that follow-up.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
```diff
@@ -136,6 +136,9 @@ public List<String> getArchitectureVersions(Architecture architecture) throws Na
             if (architecture.getId() == architectureDoc.getInteger("architectureId")) {
                 // Extract the versions map from the matching pattern
                 Document versions = (Document) architectureDoc.get("versions");
+                if (versions == null) {
```
aamanrebello, line 139: could be made a function if you're repeating this in two spaces, it's kind of a repeatable gate. Up to you on this one though.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
```diff
@@ -125,6 +125,9 @@ public List<String> getPatternVersions(Pattern pattern) throws NamespaceNotFound
             if (pattern.getId() == patternDoc.getInteger("patternId")) {
                 // Extract the versions map from the matching pattern
                 Document versions = (Document) patternDoc.get("versions");
+                if (versions == null) {
```
aamanrebello, line 128: This is also a repeatable gate.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoTimelineStore.java
```diff
@@ -117,6 +117,9 @@ public List<String> getTimelineVersions(Timeline timeline) throws NamespaceNotFo
             if (timeline.getId() == timelineDoc.getInteger("timelineId")) {
                 // Extract the versions map from the matching timeline
                 Document versions = (Document) timelineDoc.get("versions");
+                if (versions == null) {
```
aamanrebello, line 120: You may even be able to make a helper function across the different files.

aamanrebello (dismissed):
I would advise simplifying repetitive code.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoTimelineStore.java
YoofiTT96: Same as above

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
YoofiTT96: Yeah fair point. This implementation was preferred because each throws a different checked exception type and a central method would have had to do some awkward throws signature. More convoluted than the fix requires

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
YoofiTT96: Same as above

rocketstack-matt (commented):
Re-reviewed after the push — all three requested write-path tests are in (`cf42cc4b`):

- `testCreatePatternForVersion_whenVersionsDocumentIsNull_throwsPatternNotFoundException`
- `testUpdatePatternForVersion_whenVersionsDocumentIsNull_throwsPatternNotFoundException`
- `testUpdateTimelineForVersion_whenVersionsDocumentIsNull_throwsTimelineNotFoundException`

Each builds a resource document with no `versions` sub-document, calls the correct write method, and asserts the right exception — consistent with the existing get-path tests. I ran the two affected Nitrite classes locally: **55 tests, 0 failures, BUILD SUCCESS**.

Every store touched by this PR (architecture / pattern / timeline, both backends, get + write paths) is now guarded and covered. This closes the scope of #2498. The remaining `interface` / `standard` drift stays tracked in #2564.

LGTM. 🟢

rocketstack-matt (approved):
Approved — null-versions guards across architecture/pattern/timeline stores (both backends, get + write paths), all covered by tests (ran the Nitrite classes locally: 55 passed). Closes #2498; interface/standard follow-up tracked in #2564.

Issue #2577, opened 2026-06-05
Enhance Docify To Render PNG/SVG images


LeighFinegold:
### Target Project:

**`cli`** (`@finos/calm-cli`) — the `calm docify` command — with a secondary update to `calm-ai/tools/documentation-creation.md` to document the new capability.

### Description of Feature:

Today `calm docify` embeds Mermaid diagram source code (` ```mermaid … ``` `) in generated Markdown/HTML output. Downstream consumers that want to embed diagrams into their own documentation websites, slide decks, or Word/PDF documents must run a separate Mermaid CLI step to convert those code blocks into image files.

This feature extends `calm docify` with an `--export-diagrams` flag that renders every Mermaid block produced during docification into a standalone **PNG** or **SVG** file, placing them alongside the generated output, and replacing the fenced code block with an `<img>` / `![]()` reference so the images are immediately usable.

The VSCode extension already demonstrates this at runtime — `mermaid-renderer.ts` calls `mermaid.render()` and returns an inline SVG string. This feature brings the same capability to the CLI, writing files to disk instead of injecting into a webview, with no browser dependency (safe for corporate environments common in the regulated financial-services sector CALM targets).

A second part of the work updates the `calm-ai/tools/documentation-creation.md` skill so AI assistants are aware of the new option and can guide users through it.

### User Stories:

- **As a technical writer**, I want `calm docify` to produce PNG images for every architecture diagram, so that I can drag-and-drop them directly into a Confluence page, Google Slide, or PowerPoint deck without needing to install extra tooling.
- **As a platform engineer**, I want `calm docify --export-diagrams svg` to emit SVG files next to the generated Markdown, so that I can include them as first-class assets in my organisation's documentation website (e.g. Docusaurus, MkDocs) via normal `<img>` tags.
- **As a CI/CD pipeline author**, I want diagram images to be generated in a single `calm docify` invocation, so that my pipeline remains simple and produces fully self-contained artefacts with no post-processing step.
- **As an AI assistant user**, I want the documentation-creation skill to tell me about `--export-diagrams`, so that I receive accurate guidance when asking how to bundle CALM diagrams into my own docs site.

### Current Limitations:

- `calm docify` always emits raw ` ```mermaid ``` ` fenced code blocks; it never produces binary image files.
[...]

PR #2634, opened 2026-06-11
feat(cli): add --export-diagrams flag to render mermaid diagrams via local browser

YoofiTT96:
## Description
Adds a `--export-diagrams <svg|png>` flag to `docify`. When set, after the normal template rendering pass, docify post-processes the generated markdown and renders any ` ```mermaid ``` ` code blocks into SVG/PNG image files using a real, locally-installed Chromium-family browser (`playwright-core`), replacing the code block with an image reference (`Diagram N`).

This addresses #2577 using the "local browser" approach discussed on that issue: rather than bundling/downloading a browser or building a custom dagre/ELK layout renderer, we drive mermaid + `@mermaid-js/layout-elk` inside an already-installed Chrome/Edge (or any Chromium-based browser via `--browser-path`), giving visual parity with the VSCode previews.

Key points:
- **Opt-in only**: with no `--export-diagrams`, behaviour and output are unchanged (verified no `playwright-core`/`mermaid`/`@mermaid-js/layout-elk` references reach the `calm-server` bundle, and the VSCode extension never triggers this path).
- **Browser detection**: tries `playwright-core`'s `chrome` then `msedge` channels (zero-config on most corporate machines); `--browser-path` is an escape hatch for other Chromium-based browsers (Brave, Vivaldi, Chromium, etc).
- **Recoverable failure modes**: no browser found, or a single diagram failing to render (timeout / invalid mermaid syntax), produce a warning and leave that diagram as a mermaid code block -- the rest of the docify output is unaffected. An invalid `--browser-path` is the one fatal case (`process.exit(1)`).
- **Performance tradeoff surfaced**: help text, a one-time info log when the pass starts, and a final summary line with elapsed time and rendered/failed counts.
- New `--diagram-render-timeout <ms>` option controls the per-diagram render timeout (default 30s).

## Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code style/formatting changes
- [ ] Refactoring (no functional changes)
- [ ] Performance improvements
- [ ] Test additions or updates
- [ ] Chore (maintenance, dependencies, CI, etc.)

## Affected Components
- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Commit Message Format
- `feat(cli): add --export-diagrams flag to render mermaid diagrams via local browser`

## Testing
- [x] All existing tests pass

New unit tests cover `browser-launch`, `mermaid-browser-renderer`, `markdown-diagram-processor`, and `docifier`/`cli` wiring. A new `docifier-diagram-export.e2e.spec.ts` exercises the full pipeline with a real local browser (skips gracefully if none is available).

Manually verified via `npm run build:cli && npm run link:cli` then `calm docify -a <fixture> -o <out> --export-diagrams svg|png`, confirming `_diagrams/*.svg` / `*.png` are produced and referenced from the generated markdown, and that omitting `--export-diagrams` leaves output unchanged.

## Checklist

## Security considerations
- Mermaid is initialized with `securityLevel: 'strict'`. Diagram source is built from CALM model content (node/relationship names, descriptions, etc.), so this sanitizes/sandboxes any HTML that ends up in diagram labels rather than rendering it directly.
- The local asset server (which serves the `mermaid`/`@mermaid-js/layout-elk` bundles to the headless browser) binds to `127.0.0.1` on an OS-assigned port (`listen(0)`) -- never reachable over the network -- and is only alive for the duration of the export pass before being disposed.
- That server includes a path-traversal guard: requested asset paths are resolved and must stay within the relevant package's `dist` directory, or the request gets a 403.
- No browser binaries are bundled or downloaded -- `playwright-core` (not `playwright`) only drives a Chromium-family browser the user already has installed, via `chrome`/`msedge` channel auto-detection or an explicit `--browser-path`.
- `--browser-path` is validated (`fs.existsSync` + `statSync().isFile()`) before being passed to `executablePath`; an invalid path fails fast with a fatal error rather than being passed through silently.

## Notes for reviewers
- `playwright-core`, `mermaid`, and `@mermaid-js/layout-elk` are added as dependencies but marked `external` in the `cli`/`calm-server`/vscode `tsup` configs and are only `import()`/`require.resolve()`d lazily inside the diagram-rendering code paths, so they're never bundled or loaded unless `--export-diagrams` is used.
- `--export-diagrams` and `--browser-path`/`--diagram-render-timeout` reuse naming from the shelved `feature/2577-enhance-docify-to-export-png-svg` branch (a different, dagre-based implementation of the same flag) -- whichever lands first effectively claims this flag name; the other will need to reconcile.

## Possible future direction
- The VS Code extension's preview webview already renders mermaid diagrams in a real Chromium-based context, so a similar "export diagram" feature there likely would not need `playwright-core` at all -- it could grab the already-rendered `<svg>` (and rasterize to PNG via an in-webview `<canvas>`) and send it back to the extension to save. This is a separate, smaller piece of follow-up work and out of scope for this PR.


shared/src/docify/docifier.spec.ts
```diff
@@ -1,9 +1,31 @@
...
+import type { Browser } from 'playwright-core';
+import type { Logger } from '../logger';
 
 vi.mock('../template/template-processor');
 vi.mock('../template/url-mapping');
+vi.mock('./diagram-rendering/browser-launch');
+vi.mock('./diagram-rendering/mermaid-browser-renderer');
+vi.mock('./diagram-rendering/markdown-diagram-processor');
+
+const mockLogger: Logger = {
```
LeighFinegold: This appears in a lot of test classes. Probably worth looking at seeing this can be moved to a test utility class.

shared/src/docify/docifier.ts
```diff
@@ -51,5 +70,49 @@ export class Docifier {
...
+        const format = this.exportDiagrams!;
+        const startTime = Date.now();
+
+        logger.info(`ℹ️ --export-diagrams ${format}: rendering mermaid diagrams via a local browser. ` +
+            'This adds processing time (browser startup plus per-diagram rendering).');
+
+        let launched: LaunchedBrowser;
+        try {
+            launched = await launchBrowser({ browserPathOverride: this.browserPath });
+        } catch (err) {
```
LeighFinegold, line 89: So if --export-diagrams is requested and no browser is found, docify still succeeds and leaves Mermaid blocks unchanged?

shared/src/docify/docifier.ts
```diff
@@ -51,5 +70,49 @@ export class Docifier {
 
     public async docify(): Promise<void> {
         await this.templateProcessor.processTemplate();
+        if (this.exportDiagrams) {
+            await this.renderDiagrams();
+        }
+    }
+
+    private async renderDiagrams(): Promise<void> {
+        const logger = initLogger(false, 'docify-diagram-export');
```
LeighFinegold: If --verbose is used on cli will that be respected here?

shared/src/docify/docifier.ts
```diff
@@ -51,5 +70,49 @@ export class Docifier {
...
+        const format = this.exportDiagrams!;
+        const startTime = Date.now();
```
LeighFinegold, line 81: Wondering if this timing could be abstracting in the future where we are timing from outside

await this.renderDiagrams();

shared/src/docify/docifier-diagram-export.e2e.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    __dirname,
+    '../../test_fixtures/command/generate/expected-output'
+);
+
+const OUTPUT_DIR = join(
+    __dirname,
+    '../../test_fixtures/docify/diagram-export/actual-output'
+);
+
+const PNG_MAGIC_BYTES = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]);
```
LeighFinegold: Minor: Could we rename this to something like PNG_SIGNATURE or PNG_FILE_SIGNATURE, or add a short comment indicating these are the standard PNG header bytes being used to verify the generated file is a valid PNG

const PNG_SIGNATURE = Buffer.from('89504E470D0A1A0A', 'hex');

shared/src/docify/docifier.spec.ts
YoofiTT96: Makes sense will update with a small helper method for that

shared/src/docify/docifier.ts
YoofiTT96: Correct. Diagram processing is a post step hence if no browser is found a warning message is logged and mermaid is left intact

shared/src/docify/docifier.ts
YoofiTT96: No. Updating to obtain from set environment so it does

shared/src/docify/docifier.ts
YoofiTT96: It could. Depends on if there are any immediate operations we would like to time. This is the only place in docifier.ts doing "time an operation and report it," so a generic abstraction would add indirection for a single call site with no current reuse

rocketstack-matt:
@YoofiTT96 anything remaining here other than resolving the `package-lock.json` conflict?

YoofiTT96:
> @YoofiTT96 anything remaining here other than resolving the `package-lock.json` conflict?

@rocketstack-matt Nothing from an implementation point of view. Was wondering if @LeighFinegold had any issues with docify leaving the mermaid intact and logging a warning when --export-diagram fails

shared/src/docify/docifier.ts
LeighFinegold: We can cover in subsequent PR

LeighFinegold:
> > @YoofiTT96 anything remaining here other than resolving the `package-lock.json` conflict?
> 
> @rocketstack-matt Nothing from an implementation point of view. Was wondering if @LeighFinegold had any issues with docify leaving the mermaid intact and logging a warning when --export-diagram fails

We can await feedback from people post merge.

LeighFinegold (approved):

rocketstack-matt (approved):

Issue #2645, opened 2026-06-15
User Facing CALM Hub API


jpgough-ms:
Based on feedback in #2628, #2632, #2631, #2609, #1059, we are looking to make significant changes to the CalmHub API. The existing API will be moved to the `/api` route, and continue to work as before. Rather than having a Front Controller/Mapping Controller, the following changes will be made

## API Changes

- **New `/calm` base path** — `MappingControllerResource` moved from `/calm/namespaces` to `/calm`, covering both namespace resources and domain control resources under a single unified entrypoint.
- **Dual-surface design preserved** — the existing numeric-ID `/api` endpoints (PatternResource, ArchitectureResource, etc.) are unchanged; `/calm` is an additive user-facing layer on top.
- **Three write endpoints on `/calm`:**
  - `POST /calm` — document-body-only create/add-version, version driven entirely by `$id`
  - `POST /calm/namespaces/{namespace}/{type}/{name}/versions/{version}` — explicit path + `$id` must agree
  - `PUT /calm` — in-place version update, gated by `allow.put.operations` config flag (default `false`)
- **`GET /calm/namespaces/{namespace}/{type}/{name}/versions/{version}`** — returns a specific version with the versioned `$id` rewritten into the response body. Only explicit versions are supported; the versionless/latest GET has been removed.
- **Domain control resources** routed via the same `POST /calm` endpoint — domain vs namespace is determined by inspecting whether the `$id` starts with `{baseUrl}/calm/domains/` or `/calm/namespaces/`.
- **Write-scope enforcement** — `POST /calm` and `PUT /calm` now call `CalmHubPermissionChecker` programmatically (`canWrite` for namespace resources, `canWriteByDomain` for domain controls); returns `403` on failure.

## `POST /calm` — How It Works

1. **Parse `$id`** — `CalmDocumentParser` extracts the `$id` field from the request body JSON.
2. **Route by prefix** — if `$id` starts with `{baseUrl}/calm/domains/` it is a domain control document; otherwise it is a namespace resource.
3. **Validate `$id` structure** — namespace, resource type, name and version are extracted and validated against their respective regexes (`NAMESPACE_REGEX`, `CUSTOM_ID_REGEX`, `VERSION_REGEX`). Version is mandatory; the `$id` must start with the configured `calm.hub.base-url`.
4. **Permission check** — `canWrite(identity, namespace)` or `canWriteByDomain(identity, domain)` is called; returns `403` if denied.
[...]

PR #2653, opened 2026-06-15
New User Facing CalmHub API

jpgough-ms:
## Description

Rework for #2645

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [ ] All existing tests pass

## Checklist

### DIFF
--- a/calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
+++ b/calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
@@ -68,8 +68,8 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
         }
         const versionPath = data.version.replace(/\./g, '-');
         const target = isSlug(data.id)
-            ? `/calm/namespaces/${data.name}/${data.id}/versions/${versionPath}`
-            : `/calm/namespaces/${data.name}/architectures/${data.id}/versions/${versionPath}`;
+            ? `/api/calm/namespaces/${data.name}/${data.id}/versions/${versionPath}`
+            : `/api/calm/namespaces/${data.name}/architectures/${data.id}/versions/${versionPath}`;
         calmService.fetchDecoratorValues(data.name, target, 'deployment').then(setDecoratorsState);
     }, [data, fileInstance, decoratorsProp, calmService]);
 
--- a/calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
+++ b/calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
@@ -66,11 +66,27 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
             setDecoratorsState([]);
             return;
         }
+        let cancelled = false;
         const versionPath = data.version.replace(/\./g, '-');
-        const target = isSlug(data.id)
-            ? `/api/calm/namespaces/${data.name}/${data.id}/versions/${versionPath}`
-            : `/api/calm/namespaces/${data.name}/architectures/${data.id}/versions/${versionPath}`;
-        calmService.fetchDecoratorValues(data.name, target, 'deployment').then(setDecoratorsState);
+        const namespace = data.name;
+        const rawId = data.id;
+
+        async function fetchDecorators() {
+            let numericId: string = rawId;
+            if (isSlug(rawId)) {
+                const mappings = await calmService.fetchMappings(namespace, 'Architectures');
+                const match = mappings.find((m) => m.customId === rawId);
+                if (!match) return; // cannot resolve slug — skip decorator fetch
+                numericId = String(match.numericId);
+            }
+            if (cancelled) return;
+            const target = `/api/calm/namespaces/${namespace}/architectures/${numericId}/versions/${versionPath}`;
+            const values = await calmService.fetchDecoratorValues(namespace, target, 'deployment');
+            if (!cancelled) setDecoratorsState(values);
+        }
+
+        fetchDecorators().catch(() => { if (!cancelled) setDecoratorsState([]); });
+        return () => { cancelled = true; };
     }, [data, fileInstance, decoratorsProp, calmService]);
 
     const decorators = decoratorsProp ?? decoratorsState;
--- a/calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
+++ b/calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
@@ -306,6 +306,9 @@ public class MongoMappingControllerIntegration {
                 .then()
                 .statusCode(201)
                 .header("Location", containsString("/calm/domains/security/controls/access-control/requirement/versions/1.0.0"));
+    }
+
+    @Test
     @Order(22)
     void user_facing_get_controls_returns_created_control() {
         given()
--- a/calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
+++ b/calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
@@ -54,7 +54,7 @@ public class MongoMappingControllerIntegration {
     @Test
     @Order(1)
     void create_pattern_via_name_based_api() {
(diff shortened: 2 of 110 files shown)


calm-hub/src/main/java/org/finos/calm/resources/MappingControllerResource.java
```diff
@@ -29,380 +31,552 @@
...
+                    "For an existing resource the requested version is created (409 if it already exists)."
     )
-    @PermissionsAllowed(CalmHubScopes.WRITE)
-    public Response createOrUpdateResource(
-            @PathParam("namespace") @jakarta.validation.constraints.Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
-            @PathParam("customId") @jakarta.validation.constraints.Pattern(regexp = CUSTOM_ID_REGEX, message = CUSTOM_ID_MESSAGE) String customId,
-            String requestBody
-    ) throws URISyntaxException {
+    @Authenticated
+    public Response createResourceFromDocument(String requestBody) throws URISyntaxException {
```
willosborne, line 99: i think the code in the body of this method is more or less exactly the same as `handlePost`, other than that it has to parse the ID. could maybe reduce some duplication by refactoring to parse and then call that?

calm-hub/src/main/java/org/finos/calm/resources/MappingControllerResource.java
```diff
@@ -29,380 +31,552 @@
...
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity("Unsupported resource type: " + STRICT_SANITIZATION_POLICY.sanitize(type)).build();
+        }
         try {
-            ResourceMapping mapping = mappingStore.getMapping(namespace, customId);
+            ResourceMapping mapping = mappingStore.getMapping(namespace, name);
             List<String> versions = getVersionsForMapping(mapping);
             List<String> sortedVersions = versions.stream()
-                    .sorted(Comparator.comparing(Semver::tryParse))
+                    .sorted(java.util.Comparator.comparing(org.finos.calm.domain.Semver::tryParse))
```
willosborne: does this need fully qualifying?

calm-hub/src/main/java/org/finos/calm/resources/MappingControllerResource.java
willosborne, line 1: i think this is good behaviour wise, but it is now a 1000+ line file 😆 not a blocker by any means but I think we should explore whether we can split some of this up e.g. by delegating to a service

calm-hub/src/main/resources/application.properties
```diff
@@ -2,6 +2,9 @@ quarkus.swagger-ui.always-include=true
 quarkus.smallrye-openapi.info-title=Calm Hub
 #If using in quarkus dev mode can be 'mongo' or 'standalone'
 calm.database.mode=mongo
+# Base URL used to verify and rewrite $id in POST bodies to the name-based API.
+# When unset, $id verification is skipped (developer-friendly default).
+# calm.hub.base-url=http://localhost:8080
```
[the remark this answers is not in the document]
willosborne: this seems valid, but IMO this mode should never be used 😆

calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
willosborne, line 1: Test cases look good.

willosborne (dismissed):

willosborne (approved):
fixes look good

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -66,11 +66,27 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
...
+        async function fetchDecorators() {
+            let numericId: string = rawId;
+            if (isSlug(rawId)) {
+                const mappings = await calmService.fetchMappings(namespace, 'Architectures');
+                const match = mappings.find((m) => m.customId === rawId);
+                if (!match) return; // cannot resolve slug — skip decorator fetch
+                numericId = String(match.numericId);
+            }
+            if (cancelled) return;
+            const target = `/api/calm/namespaces/${namespace}/architectures/${numericId}/versions/${versionPath}`;
```
rocketstack-matt, line 83: The decorator `target` filter prefix now diverges between two sibling components. This was updated to `/api/calm/namespaces/{ns}/architectures/{numericId}/versions/{v}`, but `DiagramSection.tsx:224` (not touched by this PR) still uses `/calm/namespaces/{ns}/architectures/{id}/versions/{v}`. Both feed the same `fetchDecoratorValues(namespace, target, 'deployment')`, where `target` is an exact-match `?target=` filter against the stored decorator target string. Before this PR both used `/calm/...`; now they disagree on the prefix, so for any given stored decorator at most one of the two surfaces will match.

Whatever the canonical persisted form is, these two should agree — worth aligning `DiagramSection` (and confirming which prefix decorator targets are actually stored with) so deployment decorators render consistently in both views.

calm-hub/src/main/java/org/finos/calm/resources/CalmDocumentParser.java
```diff
@@ -0,0 +1,381 @@
...
+        }
+        String expectedPrefix = normalizeBase() + "/calm/namespaces/";
+        if (!id.startsWith(expectedPrefix)) {
+            throw new IllegalArgumentException(
+                    "$id must start with: " + expectedPrefix + "{namespace}/{type}/{name}/versions/{version}");
+        }
+        // remainder: {namespace}/{type}/{name}/versions/{version}
+        String remainder = id.substring(expectedPrefix.length());
+        String[] parts = remainder.split("/", -1);
+        if (parts.length < 5 || !"versions".equals(parts[3])) {
```
rocketstack-matt, line 172: `parseCanonicalId` is lenient on trailing `$id` segments: this checks `parts.length < 5` rather than `== 5`, so an `$id` like `.../{name}/versions/{v}/extra` parses and silently ignores the tail. The path-based POST uses a strict `equals` (`resolveAndVerify`), but `POST /calm` and `PUT /calm` go through this lenient parser. Low impact (the version regex still constrains `parts[4]`), but a stricter length check would tighten the contract and reject malformed canonical URLs.

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
```diff
@@ -0,0 +1,662 @@
...
+            return invalidJsonResponse("Request body is required");
+        }
+        // Validate $id against canonical path URL
+        String expectedId = documentParser.normalizeBase() + "/calm/domains/" + domain
+                + "/controls/" + controlName + "/requirement/versions/" + version;
+        String actualId;
+        try {
+            actualId = documentParser.extractIdFromJson(requestBody);
+        } catch (Exception e) {
+            return invalidJsonResponse("Cannot parse request body as JSON");
```
rocketstack-matt, line 227: Dead catch branch (cosmetic). `extractIdFromJson` swallows `JsonProcessingException` internally and returns `null`, so this `try/catch (Exception) -> "Cannot parse request body as JSON"` is unreachable — malformed JSON falls through to the `"$id is required"` response below instead. Same 400 status, just a slightly less precise message. Same pattern repeats in `handleControlConfigurationPost` (~line 293).

rocketstack-matt (commented):
A few findings from an independent review (separate from the points already addressed). Full context is in each inline comment.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
jpgough-ms: Created a bug so we can get the big PR merged and fix this

Issue #2654, opened 2026-06-15
CALM Hub UI displays wrong in memory resources for different namespace when user is not authorised


lbulanti-ms:
## Bug Report

The CALM Hub UI displays the resources that are in memory when a user is trying to access a resource (Architecture, Pattern, etc.) without being authorised to it. 

This happens only when a user first view a specific resource in a namespace where they have the correct permission and then go on to try access the same resource in a different namespace. 

Example:

user view architectures in finos namespace, the dropdown shows Architecture-1
user tries to view architectures in user-cannot-see-namespace, the dropdown down still shows the in memory Architecture-1

### Steps to Reproduce:
 1. start CALM Hub using the secure profile
 2. access resource that user can access
 3. try to access the same resource that user cannot see 

### Expected Result:
I would expect to see the pop up that says user is not authorised to view the resource and in the dropdown in the left no resources are showing in the resource section of the namespace which the user is not authorised to see

### Actual Result:
I see the pop up that says user is not authorised, but in the drop down the in memory architecture is showing

willosborne:
@lbulanti-ms this is fixed now, right?

PR #2666, opened 2026-06-16
fix(calm-hub-ui): calm ui resets the tree-navigation and visualizatio…

lbulanti-ms:
…n when switching namespace

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [X] CALM Hub UI (`calm-hub-ui/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [X] All existing tests pass

## Checklist
- [X] I have updated documentation if necessary


calm-hub-ui/src/hub/components/tree-navigation/TreeNavigation.tsx
```diff
@@ -652,17 +673,20 @@ export function TreeNavigation({ onDataLoad, onAdrLoad, onControlLoad, onInterfa
...
                 });
             } else if (type === 'ADRs') {
+                setAdrSummaries([]);
                 adrService.fetchAdrSummaries(selectedNamespace).then(setAdrSummaries);
             } else {
                 const setter = type === 'Architectures' ? setArchitectureSummaries
                     : type === 'Patterns' ? setPatternSummaries
                     : type === 'Flows' ? setFlowSummaries
                     : setStandardSummaries;
+                setter([]);
```
rocketstack-matt, line 591: Clearing the list before fetching handles the pending window, but a slower fetch from the *previous* namespace can still resolve afterward and repopulate it — the `fetcher(...).then(...)` below calls `enrichWithMappings`, which calls `setter` unconditionally (line 521) with no guard on the current selection. So #2654 can still reproduce when the authorised namespace's request is slow: switch namespace → 403 clears the list → the in-flight prior request resolves → stale summaries reappear under the "Access denied" modal. Track the latest request with a `useRef` sequence (this file already uses `useRef` for the load callbacks) and drop non-current responses:

```ts
const seq = ++reqSeq.current;
fetcher(selectedNamespace).then((summaries: ResourceSummary[]) => {
    if (seq === reqSeq.current) enrichWithMappings(type, summaries, setter);
});
```

The ADR and interface fetches need the same guard. (This is Copilot's low-confidence point — flagging because it's the root of the reported bug, not just a pending-state nicety.)

rocketstack-matt (commented):
Solid, correctly-scoped fix with good test coverage; closes the reported scenario for the common fast-403 path. One substantive gap remains — see the inline note: the stale-response race (Copilot's low-confidence point) is the actual root of #2654 and can still reproduce it when the authorised namespace's request is slow, since the clear-before-fetch only covers the pending window, not late resolution.

rocketstack-matt (changes requested):

aidanm3341 (dismissed):

jpgough-ms:
Pull in latest, changes needed @lbulanti-ms

calm-hub-ui/src/hub/components/tree-navigation/TreeNavigation.tsx
lbulanti-ms: @rocketstack-matt I addressed this comment - can you please review again?

rocketstack-matt (approved):
The `reqSeq` guard plus the three late-resolution tests address my earlier changes-requested on the stale-response race, for every path that can actually reach the screen: resource lists render gated on `selectedType`/`selectedDomain`, both of which the new `authStore` handler clears, so a response from a superseded request can't repopulate the view after a 403. Copilot's missing-`.catch` points are all handled too. Good to merge.

PR #2675, opened 2026-06-17
feat(vscode): export diagrams as SVG/PNG from the Docify preview panel

### LINKED ISSUE #2674: feat(vscode): export diagrams as SVG/PNG from the Docify preview panel
### Target Project:
`calm-plugins/vscode` — the CALM VSCode extension.

### Description of Feature:
Adds an **Export** dropdown to the diagram toolbar in the extension's Docify preview panel, letting users save any rendered Mermaid diagram directly to disk as an SVG or PNG file — without leaving the IDE.

### User Stories:
- As an **architect**, I want to export a rendered CALM diagram as an SVG so that I can embed a vector-quality image in documentation or presentations.
- As a **developer**, I want to export a diagram as a PNG so that I can share a snapshot with team members who don't have the extension installed.
- As a **tech writer**, I want a one-click export from the preview panel so that I don't have to run the CLI's `--export-diagrams` flag separately just to get an image of a single diagram.

### Current Limitations:
The CLI's `--export-diagrams` flag (merged in #2634) renders all Mermaid blocks in a documentation output directory via a local browser. That works well for batch documentation generation, but the VSCode extension's Docify preview panel has no export capability of its own. Users who want a single diagram image must either:
- Run `calm docify --export-diagrams svg` from the terminal and hunt for the output file, or
- Take a screenshot of the preview panel.

### Proposed Implementation:

**Webview side** (`calm-plugins/vscode/src/features/preview/webview/`):

https://github.com/user-attachments/assets/e8da5694-0ffe-4d5d-9432-aef119f44364

- `diagram-export.ts` — captures the live SVG from the diagram container and applies three fixes before serialising. For PNG  rasterises it via `html-to-image`.
- `diagram-controls.ts` — replaces the former separate SVG/PNG buttons with a single **Export** dropdown using `<vscode-button>` + `<vscode-context-menu>` from `@vscode-elements/elements`.

**Extension host side** (`calm-plugins/vscode/src/features/preview/`):
- `commands.ts` — new `ExportDiagramCmd` command receives the `exportDiagram` message from the webview.
- `preview-panel.ts` — `handleExportDiagram()` shows VSCode's native save dialog (pre-filled with `<arch-basename>-diagram-<n>.<ext>` in the same directory as the open file), then writes the SVG string or base64-decoded PNG buffer via `vscode.workspace.fs`.

**API changes:** none — communication uses the existing webview message bus.
**Data model changes:** none.
**Dependencies added:** `html-to-image` (webview rasterisation); `@vscode-elements/elements` is already used elsewhere in the extension.

### Alternatives Considered:
- **Separate SVG and PNG buttons**: Implemented first, replaced with the Export dropdown to keep the toolbar compact as more export formats could be added in future.
- **Reuse the CLI's browser-based renderer**: The CLI launches a Playwright browser to render Mermaid server-side. Pulling that into the extension host would add significant startup latency per diagram and require bundling Playwright; the webview already has a live, fully-rendered SVG, so capturing it client-side is lighter and instant.
- **Canvas-based PNG export**: Considered `drawSvgToCanvas` approaches but `html-to-image` handles off-screen layout and style computation more reliably across the range of SVG shapes Mermaid produces.

### Testing Strategy:
- **Unit tests** (`diagram-export.spec.ts`): cover `serializeSvgElement`, `rasterizeSvgElementToPng`, `exportDiagram`, and each of the three SVG fix helpers (`inlineResolvedFont`, `restoreDiagramViewBox`, `preventForeignObjectClipping`) with jsdom mocks.
- **Unit tests** (`diagram-controls.spec.ts`): verify the Export dropdown is rendered when callbacks are provided, is absent when neither callback is provided, and fires the correct callback on menu selection.
- **Unit tests** (`preview-panel.spec.ts`): verify `handleExportDiagram` calls `vscode.window.showSaveDialog`, writes the correct buffer, and shows the confirmation message.
- **Manual / E2E**: open a CALM architecture file in the Extension Development Host, switch to the Docify tab, use the Export dropdown, and verify the saved SVG opens correctly in a browser and the PNG is a valid image.

- Update `calm-plugins/vscode/README.md` to document the Export dropdown in the Docify preview panel.
- Update `calm-plugins/vscode/AGENTS.md` under the Webview Preview section to reference the new `diagram-export.ts` and `diagram-controls.ts` modules.

### Implementation Checklist:
- [ ] Documentation updated

### Additional Context:
This feature is the VSCode companion to the CLI `--export-diagrams` flag (merged in #2634). Both share the same underlying goal — high-fidelity diagram export — but serve different workflows: the CLI targets batch documentation pipelines, while this targets interactive single-diagram export from within the IDE.

YoofiTT96:
## Description

Adds SVG and PNG diagram export to the VSCode extension's Docify preview panel. An **Export** dropdown (built with `<vscode-button>` + `<vscode-context-menu>`) appears in the diagram toolbar alongside the existing zoom controls. Selecting a format triggers VSCode's native save dialog and writes the file to disk.

Closes #2674.

The VSCode companion to the CLI `--export-diagrams` flag (#2634): the CLI handles batch export during documentation generation; this handles interactive single-diagram export from within the IDE.

### Key implementation details

**Webview (`diagram-export.ts`)** — three fixes are applied to the live SVG before serialising, to ensure the exported file renders correctly as a standalone image:
- Restores Mermaid's original `viewBox` (stashed as `data-original-viewbox` before `svg-pan-zoom` strips it), and sets explicit `width`/`height`, so the file has correct intrinsic dimensions.
- Inlines the computed `font-family` as a presentation attribute on the root `<svg>`, so unstyled edge-label text does not fall back to the viewer's default serif font.
- Sets `overflow: visible` on all `<foreignObject>` elements, so labels are never clipped if a different renderer measures glyphs slightly wider.

For PNG, the same fixed clone is mounted in an off-screen `<div>` (not on the clone itself, to avoid `html-to-image` inlining `position: fixed` into the rasterised image) and rasterised at 2× pixel ratio.

**Extension host (`preview-panel.ts`)** — `handleExportDiagram()` opens a save dialog pre-filled with `<arch-basename>-diagram-<n>.<ext>` in the same directory as the open CALM file, then writes the SVG string or base64-decoded PNG buffer via `vscode.workspace.fs`.

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Commit Message Format ✅

Key commits follow conventional format:
- `feat(vscode): export Mermaid diagrams as SVG/PNG from Docify preview`
- `feat(vscode): replace SVG/PNG export buttons with an Export dropdown`
- `fix(shared): center exported diagrams and match export font to IDE rendering`

## Testing
- [x] All existing tests pass

New unit tests added:
- `diagram-export.spec.ts` — covers `serializeSvgElement`, `rasterizeSvgElementToPng`, `exportDiagram`, and all three SVG fix helpers with jsdom mocks.
- `diagram-controls.spec.ts` — verifies the Export dropdown renders correctly, is absent when no callbacks are provided, and dispatches the correct callback on selection.
- `preview-panel.spec.ts` — verifies `handleExportDiagram` opens the save dialog, writes the correct buffer, and shows the confirmation message.

## Checklist


LeighFinegold (commented):
I applied  a comment in wrong place

LeighFinegold (commented):
This looks great overall. Solid feature, good test coverage.

I'd be curious to hear what model/tooling you used and how you prompted it - and in particular what context you fed it about the existing architecture. Reason being: there are a few design choices that go against patterns already established in this extension, and I want to understand whether that's intentional divergence or something that slipped through.

The issues are all fixable and I'd be happy to pair on them. I've left inline comments on the specific files below.

calm-plugins/vscode/src/features/preview/preview-panel.ts
```diff
@@ -334,6 +336,30 @@ export class CalmPreviewPanel {
     this.viewModel.handleToggleLabels(showLabels)
   }
 
+  public async handleExportDiagram(format: 'svg' | 'png', data: string, diagramIndex: number): Promise<void> {
```
LeighFinegold, line 342: This method combines three responsibilities in one place:
1. Path/filename computation (business logic - `dirname`, `basename`, fallback to workspace root)
2. Save dialog interaction (VSCode UI - `showSaveDialog`)
3. File write + error handling (I/O - `workspace.fs.writeFile`)

The extension follows MVVM where ViewModels are framework-free (no `vscode` imports) and the panel is the View layer that only handles VSCode-specific I/O (see [`calm-plugins/vscode/AGENTS.md` - Key Design Principles](https://github.com/finos/architecture-as-code/blob/main/calm-plugins/vscode/AGENTS.md#key-design-principles)):

> 1. **Framework Isolation**: ViewModels have NO `vscode` imports
> 2. **Dependency Inversion**: Core depends on ports, not VSCode

The path computation logic should live in a ViewModel or a small `ExportService` - the panel just calls `showSaveDialog` and `writeFile` with whatever the ViewModel gives it.

Compare with how `handleRunDocify` (same file) delegates to `DocifyService` and emits results via the ViewModel event system. This new handler should follow the same shape.

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
```diff
@@ -0,0 +1,111 @@
...
+    return clone
+}
+
+/**
+ * Rasterizing the live SVG directly would reproduce the same clipped labels as the
+ * unfixed export (the live SVG is also missing Mermaid's original viewBox, since
+ * svg-pan-zoom strips it). Build the same fixed clone used for SVG export and render
+ * it off-screen so html-to-image can compute its layout and styles before rasterizing.
+ */
+export async function rasterizeSvgElementToPng(svg: SVGSVGElement): Promise<string> {
```
LeighFinegold: Two concerns with this dependency:

**1. CSP silent degradation:** The webview's Content Security Policy ([`media/preview.html`](https://github.com/finos/architecture-as-code/blob/main/calm-plugins/vscode/media/preview.html#L7)) sets `default-src 'none'` with no `connect-src` override, which means all fetch/XHR requests from within the webview are blocked. `html-to-image` uses `fetch()` internally to download font files so it can inline them into the exported image. Those fetches fail silently, so the library falls back to whatever system font is available.

Having watched the demo video, I think the fonts in the exported PNG are actually slightly different from what's rendered in the webview - but it's not noticeable unless you're looking for it. It'll get worse with custom fonts/themes though, and there's no warning when it happens.  

Let me know if this is not the case.

**2. Unnecessary dependency:** You already have the fixed SVG clone from `buildExportClone`. Can we rasterize using native browser APIs available in the webview?

calm-plugins/vscode/src/features/preview/webview/diagram-controls.ts
```diff
@@ -79,6 +91,61 @@ export class DiagramControls {
         return button;
     }
 
+    /**
+     * Create the "Export" button with an SVG/PNG dropdown menu.
+     * Returns null if neither export callback was provided.
+     */
+    private createExportControl(): HTMLElement | null {
```
LeighFinegold: `DiagramControls` was a focused single-responsibility class: zoom/pan UI controls ([original file before this PR](https://github.com/finos/architecture-as-code/blob/main/calm-plugins/vscode/src/features/preview/webview/diagram-controls.ts) - 110 lines, just buttons + PanZoomManager calls).

This PR adds:
- Web component imports (`@vscode-elements/elements/dist/vscode-button/index.js`, `@vscode-elements/elements/dist/vscode-context-menu/index.js`)
- A dropdown menu with show/hide toggle state
- Export callback wiring

Two issues:
1. **SRP:** Extract this to a separate `DiagramExportControl` class composed alongside `DiagramControls` in `docify-tab.view.ts`, not nested inside it.
2. **Consistency:** The existing zoom buttons are plain `<button class="diagram-control-btn">` elements. The export control uses `<vscode-button>` + `<vscode-context-menu>` web components - creating visual inconsistency. Also, the imports use private distribution paths (`/dist/vscode-button/index.js`) that aren't part of the package's public API and can break on minor bumps.

calm-plugins/vscode/src/features/preview/commands.ts
```diff
@@ -89,6 +91,13 @@ export class ToggleLabelsCmd implements WebviewCommand<{ type: 'toggleLabels'; s
   constructor(private p: PreviewCommandTarget) { }
   execute(m: { type: 'toggleLabels'; showLabels: boolean }) { this.p.handleToggleLabels(!!m.showLabels) }
 }
+export class ExportDiagramCmd implements WebviewCommand<{ type: 'exportDiagram'; format: 'svg' | 'png'; data: string; diagramIndex: number }> {
+  readonly type = 'exportDiagram' as const
+  constructor(private p: PreviewCommandTarget) { }
```
LeighFinegold, line 103: This is the first command in the registry that returns `Promise<void>` from `execute()`. But `CommandRegistry.dispatch` (line 43, same file) doesn't await it:

```ts
dispatch(msg: InMsg) { this.map.get(msg.type)?.execute(msg as any) }
```

If the async chain rejects outside the `try/catch` in `handleExportDiagram` (e.g. panel disposed mid-write, a race with `dispose()`), the promise rejection is unhandled.

Minor fix - make dispatch async-aware:
```ts
async dispatch(msg: InMsg) { await this.map.get(msg.type)?.execute(msg as any) }
```

Or add `.catch()` in the command itself. Not a blocker but worth hardening now that async commands are a thing.

YoofiTT96:
> This looks great overall. Solid feature, good test coverage.
> 
> I'd be curious to hear what model/tooling you used and how you prompted it - and in particular what context you fed it about the existing architecture. Reason being: there are a few design choices that go against patterns already established in this extension, and I want to understand whether that's intentional divergence or something that slipped through.
> 
> The issues are all fixable and I'd be happy to pair on them. I've left inline comments on the specific files below.

For tooling Sonnet 4.6 with High Effort. Likely slipped through on these design choices that are present in the AGENTs.md. Main context deliberately called out was the merge of the previous docify feature addition. Expected it to detect the CLAUDE.md which points to the AGENTs.md but seems that did not happen. The session pattern-matched against the nearest existing code — handleExportDiagram got bolted onto the same function shape as whatever wasalready in preview-panel.ts, and the export dropdown got added directly inside DiagramControls because that's where the zoom buttons already lived . Will go through again along with proposed fixes

Using the @vscode-elements/elements and html-to-image were deliberate. The former for the look and feel of the buttons to look more like native vs code options. It was previously rendering them as svg buttons much like the zoom controls. However this was more a cosmetic decision as the plain button elements were also suitable. The latter for PNG rasterisation but without checking the web views CSP.

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
YoofiTT96: Yes there was a very slight difference in node labels that I put down to rasterisation process + different formats. However, my attempts with using native browser apis + <canvas> pipeline failed consistently with:   
`SecurityError: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported.`

This is because drawing an SVG that contains a <foreignObject> onto a canvas via the browser's native `` decoder taints the canvas in Chromium even when the foreignObject content is 100% local/self-contained. Mermaid renders node/edge labels via `<foreignObject>` by default, so this fails on every diagram, not just ones with custom content.

`html-to-image` solves the problem by walking the DOM directly and inlining everything into their own representation before rasterizing, rather than handing the browser's native SVG decoder something it'll refuse to export. So the dependency is specific to mermaids rendering approach.

calm-plugins/vscode/src/features/preview/preview-panel.ts
YoofiTT96: Agreed. New change pulls the filename/path logic and the SVG-vs-PNG data decoding into a new `DiagramExportService`— a plain class with no vscode import, same shape as the existing ModelService. The panel now just calls `diagramExportService.computeDefaultPath(...)` to get a path, hands that to `showSaveDialog`, then calls `diagramExportService.decodeExportData(...)` before `workspace.fs.writeFile`

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
YoofiTT96: Performed some more tests and the taint issue can be circumvented by switching from setting the img source from a blob url to data uri built fromt he SVG string. Data uris are known to increase bloat but it is transient and only lives for the duration of the call. For a hypothetical very large architecture multi MB diagram it would add a one time decode cost. Flagging this possibility but would be fine for majority of use cases.

The rasterization pipeline now does:

  1. Serialize the already-fixed SVG clone to a string (no network access).
  2. Encode it as a data: URI via Blob + FileReader (no network access — this is local encoding, not a fetch).
  3. Load that into an Image, draw to <canvas>, read back via toDataURL() (no network access).

Also side steps the issue with possible future custom fonts and removes any issues with the CSP as no fetches are performed

calm-plugins/vscode/src/features/preview/webview/diagram-controls.ts
YoofiTT96: Fair point on mixing of concerns that should duly be extracted. 
For the latter I used vscode elements intiially for the look and feel but was prepared to revert based on feedback as its cosmetic. It also provided some accessibility controls with using the keyboard to navigate the menu options. For now those controls can be implemented without the dependency. If more controls and menu options come up in the future the dependency can be reintroduced

Reverted to regular buttons

Screenshot 2026-06-19 at 11 55 03 AM

YoofiTT96:
Summary of what changed, plus a couple of things the fixes themselves surfaced along the way.

**1. `handleExportDiagram` mixing path computation, save-dialog UI, and file I/O** — extracted a `DiagramExportService` (`core/services/diagram-export-service.ts`), framework-free like `ModelService`. It owns the default-path computation and SVG/PNG data decoding. `preview-panel.ts` now only calls `showSaveDialog`/`workspace.fs.writeFile` with whatever the service hands it.

**2. `html-to-image` + CSP-blocked font fetches** — removed the dependency entirely. PNG export now rasterizes the already-fixed SVG clone natively via `Image` + `<canvas>`, with no fetches involved at all. One thing this surfaced: my first attempt loaded the SVG via a `blob:` URL, which hits a real Chromium restriction — canvases are tainted when rasterizing an SVG containing `<foreignObject>` (which Mermaid uses for labels) via `blob:`, but never via `data:` URIs, in any browser, independent of content. Switched to encoding via `Blob` + `FileReader.readAsDataURL()` (base64, Unicode-safe) instead of `URL.createObjectURL()`, which resolved it cleanly.

**3. `DiagramControls` SRP violation, visual inconsistency, and private import paths** — extracted `DiagramExportControl` as its own class, composed alongside `DiagramControls` in `docify-tab.v

iew.ts`. Dropped `@vscode-elements/elements` entirely; the Export button and menu are now plain elements styled to match the existing `.diagram-control-btn` toolbar, so all four toolbar controls look consistent. This did mean losing the keyboard navigation `vscode-context-menu` provided for free, so I added it back by hand: `Escape` closes the menu and refocuses the trigger, arrow keys move between and wrap around menu items (and open the menu directly when pressed on the trigger), and focus leaving the control closes it. Flagging the hand implemented accessibility controls in case we want to change the button rendering to use a dependency that handles it

**4. `CommandRegistry.dispatch` not awaiting async commands** — `dispatch` now attaches a `.catch()` to whatever `execute()` returns, so a rejection (e.g. a disposed panel racing a write) cannot become an unhandled promise rejection.

All four areas have new or updated unit tests; full suite passing (489 tests), lint clean, build verified, and manually exercised in the Extension Development Host (export SVG/PNG, toolbar styling, keyboard nav on the Export menu).

https://github.com/user-attachments/assets/4c932192-2f29-434e-b669-1f593512ebf6

rocketstack-matt:
@YoofiTT96 are you able to resolve the merge conflict and mark any comments you've addressed as resolved please.

Is there anything you think still outstanding?

YoofiTT96:
> @YoofiTT96 are you able to resolve the merge conflict and mark any comments you've addressed as resolved please.
> 
> Is there anything you think still outstanding?

@rocketstack-matt Resolved merge conflicts. @LeighFinegold had one more review so outside of that nothing outstanding

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    const canvas = document.createElement('canvas')
+    canvas.width = width * pixelRatio
+    canvas.height = height * pixelRatio
+
+    const ctx = canvas.getContext('2d')
+    if (!ctx) {
+        throw new Error('Canvas 2D context unavailable')
+    }
+    ctx.drawImage(image, 0, 0, canvas.width, canvas.height)
```
LeighFinegold, line 98:
```suggestion
 const bgColor = getComputedStyle(document.body).getPropertyValue('--vscode-editor-background').trim() || '#ffffff'
ctx.fillStyle = bgColor
ctx.fillRect(0, 0, canvas.width, canvas.height)
ctx.drawImage(image, 0, 0, canvas.width, canvas.height)
```

Otherwise based on local test I was getting the following...

Image

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
LeighFinegold: Ok since you now removed I can mark this one as resolved

calm-plugins/vscode/src/features/preview/webview/diagram-controls.ts
LeighFinegold: Ok happy to resolve per update

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
LeighFinegold: Even though perhaps transparent isn't so bad, because the other box colours and fonts are taken from the theme I don't think it makes sense to not be consistent.

LeighFinegold:
> > @YoofiTT96 are you able to resolve the merge conflict and mark any comments you've addressed as resolved please.
> > Is there anything you think still outstanding?
> 
> @rocketstack-matt Resolved merge conflicts. @LeighFinegold had one more review so outside of that nothing outstanding

As indicated on Friday, based on the commit updates on Friday afternoon, I wouldn't have time to personally review until Sunday. I didn't want to blanket approve without trialing out the plugin myself. I have left an additional comment

https://github.com/finos/architecture-as-code/pull/2675#discussion_r3448975494

@YoofiTT96 - if you make that one change, then also bump up the vscode plugin version as that is still manual. Then can close and release a new version of the vscode extension

calm-plugins/vscode/src/features/preview/webview/diagram-export.ts
YoofiTT96: Looking into it. Had viewed them in vs code so didnt notice the background

YoofiTT96: Fixed

LeighFinegold (approved):

PR #2661, opened 2026-06-16
chore(calm-suite): A1 — remove vestigial config and rewrite file: deps to bare workspace

gjs-opsflo:
## Description

A1 of the 5-PR Phase 1 sequence under [Track A — #2649](https://github.com/finos/architecture-as-code/issues/2649) (parent epic [#2600](https://github.com/finos/architecture-as-code/issues/2600)). Removes vestigial config from `calm-suite/calm-studio/` and rewrites every deep `file:` workspace dep to bare semver `"*"`, then regenerates the root lockfile. No behaviour change.

The purpose of A1 is to **shrink the surface area before A3's large `git mv`**: npm 11 auto-symlinks `"*"` workspace deps regardless of path depth, so once these refs become path-agnostic, A3 can move `calm-suite/calm-studio/` → root without touching dependency declarations.

### What this PR does

**Deletions (8 paths)**
- `calm-suite/calm-studio/vercel.json` — no Vercel deploy target for this workspace.
- `calm-suite/calm-studio/package.json` — the `calmstudio-workspace` root manifest was a nested-workspace artifact; the repo-root `package.json` already declares `calm-suite/calm-studio/packages/*` and `calm-suite/calm-studio/apps/*` as workspaces.
- `calm-suite/calm-studio/packages/calmscript/` — DSL stub, not consumed by any other package.
- `calm-suite/calm-studio/apps/studio/.releaserc.json`
- `calm-suite/calm-studio/packages/calm-core/.releaserc.json`
- `calm-suite/calm-studio/packages/extensions/.releaserc.json`
- `calm-suite/calm-studio/packages/mcp-server/.releaserc.json`
- `calm-suite/calm-studio/packages/calmscript/.releaserc.json` (via dir delete)

**Rewrites (11 `file:` → bare `"*"` across 6 package.json files)**
- `packages/calm-core`: `@finos/calm-models`
- `packages/mcp-server`: `@calmstudio/calm-core` (deps + devDeps), `@finos/calm-models`
- `packages/github-action`: `@calmstudio/calm-core`, `@calmstudio/mcp`
- `packages/vscode-extension`: `@calmstudio/mcp`
- `packages/web-component`: `@calmstudio/calm-core`, `@calmstudio/extensions`
- `apps/studio`: `@calmstudio/calm-core`, `@calmstudio/extensions`

[...]

Issue #2651, opened 2026-06-15
[Track C] CALM Studio as a desktop editor over the platform


gjs-opsflo:
# [Track C] CALM Studio as a desktop editor over the platform

_Sub-task of #2600 — Track C. Depends on #2378 (`calm workspace`)._

## Goal

Reshape CALM Studio from a hosted web app with bundled content into a **downloadable desktop editor** over a local `calm workspace`, consuming shared platform capabilities and syncing to one or more CALM Hubs.

## Scope (C1–C9 from #2600)

- [ ] **C1.** Studio surface cleanup (no external deps):
  - Remove **C4-model level navigation** (`All / Context / Container / Component` tabs)
  - Remove **Demos** control (bundled demo files deleted in A2)
  - Remove **all references to calmscript** — editor tab + Export option (package deleted in A1)
- [ ] **C2.** Make Studio desktop-only:
  - Drop hosted web deployment (remove `@sveltejs/adapter-vercel`/`adapter-auto`, `vercel.json`, web-deploy workflow)
  - Retain static build as Tauri `frontendDist` only
  - **Studio is fully usable without MCP server** — MCP is optional, separately installed, never bundled
- [ ] **C3.** **[Linchpin]** Extract `calm workspace` core (manifest, resolver, bundle, `CalmHubClient`, push) out of `cli/` into `@finos/calm-shared` as an **importable library, no subprocess/CLI dependency**. CLI becomes one consumer. *(Depends on #2378.)*
- [ ] **C4.** Implement `pull` (currently TODO in #2378) + support **one or more remote Hubs per workspace** (e.g. pull reference patterns from FINOS public Hub; push/pull work to corporate Hub).
- [ ] **C5.** Wire Studio's **New / Open / Save / publish** onto shared `calm workspace` library, called **in-process via native Tauri file access — not by shelling out to CLI**:
  - Studio is **standalone**: can create/operate workspaces wholly on local filesystem with no Hub
  - Supports **multiple local workspaces**
  - When remote Hubs configured: "publish" = `workspace push`, "open from catalog" = `pull`
- [ ] **C6.** Replace bundled `_template` JSON with Hub-sourced patterns/architectures. Keep at most 1-2 generic local starters authored as real CALM patterns. Project-specific samples (FluxNova, OpenGRIS) published by their projects via Hub, not bundled.
- [ ] **C7.** Add Hub-initiated launchers in `calm-hub-ui`:
[...]

Issue #2650, opened 2026-06-15
[Track B] Capability convergence — validators, MCP, VS Code extension


gjs-opsflo:
# [Track B] Capability convergence — validators, MCP, VS Code extension

_Sub-task of #2600 — Track B. Supersets #2562 (MCP)._

## Goal

Converge platform-wide duplicated capabilities onto single implementations: one CALM type model + validator, one MCP server, one VS Code extension.

## Scope (B1–B4 from #2600)

- [ ] **B1.** `calm-guard` consumes in-repo `@finos/calm-models` + shared validation library **in-process** (not by shelling out to `@finos/calm-cli`). Retire bespoke Zod model.
- [ ] **B2.** **[Highest-risk]** Fold `@calmstudio/calm-core`'s Ajv + semantic rules into `@finos/calm-shared` — **additive**, preserving:
  - Orphan/dangling/self-loop checks
  - Spectral + document-position enrichment
  - No net loss of coverage (parity-gated tests required before removing `calm-core`'s engine)
- [ ] **B3.** **MCP consolidation per #2562 Option B** — see linked discussion issue for MCP-vs-skills decision:
  - Rename `@calmstudio/mcp` → `@finos/calm-mcp`
  - Add `hub_*` tools wrapping CALM Hub REST API
  - Deprecate then remove the Java MCP in CALM Hub (REST API unchanged)
  - *(The relocation #2562 proposes is subsumed by Track A's A3.)*
- [ ] **B4.** Remove the studio VS Code extension:
  - Preview superseded by FINOS `CALM Tools`
  - MCP bundling superseded by published `@finos/calm-mcp`
  - "Open in editor" capability replaced by Hub-initiated launchers (C7)

## Testing Strategy

- **B2 parity gate:** unified `shared` validator must reproduce every existing `calm-core` semantic check + retain Spectral + position enrichment before `calm-core`'s engine is removed
- **B3:** MCP tool-parity tests between old CALM Hub Java `@Tool` methods and new `hub_*` tools; CALM Hub REST API unchanged
- **B4:** test that FINOS extension registers `@finos/calm-mcp` without bundling it

## Why this matters

Removes:
- Three CALM type models / validators (B1, B2)
- Two non-interoperable MCP servers (B3)
- Two VS Code extensions, one bundling its own MCP — supply-chain liability for regulated users (B4)

## Out of scope

- Studio editor behaviour (Track C)
- Hub-initiated deep-link launchers (Track C — C7)
[...]

Issue #2649, opened 2026-06-15
[Track A] Structural cleanup of calm-suite (no behaviour change)


gjs-opsflo:
# [Track A] Structural cleanup of calm-suite (no behaviour change)

_Sub-task of #2600 — Track A._

## Goal

Land the lowest-risk structural cleanup of the `calm-suite/` contribution: remove vestigial config, hackathon artefacts, promote `calm-guard`/`calm-studio` to top-level workspaces, and unify tooling — **with no behaviour change**.

## Scope (A1–A5 from #2600)

- [ ] **A1.** Delete vestigial config: dead `calmstudio-workspace` package.json, `@calmstudio/calmscript` stub, per-package `.releaserc.json` files, and `vercel.json`. Replace deep relative `file:` deps with bare workspace resolution.
- [ ] **A2.** Remove hackathon artefacts:
  - `docs/docs/calm-studio/adrs/*`
  - `calm-suite/calm-studio/docs/{AIGF_CATALOGUE.json, CALM_1.2_CONTROLS_SCHEMA.md, REQ_fluxnova_aigf_integration.md}`
  - Bundled demo architectures: `examples/*`, `apps/studio/static/demos/*`
  - **Retain** genuine user docs under `docs/docs/calm-studio/` (folded into D1)
- [ ] **A3.** `git mv calm-guard` and `calm-studio` to top level. Update:
  - Root `workspaces` globs
  - CI workflow paths
  - `repository.directory` fields
  - Remove `calm-suite/`
- [ ] **A4.** Reconcile remaining identities to FINOS:
  - In-app "Report an issue" link → `finos/architecture-as-code`
  - Docs deploy URL
  - Remove `DEPLOYMENT-RECONCILIATION.md` once actioned
- [ ] **A5.** Unify tooling: single husky/commitlint/lint-staged at root; align vitest/eslint/vite; one coverage gate.

## Why this first

Per #2600 critical path: Track A → B1 → B2 → B3 → B4. A is no-behaviour-change and unblocks every downstream track by collapsing the directory layout and removing dead config that would otherwise compound conflicts in B and C.

## Out of scope

- Capability convergence (Track B)
- Studio behaviour changes (Track C)
- Docs site consolidation (Track D — D1 picks up A2-retained user docs)

## Acceptance criteria

- `calm-suite/` directory removed; `calm-guard/` and `calm-studio/` are top-level workspaces
- All workspaces resolve via bare names (no `file:../../../`)
[...]

PR #2683, opened 2026-06-18
fix(calm-suite): adopt canonical nested relationship-type in CALMGuard for cross-tool interop

eddie-knight:
## Description

**The journey this fixes:** You model an architecture in **CALM Studio**, export it (or grab any CALM document the **CLI** / **Hub** produces), then open **CALMGuard** to run compliance analysis on it. Today that fails — the document is **rejected the moment you upload it**, before any analysis runs. 

Concretely: take Studio's own `ecommerce` demo export and drop it into CALMGuard's upload zone → it's refused at the CALM-schema step.

The cause is a dialect mismatch. Canonical CALM 1.2 nests the relationship variant inside an object:

```json
"relationship-type": { "connects": { "source": { "node": "a" }, "destination": { "node": "b" } } }
```

…but CALMGuard hand-rolled its own Zod schema using a **flat string discriminant** with a sibling key (`"relationship-type": "connects"` + a sibling `"connects": {…}`). So CALMGuard was the only tool in the suite speaking a private dialect: it couldn't read what Studio/CLI/Hub/Visualizer produce, and the remediation PRs it wrote back were schema-invalid to those same tools. It also enforced a closed 9-value `node-type` enum, so Studio extension node-types (e.g. `aws:lambda`) were rejected too.

**After this PR**, the round trip works:

- ✅ A canonical CALM document from Studio / CLI / Hub uploads and analyzes in CALMGuard.
- ✅ CALMGuard's remediation PR output is canonical CALM that `calm validate`, the Hub, and the Visualizer accept.
- ✅ Extension `node-type` values from Studio packs are accepted.

### What changed
- Rewrote CALMGuard's relationship schema to the **canonical nested `relationship-type` object** (exactly-one-variant enforced via `superRefine`) and relaxed `node-type` to an open string, matching the CALM core spec.
- Added `getRelationshipVariant()` and routed every reader (CALM extractor, the React-Flow graph mapping, the learning extractor) through the nested shape.
- Made the v1.0 normalizer emit the nested form.
- Added a **compile-time conformance check** against `@finos/calm-models` types (gated by `tsc`, so drift fails the typecheck) plus interop tests that parse real canonical fixtures authored by other CALM tools and assert the legacy flat form is now rejected.
- Converted CALMGuard's demo examples and inline test fixtures to nested; updated `calm-guard/AGENTS.md`.

This PR also includes a small unrelated convenience commit, `chore(calm-suite): add Makefile to launch suite apps on demand` — happy to drop it if you'd prefer the PR scoped purely to the fix.

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Commit Message Format ✅
Conventional Commits, DCO signed-off:
- `fix(calm-suite): adopt canonical nested relationship-type in CALMGuard for cross-tool interop`
- `chore(calm-suite): add Makefile to launch suite apps on demand`

## Testing
- [x] All existing tests pass

Verified on Node 22: `npm run typecheck`, `npm run lint`, `npm run test:run` (117 tests, including new interop + conformance coverage), and `npm run build` for `calmguard` all pass. Manually confirmed against a real Studio export: pre-fix CALMGuard rejects it at the upload zone; post-fix it parses and proceeds to analysis.

## Checklist


rocketstack-matt:
Thanks @eddie-knight — this stack (#2683, #2691, #2692, #2693) is squarely in the direction of the convergence epic #2600; canonical CALM interop across the suite is exactly the end-state we're aiming for.

My one concern before merging is sequence. Each PR overlaps a track that's already scoped as an issue:

- **#2683 / #2693** modify CALMGuard's validator — Track B / B1 (#2650), which plans to *retire* the bespoke Zod model and validate **in-process**, not by shelling out to the CLI (which #2693 introduces).
- **#2691 / #2692** rework Studio's open/save/projection — Track C (#2651).
- All four sit under `calm-suite/**`, which Track A / A3 (#2649) will `git mv` to top level and remove `calm-suite/`.

The fixes are real, and some of the work is a genuine down-payment on the tracks (e.g. the `@finos/calm-models` conformance check in #2683). But addressing them out of sequence, with no link back to the issues, makes the work hard to see against the plan and risks duplicate or throwaway effort once the tracks start.

Two asks:

1. Please **reference the relevant track from each PR** — #2683/#2693 → #2650, #2691/#2692 → #2651, all → A3 in #2649 — so the overlap is visible.
2. @gjs-opsflo, you put the track breakdown together, so I'd like your call on sequencing:
   - **(a)** merge these now and fold the superseded parts (Zod model, CLI shell-out) into B1 when it starts; or
   - **(b)** hold the stack and address the issues in the requested order (A → B → C) for clarity.

   Either works for me — I'd just like us aligned before these merge.

gjs-opsflo (changes requested):
Strong PR — schema migration is clean and the compile-time `conformance.ts` gate vs `@finos/calm-models` is a great move. New `interop.test.ts` covering real cross-tool fixtures (`complex.architecture.json`, all 4 concrete variants) is the right anchor test. Normalizer leniency on input + strict canonical output is exactly the contract we need for round-trip with Studio and the CLI.

### Requesting one change before merge

**`calm-suite/calm-guard/package.json:45`** — `"@finos/calm-models": "file:../../calm-models"`

`calm-models` is already a root npm workspace, so this should use the workspace protocol so the package resolves via the workspace symlink rather than a relative file path:

```json
"@finos/calm-models": "*"
```

Why this matters:
- `file:` bypasses npm's workspace resolution and creates a hard relative-path coupling. If `calm-suite/` is ever restructured (or if calmguard is consumed outside the monorepo), this breaks in non-obvious ways.
- Renovate / dependabot treat `file:` deps differently from workspace deps.
- The CI workflow's explicit `Build @finos/calm-models` step (workflow lines added in this PR) is correct either way — the change is purely about how the dep is declared.

### Non-blocking observations (no action needed, just calling out)

1. **`node-type` widening is silent at the UI layer.** AGENTS.md is honest about "9 well-known values get first-class UI, any string accepted", but the React Flow custom-node mapping and the compliance heat-map are still keyed off the original 9-value enum, so an `aws:lambda` node will parse but fall through to a default renderer / bucket as "unknown" in the heat-map. The user story (parse, don't reject) is fixed here — render-side work is a reasonable follow-up issue.

2. **CI test job does not build `@finos/calm-models`.** Workflow adds the build step before lint and before build, but not before test. Currently safe because tests don't import calm-models at runtime (only `conformance.ts` does, type-only, esbuild strips). Worth a comment in `conformance.ts` so a future author who adds a runtime import knows to update the test job.

3. **`detectCalmVersion` returns `"1.1"` for canonical nested docs missing `adrs`/`decorators`/`timelines`.** Parses correctly, but the version label on `ParseSuccess.version` is technically wrong. No functional impact.

4. **Missing test scenario**: a relationship with TWO variant keys (e.g. both `connects` and `interacts`). The `superRefine` rejects this, which is the strictest-vs-canonical-`oneOf` behavior and worth pinning explicitly.

5. **Makefile is unrelated scope creep** (PR body acknowledges, offers to split). The Makefile itself is fine. No strong opinion either way.

Will re-approve once the `file:` → `"*"` flip lands. Everything else can be a follow-up.

package-lock.json
```diff
@@ -295,6 +295,7 @@
                 "@ai-sdk/xai": "^3.0.57",
                 "@dagrejs/dagre": "^2.0.4",
                 "@finos/calm-cli": "^1.33.0",
+                "@finos/calm-models": "file:../../calm-models",
```
eddie-knight, line 298:
```suggestion
                "@finos/calm-models": "*",
```

@gjs-opsflo

calm-suite/calm-guard/package.json
```diff
@@ -42,6 +42,7 @@
     "@ai-sdk/xai": "^3.0.57",
     "@dagrejs/dagre": "^2.0.4",
     "@finos/calm-cli": "^1.33.0",
+    "@finos/calm-models": "file:../../calm-models",
```
eddie-knight, line 45:
```suggestion
    "@finos/calm-models": "*",
```

@gjs-opsflo

gjs-opsflo:
👍 Both suggestions are exactly the change requested — please apply.

One heads-up on the `package-lock.json:298` edit: rather than hand-applying the suggestion, safer to let npm rewrite the resolution entry. After bumping `package.json` to `"*"`:

```bash
rm -rf node_modules package-lock.json && npm install
```

(or `npm install --package-lock-only` if you want to avoid the full reinstall). The `validate-lockfile` CI workflow checks all platform-specific optional deps are present in the lockfile, and a hand-edit risks drift if npm's resolution metadata for the workspace symlink differs from the literal `"*"` string. We've been bitten by lockfile regen edge cases before (PR #2661 broke `vsce` with `--force` regen), so the full delete-then-install path is the safest.

Once that lands I'll re-review and flip to APPROVE. Thanks for the quick turnaround!

eddie-knight:
Looks like #2730 got merged ahead of this one, so I'll need to revisit the entire PR stack.

PR #2693, opened 2026-06-19
fix(calm-suite): validate remediated CALM before opening a CALMGuard PR

eddie-knight:
## Description

**The journey this fixes:** You run a CALMGuard analysis, then click **"Create remediation PR"** to have CALMGuard open a pull request that applies its compliance fixes to your CALM document. Today, CALMGuard commits and opens that PR **without validating the remediated document first** — so if the remediation produces a schema-invalid CALM file, an invalid document lands in a PR against your repo, and downstream tools (the CLI validator, the Hub, anything consuming CALM) choke on it.

The most realistic way this happens: `applyChangesToCalm` writes an LLM-supplied control key (`change.after`) directly as a control-object key, with no check against the CALM schema's key pattern `^[a-zA-Z0-9-]+$`. An LLM emitting something like `"PCI DSS Req 8.4.2"` as a key yields a schema-invalid document — and nothing was stopping it from being committed.

**After this PR**, the remediated document is validated **before any GitHub write**:

- ✅ The remediated CALM is checked with the canonical CALM CLI (the same validator the `/api/calm/validate` route uses) right after the merge and **before** the branch/commit/PR.
- ✅ On failure the run aborts cleanly — **no branch, commit, PR, or label is created** (fail-closed) — and the user sees why in the streamed error.
- ✅ A validator that *fails to run* is distinguished from a genuinely invalid document (clear "could not validate" message), both fail-closed.

### What changed
- `src/app/api/github/create-pr/route.ts` (remediation branch): after `applyChangesToCalm` produces the remediated CALM, run `validateWithCalmCli`; on `{valid:false}` or a validator throw, `throw` → the route's existing SSE error/close path aborts before the first GitHub mutation (`createBranch`).
- Pin the route to the Node runtime (`export const runtime = 'nodejs'`) — the validator shells out via `child_process`, which is unavailable on the Edge runtime. Matches `/api/calm/validate`.
- Added a route-level test (`src/__tests__/api/create-pr.test.ts`): an invalid result never reaches `githubFetch` (no commit); a valid result proceeds past the gate.

### Scope notes
- **Remediation write-backs only.** The `pipeline`/`infra` write-backs commit YAML/IaC, not CALM, so they're untouched.
- This gates the **committed output**. The input document is Zod-checked at ingest (not CLI-checked); validating the artifact that actually lands in the PR is the meaningful guarantee.
- This is defense-in-depth against bad LLM control keys and any future regression in the merge — the merge only ever *adds* (never drops), so it can't produce dangling refs / dropped required fields.

> ℹ️ **Components note:** the change is in CALMGuard (`calm-suite/calm-guard/`), which has no checkbox in the template's "Affected Components", so none are ticked.

> 🔗 **Stacked PR:** targets `fix/calmguard-canonical-relationship-type` (the CALMGuard nested-relationship fix) rather than `main` — it depends on that PR so the remediated output is canonical and actually passes validation. Re-point to `main` once the parent merges. (Sibling to the Studio doc-keys PR, which also branches off it.)

## Type of Change
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] ♻️ Refactoring
- [ ] 🔧 Chore

## Affected Components
*(All unchecked — surface is CALMGuard (`calm-suite/calm-guard/`), which has no template checkbox.)*

## Commit Message Format ✅
Conventional Commits, DCO signed-off:
- `fix(calm-suite): validate remediated CALM before opening a CALMGuard PR`

## Testing
- [x] All existing tests pass

Verified on Node 22: `npm run typecheck --workspace=calmguard` (clean), `npm run lint --workspace=calmguard` (clean), `npm run test:run --workspace=calmguard` (**119 pass**, incl. 2 new), and `npm run build --workspace=calmguard`. The new route-level test mocks the CLI validator, the LLM remediation agent, and the GitHub client, and asserts that an invalid validation result aborts the stream and **never calls `githubFetch`** (no GitHub write), while a valid result proceeds past the gate.

## Checklist
- [x] Conventional commit format
- [x] Documentation updated if necessary
- [x] Tests added
- [x] Follows coding standards

### DIFF
--- a/calm-suite/calm-guard/src/__tests__/api/create-pr.test.ts
+++ b/calm-suite/calm-guard/src/__tests__/api/create-pr.test.ts
@@ -1,0 +1,143 @@
+/**
+ * API Route test: POST /api/github/create-pr — remediation validation gate.
+ *
+ * Verifies that a remediated CALM document is validated BEFORE any GitHub
+ * mutation: an invalid document (or a validator that fails to run) aborts the
+ * stream and never reaches githubFetch (no branch/commit/PR), while a valid
+ * document is allowed through the gate. The gate validates the MERGED output.
+ *
+ * All external dependencies (the CLI validator, the LLM remediation agent, and
+ * the GitHub client) are mocked, so no API keys / network are required.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { NextRequest } from 'next/server';
+
+// ---- Mocks (declared before the handler import) ----
+vi.mock('@/lib/calm/cli-validator', () => ({
+  validateWithCalmCli: vi.fn(),
+}));
+vi.mock('@/lib/github/client', () => ({
+  githubFetch: vi.fn(),
+}));
+// remediateCalm is the LLM agent — fully mocked so no provider/API key is touched.
+vi.mock('@/lib/agents/calm-remediator', () => ({
+  remediateCalm: vi.fn(),
+}));
+
+// ---- Import handler AFTER mocks ----
+import { POST } from '@/app/api/github/create-pr/route';
+import { validateWithCalmCli } from '@/lib/calm/cli-validator';
+import { githubFetch } from '@/lib/github/client';
+import { remediateCalm } from '@/lib/agents/calm-remediator';
+
+const REMEDIATION_BODY = {
+  type: 'remediation',
+  owner: 'o',
+  repo: 'r',
+  filePath: 'arch.json',
+  fileSha: 'abc',
+  defaultBranch: 'main',
+};
+
+function makeReq(body: unknown): NextRequest {
+  return new NextRequest('http://localhost/api/github/create-pr', {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+  });
+}
+
+async function readStream(res: Response): Promise<string> {
+  const reader = res.body!.getReader();
+  const decoder = new TextDecoder();
+  let out = '';
+  let done = false;
+  while (!done) {
+    const chunk = await reader.read();
+    done = chunk.done;
+    if (chunk.value) out += decoder.decode(chunk.value, { stream: !done });
+  }
+  return out;
+}
+
+describe('POST /api/github/create-pr — remediation validation gate', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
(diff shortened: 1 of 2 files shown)


eddie-knight:
Stacked on #2692, can be taken out of draft when that's merged

gjs-opsflo (approved):
Small, surgical, fail-closed. Gate placement is correct — sits between `applyChangesToCalm` (line 352) and the first GitHub mutation `getHeadSha` (line 377). No mutation precedes the gate, so an abort leaves no branch / commit / PR behind, exactly what the PR body claims.

### Strengths

- **Distinguished failure messages**: "Could not validate the remediated CALM..." for validator-crash vs "Remediated CALM failed validation..." for genuine schema rejection. Both throw → same SSE error path → `controller.close()` in `finally`. Operators can tell infra failure from schema rejection.
- **Validates the MERGED output**, not the input. Test 3 asserts the call receives `controls.mfa-control` (a key the input never had) — confirms the real `applyChangesToCalm` ran first.
- **No new runtime dependency.** `@finos/calm-cli` already in `package.json`, already used by `/api/calm/validate`.
- **Runtime pin justified** and matches precedent in `/api/calm/validate/route.ts`. Doesn't conflict with `maxDuration = 300`.
- **Validation errors surfaced to user** (sliced to first 5). Test 1 confirms a schema-violation string reaches the stream.
- **Surgically scoped**: pipeline + infra branches (which commit YAML/IaC, not CALM) untouched. Correct call — those branches don't need a CALM validator.

### Pre-existing concern worth a follow-up issue (NOT introduced by this PR)

`cli-validator.ts` uses `join(tmpdir(), 'calm-validate-${Date.now()}.json')` — two concurrent requests landing in the same millisecond race on `writeFile` / `unlink`. The gate's existence makes this code path hotter. Worth a follow-up issue using `crypto.randomUUID()` or `mkdtemp`.

### Minor test polish (non-blocking)

- `process.env.GITHUB_TOKEN = 'test-token'` set in `beforeEach`, never restored. Prefer `vi.stubEnv` to avoid side-effect leak across test files.
- Test 3 depends on real `remediation-merge` logic (not mocked) — assertion on `controls.mfa-control` relies on fallback at `remediation-merge.ts:225-229`. Acceptable as integration-y coverage; just drift-prone if merge refactors.
- Could assert that `{type:'error'}` is the *terminal* SSE event before stream close (currently relies on `readStream` not hanging).

LGTM.

PR #2722, opened 2026-06-22
fix(calm-hub-ui): render nested containers at 3+ levels (parent-before-child node ordering)

willosborne:
## Description

Fixes a bug in the CALM Hub UI architecture visualizer where **nested container nodes were silently dropped** at 3+ levels of `deployed-in` / `composed-of` containment (boxes-within-boxes).

**Root cause:** both transformers handed React Flow (v11) a node array that was not ordered parent-before-child. React Flow requires a parent node to appear before its children in the `nodes` array; when a nested container appeared before its parent container, React Flow could not resolve the parent and dropped the container. The transformers already computed correct positions/sizes for any depth (via `sortContainersDeepestFirst`), but the final array order was not depth-sorted. A 4-level stack listed out of order (`A, C, B, system`) rendered as `A > B > [leaf]` with `C` missing.

**Fix:**
- Add a pure helper `sortNodesParentsBeforeChildren()` to `layoutUtils.ts` that sorts nodes shallowest-first so every parent precedes its descendants (stable, non-mutating, cycle-safe), refactoring the depth walk into a shared `nodeDepth()` used by both sorts.
- Apply it at the single return point of `combineNodes()` (architecture path) and `applyPatternLayout()` (pattern path).

Only the final node-array ordering changes — no layout math (positions, sizes, padding) is touched, and there is no public API change. Verified that reordering does not affect stacking (z-index is set explicitly) or any consumer (all are id-keyed). The diff view inherits the fix for free via the `*WithDiff` wrappers.

## Type of Change

## Affected Components

## Testing
Added regression tests driving a 4-level out-of-order `deployed-in` stack through `parseCALMData` and `parsePatternData`, asserting parent-before-child for every link (genuine RED→GREEN). Helper unit tests cover ordering, non-mutation, stability, and cycle-safety.

- [x] All existing tests pass

> Note: the visualizer `utils/` suite passes 169/169 locally (run via vitest with the `node` environment). The full jsdom suite and eslint could not be run on this machine due to a pre-existing local `node_modules`/transitive-dependency issue unrelated to this change; CI validates lint + the full jsdom suite on Node 22.

## Checklist

### DIFF
--- a/calm-hub-ui/src/visualizer/components/reactflow/utils/calmTransformer.test.ts
+++ b/calm-hub-ui/src/visualizer/components/reactflow/utils/calmTransformer.test.ts
@@ -360,4 +360,35 @@ describe('parseCALMData', () => {
         expect(result.edges[0].target).toBe('node-1');
         expect(result.edges[1].target).toBe('node-2');
     });
+
+    describe('nested container ordering', () => {
+        // 4-level deployed-in stack A > B > C > system, with the nested containers
+        // listed out of parent-first order (A, C, B, system) — the case that drops C.
+        const data: CalmArchitectureSchema = {
+            nodes: [
+                { 'unique-id': 'A', name: 'A', 'node-type': 'system' },
+                { 'unique-id': 'C', name: 'C', 'node-type': 'system' },
+                { 'unique-id': 'B', name: 'B', 'node-type': 'system' },
+                { 'unique-id': 'system', name: 'system', 'node-type': 'service' },
+            ],
+            relationships: [
+                { 'unique-id': 'b-in-a', 'relationship-type': { 'deployed-in': { container: 'A', nodes: ['B'] } } },
+                { 'unique-id': 'c-in-b', 'relationship-type': { 'deployed-in': { container: 'B', nodes: ['C'] } } },
+                { 'unique-id': 's-in-c', 'relationship-type': { 'deployed-in': { container: 'C', nodes: ['system'] } } },
+            ],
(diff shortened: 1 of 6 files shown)


rocketstack-matt (approved):
Verified locally against a live CALM Hub (standalone/Nitrite): seeded a 4-level (`Region > Cluster > Pod > App`) and a 5-level `deployed-in` stack whose nodes are listed child-before-parent, then rendered the **same** data through `main` vs this branch on one backend.

- **main:** inner containers disappear — the 4-level stack renders `Region > Cluster > App` (Pod gone); the 5-level loses both intermediate Pods.
- **this branch:** every container renders, correctly nested, at both depths. No regression on normal single-level containment.

Non-blocking note on the mechanism, since it differs from the description: the lost container isn't actually *dropped* by React Flow. At runtime it's present in the DOM with correct nested geometry, `opacity:1`, and **no** "parent not found" warning. All group nodes share `z-index:-1`, so paint order follows array order and the later-painted parent **occludes** the earlier-painted child container. The parent-before-child ordering fixes the visual precisely *through* stacking among equal-z-index groups — i.e. the ordering is load-bearing, so the explicit z-index shouldn't later be assumed to cover stacking on its own.

Fix confirmed working. 👍

willosborne:
Nice, thanks @rocketstack-matt - will leave this up for a bit longer in case Aaman, Aidan, Paul or any other UI folks want to have a look - and then merge.

PR #2732, opened 2026-06-23
feat(calm-models): add control diff to diff engine

aamanrebello:
## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


gjs-opsflo (dismissed):
Scope correction: gjs-opsflo isn't a CODEOWNER for `calm-models/` (per `.github/CODEOWNERS`, my coverage is `/calm-guard/` + `/calm-studio/`; this path falls under `* @finos/architecture-as-code-maintainers`). Review ended up in my queue but the canonical owners for the diff engine are @rocketstack-matt / @aidanm3341 / @lbulanti-ms — calm-hub-ui's timeline is the primary consumer.

Treating this review as withdrawn/advisory. GitHub won't let me delete a submitted review, so a maintainer dismiss is the only way to clear the CHANGES_REQUESTED state.

Apologies for the noise.

calm-models/src/diff/diff.ts
```diff
@@ -221,7 +230,74 @@ export function diffAdrs(
...
+    return result;
+}
+
+function diffControlItem(controlItemA: CalmControlSchema, controlItemB: CalmControlSchema): ControlItemDiffResult {
+    const result = {
+        descriptionDiff: diffSentences(controlItemA.description, controlItemB.description).map((diff: ChangeObject<string>) => ({
+            content: diff.value,
+            changeType: getChangeTypeFromChangeObject(diff),
+        })),
+        requirementsDiff: diffArrays<CalmControlDetailSchema>(controlItemA.requirements, controlItemB.requirements).flatMap((diff: ChangeObject<CalmControlDetailSchema[]>) => diff.value.map((val: CalmControlDetailSchema) => ({
```
rocketstack-matt: `diffArrays` defaults to reference equality (`left === right`) for array elements. Two requirement objects with identical content but different object instances — the normal case, since they come from separately parsed JSON — are never `===`, so they'll always show as removed+added even when nothing changed. Confirmed by running this against the `diff` package directly. `diffArrays` accepts a `comparator` option; `valuesEqual` (already defined above in this file) does structural comparison.

The test suite doesn't catch this because `controlEditArchitecture` changes every requirement's content — it never covers "description changed, requirements untouched."
```suggestion
        requirementsDiff: diffArrays<CalmControlDetailSchema>(controlItemA.requirements, controlItemB.requirements, { comparator: valuesEqual }).flatMap((diff: ChangeObject<CalmControlDetailSchema[]>) => diff.value.map((val: CalmControlDetailSchema) => ({
```

calm-models/src/diff/diff.ts
```diff
@@ -221,7 +230,74 @@ export function diffAdrs(
...
+    Object.keys(controlsB).forEach((id: string) => {
+        if (controlsA[id] == null) {
+            result.controlItemsAdded[id] = controlsB[id];
+        }
+    });
+
+    const editedControlIds: string[] = [];
+
+    commonControlIds.forEach((id: string) => {
+        if (JSON.stringify(controlsA[id]) === JSON.stringify(controlsB[id])) {
```
rocketstack-matt: Raw `JSON.stringify` comparison bypasses this file's own `valuesEqual`/`normalizeValue` (which sorts keys before comparing) — the same helper `nodesModified`/`edgesModified` use for exactly this "are these two objects equal" check. Two controls that are semantically identical but serialize with keys in a different order will be spuriously reported as `controlItemsModified`.
```suggestion
        if (valuesEqual(controlsA[id], controlsB[id])) {
```

calm-models/src/diff/index.ts
```diff
@@ -24,6 +24,7 @@ export {
 export type {
     AdrDiffResult,
     ArchitectureDiffResult,
+    ControlDiffResult,
```
rocketstack-matt, line 28: `diffControls`/`diffControlItem` aren't re-exported here, unlike the sibling core diff functions `diffAdrs`/`diffNodesAndRelationships` above. Intentional, or should they follow the same export pattern?

calm-models/src/diff/diff.ts
```diff
@@ -221,7 +230,74 @@ export function diffAdrs(
...
         adrDiffItems: diffArrays(adrsA, adrsB).flatMap((diff: ChangeObject<string[]>) => diff.value.map((val: string) => ({
             content: val,
-            changeType: diff.added ? 'added' : diff.removed ? 'removed' : 'unchanged',
+            changeType: getChangeTypeFromChangeObject(diff),
         })))
     };
 }
+
+/**
+ * Core diff function for CALM control objects. Identifies which controls are added, removed and unchaged by ID, and identifies control changes under an individual ID.
```
rocketstack-matt: Nit (non-blocking): typo "unchaged" → "unchanged". Also, `ControlItemDetailsDiffItem`/`ControlDiffResult` (lines 65, 101) use trailing commas where the rest of the file (`StringDiffItem`, `NodesAndRelationshipsDiffResult`) uses semicolons.
```suggestion
 * Core diff function for CALM control objects. Identifies which controls are added, removed and unchanged by ID, and identifies control changes under an individual ID.
```

rocketstack-matt (changes requested):
Control diffing is well tested for the "content actually changed" cases, but two of the new equality checks diverge from this file's existing `valuesEqual` comparison helper and will misreport unchanged data as changed. Details inline.

calm-models/src/diff/index.ts
aamanrebello: `diffControls` should be exported but not `diffControlItem`, as that is simply a function meant to support `diffControls`.

aamanrebello: Actioned.

calm-models/src/diff/diff.ts
aamanrebello: Actioned as suggested.

calm-models/src/diff/diff.ts
aamanrebello: Actioned as suggested.

calm-models/src/diff/diff.ts
aamanrebello: Fixed.

calm-models/src/diff/diff.spec.ts
```diff
@@ -380,6 +380,159 @@ describe('diff', () => {
...
+            ]);
+
+            expect(result.controlItemsModified['id-1'].requirementsDiff).toEqual([
+                {
+                    changeType: 'unchanged',
+                    content: {
+                        'requirement-url': 'https://url',
+                        'config-url': 'https://config-url',
+                    },
+                },
```
rocketstack-matt, line 531: This covers the `diffArrays` comparator fix (structurally-identical requirements), but there's no equivalent regression test for the sibling fix — `valuesEqual` replacing raw `JSON.stringify` for `controlsA[id]`/`controlsB[id]`. Worth adding a case where a control's keys are reordered but content is identical, asserting it lands in `controlItemsUnchanged` rather than `controlItemsModified` — otherwise this is the same kind of gap that let the original bug ship undetected. (non-blocking nit)

rocketstack-matt (approved):
Both correctness bugs from the earlier review are fixed and verified against the current code (`valuesEqual`/comparator used consistently now); `npm test --workspace calm-models` passes (196/196) and lint/build are clean. Two non-blocking notes below.

calm-models/src/diff/diff.ts
rocketstack-matt: The typo's fixed, but the comma-vs-semicolon inconsistency is still there — `ControlItemDetailsDiffItem` (diff-types.ts:64-67) and `ControlDiffResult` (line 101) still use trailing commas while `StringDiffItem` and the rest of the file use semicolons. The suggested diff only touched the docstring line, not these. (non-blocking nit)

calm-models/src/diff/diff.spec.ts
aamanrebello: Ill make a new PR with these fixes.

PR #2760, opened 2026-06-29
feat(cli): add --fail-if-modified to hub push

willosborne:
## Description

Adds a strict, non-bumping `--fail-if-modified` mode to `calm hub push` for `architecture`, `pattern` and `standard` documents.

By default `hub push` auto-bumps: whenever a mapping already exists it computes the next version and always creates a new version (even when the content is identical). `--fail-if-modified` makes the behaviour explicit and safe for merge-time CI:

- **brand-new mapping** → created at `1.0.0` (unchanged behaviour)
- **exists + unchanged** → skipped (no new version; the file on disk is left exactly as-is; status reported as `Unchanged`)
- **exists + modified** → the push fails

The default (auto-bump) path is completely unchanged.

The content comparison is a new shared helper, `canonicalEqual` in `@finos/calm-shared` (`shared/src/hub/canonical.ts`), which compares two documents ignoring object key ordering (array order is preserved). It is deliberately introduced in `shared` so the same comparison can be reused by other push paths (e.g. the upcoming `calm workspace push --fail-if-modified`) without duplicating logic.

> Note: scoped to the mapped-resource push commands (architecture/pattern/standard) that share `pushDocument`. The domain-scoped `control-requirement` / `control-configuration` push flow is a sensible follow-up (it needs an equivalent "get latest published version" comparison) and is intentionally not included here to keep this PR focused.

## Type of Change

## Affected Components
- [x] CLI (`cli/`)
- [x] Shared (`shared/`)

## Testing
- [x] All existing tests pass

Verified on Node 26: `cli` and `shared` test suites pass with coverage gates held, and lint is clean for both. New tests cover the create / skip-unchanged / fail-on-modified paths in `hub-commands.spec.ts` and the shared `canonical.spec.ts`.

## Checklist

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -463,6 +463,9 @@ Example:
         .choices(['patch', 'minor', 'major'])
         .default('patch');
 
+    const hubFailIfModifiedOption = new Option('--fail-if-modified', 'Strict mode: do not auto-bump. Skip if the document is unchanged from the latest published version; fail if it has changed. A brand-new mapping is still created at 1.0.0.')
+        .default(false);
+
     const hubCmd = new Command('hub').description('Interact with CALM Hub');
 
     // hub push
@@ -476,6 +479,7 @@ Example:
         .option(CALMHUB_URL_OPTION, 'URL to CALMHub instance')
         .addOption(hubOutputOption)
         .addOption(hubVersionBumpOption)
+        .addOption(hubFailIfModifiedOption)
         .action(async (architectureFile, options) => {
             const pushOptions: PushOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
@@ -483,7 +487,8 @@ Example:
                 description: options.description,
                 file: architectureFile,
                 format: options.format,
-                changeType: options.changeType.toUpperCase() as ResourceChangeType
+                changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
             };
             await runPushArchitecture(pushOptions);
         });
@@ -496,6 +501,7 @@ Example:
         .option(CALMHUB_URL_OPTION, 'URL to CALMHub instance')
         .addOption(hubOutputOption)
         .addOption(hubVersionBumpOption)
+        .addOption(hubFailIfModifiedOption)
         .action(async (patternFile, options) => {
             const pushPatternOptions: PushOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
@@ -504,6 +510,7 @@ Example:
                 file: patternFile,
                 format: options.format,
                 changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
             };
             await runPushPattern(pushPatternOptions);
         });
@@ -516,6 +523,7 @@ Example:
         .option(CALMHUB_URL_OPTION, 'URL to CALMHub instance')
         .addOption(hubOutputOption)
         .addOption(hubVersionBumpOption)
+        .addOption(hubFailIfModifiedOption)
         .action(async (standardFile, options) => {
             const pushStandardOptions: PushOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
@@ -524,6 +532,7 @@ Example:
                 file: standardFile,
                 format: options.format,
                 changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
             };
             await runPushStandard(pushStandardOptions);
         });
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -543,12 +543,14 @@ Example:
         .option(CALMHUB_URL_OPTION, 'URL to CALMHub instance')
         .addOption(hubOutputOption)
         .addOption(hubVersionBumpOption)
+        .addOption(hubFailIfModifiedOption)
         .action(async (requirementFile, options) => {
             const pushOptions: PushControlOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
                 file: requirementFile,
                 format: options.format,
-                changeType: options.changeType.toUpperCase() as ResourceChangeType
+                changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
             };
             await runPushControlRequirement(pushOptions);
         });
@@ -559,12 +561,14 @@ Example:
         .option(CALMHUB_URL_OPTION, 'URL to CALMHub instance')
         .addOption(hubOutputOption)
         .addOption(hubVersionBumpOption)
+        .addOption(hubFailIfModifiedOption)
         .action(async (configFile, options) => {
             const pushOptions: PushControlOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
                 file: configFile,
                 format: options.format,
-                changeType: options.changeType.toUpperCase() as ResourceChangeType
+                changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
             };
             await runPushControlConfiguration(pushOptions);
         });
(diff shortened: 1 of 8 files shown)


cli/src/command-helpers/hub-commands.ts
```diff
@@ -197,6 +214,23 @@ export async function pushDocument(
         // Sort defensively so the highest version is last, regardless of the order Hub returns them in.
         const sortedVersions = sortSemVer(mappedResourceVersions);
         const latestVersion = sortedVersions[sortedVersions.length - 1];
+
+        if (options.failIfModified) {
+            // Strict mode: don't auto-bump. Compare the local document to the latest published
+            // version — skip when unchanged, fail when it differs.
+            const latest = await client.getMappedResourceByVersion(namespace, mapping, latestVersion, resourceType);
+            if (canonicalEqual(JSON.parse(fileContent), latest)) {
```
jpgough-ms: Two false-positive sources on this comparison:

1. **`$id` version is included.** `canonicalEqual` has no field exclusions — it compares full documents including `$id`, which encodes the version URL (e.g. `.../versions/1.0.0`). When another committer's normal push auto-bumps Hub to `1.0.1`, the local file still holds `$id: ".../versions/1.0.0"`. `canonicalEqual` returns false and the command throws "has changed" even though no architecture content changed. Strip or ignore `$id` (and any other version-metadata fields) before comparing.

2. **`description` normalization mismatch.** `updateDocumentMetadata` writes `"description": ""` when the field is absent or null (document-id-utils.ts:100: `json['description'] = newDocumentMetadata.description ?? ''`). The stored Hub document has gone through this normalization; `fileContent` here has not. Any document without a description field will always fail the comparison, making `--fail-if-modified` unusable for such documents.

cli/src/command-helpers/hub-commands.ts
```diff
@@ -197,6 +214,23 @@ export async function pushDocument(
         // Sort defensively so the highest version is last, regardless of the order Hub returns them in.
         const sortedVersions = sortSemVer(mappedResourceVersions);
         const latestVersion = sortedVersions[sortedVersions.length - 1];
+
+        if (options.failIfModified) {
+            // Strict mode: don't auto-bump. Compare the local document to the latest published
+            // version — skip when unchanged, fail when it differs.
+            const latest = await client.getMappedResourceByVersion(namespace, mapping, latestVersion, resourceType);
```
jpgough-ms, line 245: **No null guard on `latest`.** `getMappedResourceByVersion` returns `response.data as object` with no null check (calm-hub-client.ts). If Hub returns HTTP 200 with an empty body (proxy stripping, encoding issue), `latest` is `undefined`. `canonicalEqual(JSON.parse(fileContent), undefined)` evaluates `JSON.stringify(undefined)` which returns the JS value `undefined` — not a string — so the equality check becomes `"<json>" === undefined` → false, silently triggering the misleading "has changed" error instead of surfacing the real server problem. Add a null guard before the `canonicalEqual` call.

**Also: two HTTP calls where one would do.** `getMappedResourceVersions` + `getMappedResourceByVersion` = two round-trips. `getMappedResourceLatestVersion` (already used in `pullDocument` at line 336) returns the document directly in one GET and the embedded metadata contains the version. Replace the two-call sequence with a single `getMappedResourceLatestVersion` call.

cli/src/command-helpers/hub-commands.ts
```diff
@@ -197,6 +214,23 @@ export async function pushDocument(
...
+            if (canonicalEqual(JSON.parse(fileContent), latest)) {
+                newDocumentMetadata.version = latestVersion;
+                return { action: 'skipped', metadata: newDocumentMetadata };
+            }
+            throw new HubCommandError(
```
jpgough-ms, line 255: **`--change-type` is silently discarded when `--fail-if-modified` is set.** Both code paths in this block (return early at line 224, throw here at line 226) exit before reaching `computeSemVerBump(latestVersion, changeType)` at line 234. For the first-push path (`else` branch), version is hardcoded to `1.0.0` — `changeType` is never consulted there either.

A user who hits this error and re-runs without `--fail-if-modified` will publish a PATCH bump regardless of what `--change-type` they specified, with no warning. Either reject the combination with `conflictsWith('changeType')` on the option definition, or at minimum add the mutual exclusion to the error message: "Re-run without `--fail-if-modified` (and set `--change-type` explicitly) to publish a new version."

cli/src/cli.ts
```diff
@@ -476,14 +479,16 @@ Example:
...
         .action(async (architectureFile, options) => {
             const pushOptions: PushOptions = {
                 calmHubOptions: { calmHubUrl: options.calmHubUrl },
                 name: options.name,
                 description: options.description,
                 file: architectureFile,
                 format: options.format,
-                changeType: options.changeType.toUpperCase() as ResourceChangeType
+                changeType: options.changeType.toUpperCase() as ResourceChangeType,
+                failIfModified: options.failIfModified
```
jpgough-ms, line 491: **No Commander wiring test for `--fail-if-modified`.** The `hub-commands.spec.ts` tests inject `failIfModified: true` directly into function calls — they would pass even if this line were accidentally dropped. `cli.spec.ts` has equivalent wiring tests for `--change-type` (asserting `expect.objectContaining({ changeType: 'MINOR' })` against the mocked handler). Add a parallel test that parses actual argv `['hub', 'push', 'architecture', '--fail-if-modified', ...]` and asserts `expect.objectContaining({ failIfModified: true })`.

jpgough-ms (commented):
Off-diff notes:

**Duplicate canonicalize implementation** — `shared/src/hub/canonical.ts` introduces `canonicalize` + `canonicalEqual`, but `calm-models/src/diff/diff.ts` already has `normalizeValue` + `valuesEqual` (private) and the exported `canonicalKey(v)` which returns `JSON.stringify(normalizeValue(v))`. `shared` already depends on `@finos/calm-models`, so `canonicalEqual` could be `canonicalKey(a) === canonicalKey(b)` with no new file. Two copies means a bug fixed in one won't propagate to the other.

**`--fail-if-modified` missing from control push subcommands** — `push control-requirement` and `push control-configuration` use the same SemVer-bump versioning model but have no equivalent strict mode. Passing `--fail-if-modified` to either produces a Commander `unknown option` error with no explanation. Either add support (wiring `failIfModified` into `PushControlOptions` + `orchestrateControlPush`) or document the intentional scope gap.

jpgough-ms (changes requested):
Full disclosure I'm trying out a new code review process. Please feel free to dismiss any comments that don't look right - but a few of these would be good to fix.

cli/src/command-helpers/hub-commands.ts
willosborne: Good catch — both were real. Fixed in 9a2805a6: the local document is now normalised the same way Hub stores it (`updateDocumentMetadata(fileContent, { ...newDocumentMetadata, version: latestVersion })`) **before** comparing. That rewrites the `$id` to the latest version and applies the same `description: ""` default, so a version-only difference (someone else auto-bumped Hub) and an absent `description` no longer produce a false "has changed". Added a test asserting a version-only difference is treated as unchanged.

cli/src/command-helpers/hub-commands.ts
willosborne: Fixed in 9a2805a6. Comparison + null handling now go through a shared `isUnchangedFromPublished()` helper that throws a clear "Could not read the latest published version … (empty response)" error when the body is missing/empty, instead of letting `JSON.stringify(undefined)` silently read as "changed". Test added for the empty-body case.

On the two HTTP calls: I kept `getMappedResourceVersions` + `getMappedResourceByVersion`. The versions list is needed regardless for the new-mapping/`1.0.0` vs existing decision and returns `[]` cleanly on 404, whereas `getMappedResourceLatestVersion` throws on 404 so it does not distinguish "no mapping yet". So content still needs one fetch on top of the version list — happy to switch if you feel strongly.

cli/src/command-helpers/hub-commands.ts
willosborne: Addressed in 9a2805a6 by spelling out the mutual exclusion in the error message: "Re-run without --fail-if-modified (optionally with --change-type) to publish a new version." I deliberately avoided `conflicts(changeType)` because `--change-type` has a default value, so Commander would treat it as always-set and reject every `--fail-if-modified` invocation.

cli/src/cli.ts
willosborne: Added in 9a2805a6 — `cli.spec.ts` now parses real argv (`hub push architecture … --fail-if-modified`) and asserts `expect.objectContaining({ failIfModified: true })`, plus a default-false case and a `control-requirement` wiring test.

willosborne:
Thanks for the review — pushed 9a2805a6 addressing the inline comments (see thread replies). On the two off-diff notes:

**Duplicate canonicalize / reuse `canonicalKey` from calm-models** — keeping the dedicated `shared/src/hub/canonical.ts` for now. It's intentionally a tiny, dependency-light helper in the hub layer next to the other hub utilities (`document-id-utils`, `semver`), and is also consumed by the upcoming `calm workspace push` work, so I'd rather it not reach back into `calm-models`' diff internals (`canonicalKey` is a key-generation helper for diffing, a slightly different concern). Happy to revisit consolidation separately if you'd prefer a single implementation.

**`--fail-if-modified` missing from control push** — now implemented. `push control-requirement` and `push control-configuration` support the flag with the same normalise-and-compare model (`updateControlDocumentMetadata` to the latest version, then `canonicalEqual`), with skip / fail / create-1.0.0 behaviour and tests to match.

jpgough-ms (approved):

Issue #2754, opened 2026-06-28
CALM Hub UX redesign: replace the four-level Explore tree with a browse → detail flow


rocketstack-matt:
### Target Project:
`calm-hub-ui`, with one supporting endpoint in `calm-hub`.

### Description of Feature:
Open CALM Hub today and you meet an Explore tree four levels deep: namespace, then resource type, then item, then version. To find one architecture you expand a namespace, expand its Architectures, scan, and by then the other ten namespaces have dropped below the fold. Click an empty branch and a blank container opens with no hint whether it's empty, loading, or broken. On a phone the diagram spills off the right edge and the controls hide behind an unlabelled eye.

This redesign turns CALM Hub into a place you **browse**, the way you'd browse a code host. It rests on three moves:

- A slim, always-present **rail** down the left — every namespace and control domain on one level, each with a live count, filterable from the top.
- A **namespace page** where resource types become a row of counted tabs and the items show up as cards: name, description, type, how many versions. You read the catalogue at a glance.
- The **item view** you already know — Diagram, JSON, Deployments, the version timeline, the node drawer — tightened up and made consistent.

Underneath, the URL drives everything. Click a namespace and you go to `/namespace/:ns`; the rail highlight, the breadcrumb and the page all read from the same address, so they move together and the main panel always matches what you picked.

A full design package backs this — a recommended navigation structure, the reasoning behind each screen, an eight-piece component inventory, exact tokens, and high-fidelity desktop and mobile mockups. The key parts are below, with before/after shots from a working build.

### User Stories:
- **As an architect**, I want to see a namespace's artefacts as cards with their description, type and version count, so I can take in a catalogue at a glance instead of expanding tree nodes one by one.
- **As an architect**, I want a count on every rail row and resource tab, so an empty namespace reads as empty the moment I look at it.
- **As an architect on my phone**, I want the diagram to fit the screen on load with zoom controls right there, so I can actually use it on the train.
- **As someone opening the Hub for the first time**, I want a landing page that shows me what's in the catalogue and what's been touched recently, so I know where to start.
- **As anyone navigating**, I want the URL to carry my place, so a link I paste into a ticket lands a colleague on exactly what I was looking at.

### Current Limitations:
- The four-level tree fills the viewport from one namespace, so browsing across the catalogue means a lot of expand-collapse.
- Branches give no signal of what's inside — counts and empty states are missing, so empty, loading and broken all look the same.
[...]

PR #2761, opened 2026-06-30
feat(calm-hub-ui): browse rail, URL routing and counts endpoint

rocketstack-matt:
Part of #2754.

## Description

First of a **6-PR stacked series** redesigning the CALM Hub information architecture — replacing the four-level Explore tree (Namespace → Resource type → Item → Version) with a `browse → detail` flow. This is **tier 1 (the foundation)**; the remaining phases follow as stacked draft PRs, each raised ready-for-review as its predecessor merges.

This PR:
- **Browse rail** — a one-level, always-visible rail of namespaces and control domains, each with a live count and a "Filter namespaces" search. Replaces the desktop 4-level tree; the mobile iOS-style drill-down is kept and gains the same counts + a brand-tint active row.
- **URL-backed routing** — adds `/namespace/:ns` and `/domain/:domain` browse routes alongside the existing item-detail route, so the rail navigates to a route and the rail highlight / breadcrumb / page header can't drift out of sync.
- **Counts endpoint** — new `GET /calm/namespaces/counts` and `GET /calm/domains/counts` returning per-namespace per-type counts (+ a total) and per-domain control counts, so every rail row shows a live count without N per-type list fetches.
- **Removes the standalone Visualizer** (`/visualizer` page + navbar link + its e2e). The shared ReactFlow rendering used by the in-Hub diagram view is **kept**.

**Counts auth posture (for maintainer awareness):** the two counts endpoints are `@Authenticated` and apply a per-namespace `READ` / per-domain `DOMAIN_READ` filter internally, mirroring `SearchResource` — a caller only sees counts for namespaces/domains they can read (global-admin / no-auth / public-read see everything). Covered by `TestCountsServiceShould` and `TestNamespace/DomainResourceFilteringShould`.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

`calm-hub-ui` build + unit tests pass (cases at both desktop and mobile viewports); `calm-hub` Java tests pass (incl. the counts filtering tests, JaCoCo ≥90%). Manually exercised against a standalone/NitriteDB instance.

## Checklist


calm-hub-ui/src/hub/hooks/useResourceFromRoute.ts
```diff
@@ -0,0 +1,116 @@
...
+                            domain: namespace,
+                            controlId: match.id,
+                            controlName: match.name,
+                            controlDescription: match.description,
+                        });
+                    }
+                })
+                .catch(() => undefined);
+            return;
+        }
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in e9faa65c — the control loader now matches by numeric id **or** name slug and propagates `controlTitle`, so search-originated `/controls/<name>/detail` links load again and the detail header shows the title.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
```diff
@@ -192,7 +141,7 @@ export function MobileNavMenu({ onDataLoad, onAdrLoad, onControlLoad, onInterfac
             controlService
                 .fetchControlsForDomain(domain)
                 .then((controls: ControlDetail[]) =>
-                    setLeafItems(controls.map((c) => ({ id: c.name, name: c.title ?? c.name })))
+                    setLeafItems(controls.map((c) => ({ id: c.id.toString(), name: c.name })))
                 )
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in e9faa65c — the mobile drill-down now labels controls with `title ?? name`. I kept id-based navigation rather than switching to the name slug: ids are unique, and now that the loader matches both id and name, the mobile (id) and search (name) deep-links both resolve.

calm-hub/src/test/java/org/finos/calm/services/TestCountsServiceShould.java
```diff
@@ -0,0 +1,270 @@
...
+    ControlStore mockControlStore;
+
+    CountsService service;
+
+    @BeforeEach
+    void setUp() {
+        service = new CountsService(mockNamespaceStore, mockDomainStore, mockArchitectureStore,
+                mockPatternStore, mockFlowStore, mockStandardStore, mockAdrStore,
+                mockInterfaceStore, mockControlStore);
+    }
```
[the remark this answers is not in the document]
rocketstack-matt: Left as-is. These tests stub every store they exercise and CI is green; `sizeOrZero`/`controlCount` only swallow the typed `NamespaceNotFoundException`/`DomainNotFoundException`, and the real stores return empty lists rather than null, so there is no NPE path in practice. Can add lenient empty-list defaults later if the fixtures grow.

calm-hub-ui/src/hub/hooks/useResourceFromRoute.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+        renderAt('/security/controls/5/detail');
+        await waitFor(() => {
+            expect(callbacks.onControlLoad).toHaveBeenCalledWith({
+                domain: 'security',
+                controlId: 5,
+                controlName: 'Encryption',
+                controlDescription: 'desc',
+            });
+        });
+    });
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in e9faa65c — added a test covering a control route whose id segment is a name slug, asserting the match and `controlTitle` propagation.

calm-hub-ui/src/hub/components/domain-page/DomainPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+                                <button
+                                    className="w-full text-left px-3 py-2 rounded-[7px] text-[14px] hover:bg-base-200"
+                                    style={{ color: colors.redesign.bodyStrong }}
+                                    onClick={() =>
+                                        onControlLoad({
+                                            domain,
+                                            controlId: control.id,
+                                            controlName: control.name,
+                                            controlDescription: control.description,
+                                        })
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in e9faa65c — the click handler now passes `controlTitle: control.title`.

calm-hub-ui/src/hub/components/domain-page/DomainPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+                                    }
+                                >
+                                    {control.name}
+                                </button>
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in e9faa65c — the list now renders `control.title ?? control.name`.

calm-hub-ui/src/hub/components/namespace-page/useNamespaceItems.ts
```diff
@@ -0,0 +1,74 @@
...
+    name: string;
+}
+
+export interface NamespaceItemGroup {
+    type: TypeInUI;
+    items: NamespaceItem[];
+}
+
+const summaryToItem = (s: ResourceSummary): NamespaceItem => ({
+    id: s.customId ?? s.id.toString(),
```
jpgough-ms, line 19: `summaryToItem` calls `s.id.toString()` when `s.customId` is nullish. If `s.id` is also null, the TypeError thrown inside the `.then()` callback has no outer `.catch()` or `finally` to catch it — `setLoading(false)` is never reached and the spinner hangs permanently. Add `.finally(() => setLoading(false))` on the outer Promise.all, or wrap with a `.catch()`.

calm-hub/src/main/java/org/finos/calm/services/CountsService.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    /**
+     * Sizes a store list, treating a missing namespace as an empty result. The
+     * namespace came from the store's own namespace list, so a not-found here
+     * means the resource type simply has nothing for it — count it as zero
+     * rather than failing the whole aggregate.
+     */
+    private int sizeOrZero(NamespaceListSupplier supplier) {
```
jpgough-ms, line 161: `sizeOrZero` catches only `NamespaceNotFoundException`. Any unchecked `RuntimeException` from the store (e.g. `NitriteIOException` on a locked DB file) escapes the try-catch, propagates through the `stream().map()` lambda in `getNamespaceCounts`, and turns the entire counts endpoint into a 500 — no partial results returned. Broaden the catch to `Exception` or `RuntimeException`.

calm-hub-ui/src/hub/Hub.tsx
```diff
@@ -39,41 +68,67 @@ export default function Hub() {
...
+    // Navigating to a non-detail route clears any loaded resource so namespace /
+    // domain pages (and the empty landing) don't show a stale detail panel. Keyed
+    // on react-router's location.key (which changes on every navigation, including
+    // re-selecting the already-active rail row) rather than pathname: an in-place
+    // control / interface load does not navigate, so its key is unchanged and the
+    // detail is preserved, while any navigation clears stale detail state. Runs in
+    // a layout effect so the clear happens before paint, avoiding a one-frame flash
+    // of the stale panel.
+    useLayoutEffect(() => {
+        if (!isDetailRoute) {
```
jpgough-ms: When navigating from a domain page where `controlData` was set in-place, the `!isDetailRoute` guard here prevents clearing it. `detailContent` at line 154 evaluates `controlData` first, so the old control briefly renders as the new resource until the async fetch resolves. Either clear unconditionally on `locationKey` change, or null-guard `controlData` against the current route before using it in `detailContent`.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * namespaces (with total counts) and control domains (with control counts),
+ * always visible. Rows navigate to routes; the active row is derived from the
+ * URL, so the rail highlight can never drift from the content pane.
+ *
+ * Counts are owned by {@link Hub} (the expensive counts endpoint is fetched
+ * once there and shared), so this component takes them as props rather than
+ * re-fetching them itself.
+ */
+export function ExploreRail({ namespaceCounts, domainCounts, onCollapse }: ExploreRailProps) {
+    const { ns: activeNamespace, domain: activeDomain } = useParams<RailRouteParams>();
```
jpgough-ms: `useParams` is destructured for key `ns` (matches `/namespace/:ns`). The detail route `/:namespace/:type/:id/:version` uses param key `namespace` — `activeNamespace` is always `undefined` there and the active rail row loses its highlight for the entire resource-detail session. Fix: `const { ns, namespace } = useParams<{ ns?: string; namespace?: string }>()` and use `ns ?? namespace` as the active key.

calm-hub-ui/src/hub/components/namespace-page/NamespaceResourceGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+            >
+                {type}
+            </h2>
+            <ul className="flex flex-col gap-1">
+                {items.map((item) => (
+                    <li key={`${type}-${item.id}`}>
+                        <button
+                            className="w-full text-left px-3 py-2 rounded-[7px] text-[14px] hover:bg-base-200"
+                            style={{ color: colors.redesign.bodyStrong }}
+                            onClick={() => openItem(item.id)}
```
jpgough-ms, line 58: `openItem` is called fire-and-forget — no `await`, no `.catch()`. If `fetchVersionsForResource` rejects (API 500, timeout) there is an unhandled promise rejection with no user feedback. If `resolveResourceDetailPath` returns `null` (item has no published versions), navigation is silently skipped. Chain `.catch()` and surface a visible error or disable the item when no path is available.

calm-hub-ui/src/hub/Hub.tsx
```diff
@@ -110,6 +161,20 @@ export default function Hub() {
...
+    // over the route-driven page so its detail view shows.
+    const content =
+        isDetailRoute || controlData || interfaceData || adrData || data ? (
+            detailContent
+        ) : activeNamespace ? (
+            <NamespacePage namespace={activeNamespace} total={namespaceTotal} />
+        ) : activeDomain ? (
+            <DomainPage domain={activeDomain} controlCount={domainControlCount} onControlLoad={handleControlLoad} />
+        ) : (
+            detailContent
```
jpgough-ms: On initial load at `/` all guards miss (`isDetailRoute` false, `controlData`/`interfaceData`/`adrData`/`data` all falsy, `activeNamespace`/`activeDomain` both undefined) and this `else` arm renders `detailContent` → `<DocumentDetailSection data={undefined} />`. Currently harmless, but any null-guard added to `DocumentDetailSection` will crash the landing page. Add a dedicated landing-page arm before this fallback.

calm-hub/src/main/java/org/finos/calm/services/CountsService.java
```diff
@@ -0,0 +1,143 @@
...
+        return domainStore.getDomains().stream()
+                .filter(domain -> readableDomains.map(set -> set.contains(domain)).orElse(true))
+                .map(domain -> new DomainControlCount(domain, controlCount(domain)))
+                .collect(Collectors.toList());
+    }
+
+    private NamespaceCounts countsForNamespace(String namespace) {
+        return new NamespaceCounts(
+                namespace,
+                sizeOrZero(() -> architectureStore.getArchitecturesForNamespace(namespace)),
```
jpgough-ms, line 132: `countsForNamespace` issues 6 full-collection list fetches per namespace just to call `.size()` — 6N uncached DB reads on every browse-rail request. The v1 tradeoff is noted at lines 76-80, but this will become a production scalability issue under real load. Before this ships, either add count-only methods to each store or cache the result with a short TTL on this `@ApplicationScoped` bean.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
```diff
@@ -1,41 +1,42 @@
...
 
 const RESOURCE_TYPES: TypeInUI[] = ['Architectures', 'Patterns', 'Flows', 'Standards', 'ADRs', 'Interfaces'];
 
 interface MobileNavMenuProps {
-    onDataLoad: (data: Data) => void;
-    onAdrLoad: (adr: Adr) => void;
-    onControlLoad: (control: ControlData) => void;
-    onInterfaceLoad: (iface: InterfaceData) => void;
+    /** Per-namespace counts, fetched once by {@link Hub} and passed down. */
+    namespaceCounts: NamespaceCounts[];
```
jpgough-ms, line 26: `namespaceCounts` is passed in from Hub (which already fetched it), but the `useEffect` at line 84 still independently calls `fetchNamespaces()` and `fetchDomains()` on every mount — 2 redundant HTTP requests. Replace the `namespaces` state with `namespaceCounts.map(c => c.namespace)` and `domains` with `domainCounts.map(c => c.domain)`. This also removes the potential snapshot divergence between the count badges and the row labels.

calm-hub/src/main/java/org/finos/calm/resources/NamespaceResource.java
```diff
@@ -48,6 +65,37 @@ public ValueWrapper<NamespaceInfo> namespaces() {
...
+    }
+
+    /**
+     * Resolves the caller's readable namespaces the same way {@code SearchResource} does:
+     * {@link Optional#empty()} (no filtering, see everything) when auth is disabled or the
+     * {@link UserAccessValidator} is not resolvable (no-auth / standalone profile);
+     * otherwise the validator's READ-sufficient set (which is itself empty for global-admin
+     * and public-read, meaning everything).
+     */
+    private Optional<Set<String>> resolveReadableNamespaces() {
```
jpgough-ms, line 85: `resolveReadableNamespaces()` is a verbatim copy of the same private method in `SearchResource`. `DomainResource.resolveReadableDomains()` is a third copy differing only in the `getReadable*` call. Any future change to the auth-bypass logic must be applied in all three places. Extract to a shared helper on `CalmHubPermissionChecker` or a package-private utility.

jpgough-ms (changes requested):
10 findings across correctness, efficiency, and conventions.

**Off-diff: `calm-hub-ui/src/App.tsx`** — `/visualizer` route and `Visualizer.tsx` are removed with no redirect. Users with bookmarks or links to `/visualizer` get a blank page. Add `<Route path="/visualizer" element={<Navigate to="/" replace />} />` if intentionally removed.

calm-hub-ui/src/hub/components/namespace-page/useNamespaceItems.ts
rocketstack-matt: Fixed in dc11759f — moved `setLoading(false)` into a `.finally` (guarded by `cancelled`), so a mapping error can no longer leave the spinner hanging.

calm-hub/src/main/java/org/finos/calm/services/CountsService.java
rocketstack-matt: Fixed in 0bcb5908 — `sizeOrZero` and `controlCount` now also catch `RuntimeException`, log it, and count that store as 0, so one failing store no longer 500s the whole endpoint.

calm-hub-ui/src/hub/Hub.tsx
rocketstack-matt: Fixed in dc11759f — the clear now runs on every `locationKey` change (dropped the `!isDetailRoute` guard), so navigating to a detail route clears stale in-place `controlData` before the new fetch resolves. In-place loads don't change `locationKey`, so they're still preserved.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
rocketstack-matt: Fixed in dc11759f — now reads both `ns` and `namespace` and uses `ns ?? namespace`, so the rail highlight persists on the `/:namespace/:type/:id/:version` detail route.

calm-hub-ui/src/hub/components/namespace-page/NamespaceResourceGroup.tsx
rocketstack-matt: Fixed in dc11759f — `openItem` now catches rejections and logs when `resolveResourceDetailPath` returns null, instead of failing silently. Kept to logging for now; happy to add a visible toast as a follow-up if you'd prefer.

calm-hub-ui/src/hub/Hub.tsx
rocketstack-matt: Fixed in dc11759f — added a dedicated landing arm so the empty `/` state no longer renders `DocumentDetailSection` with undefined data.

calm-hub/src/main/java/org/finos/calm/services/CountsService.java
rocketstack-matt: Added a short-TTL (5s) per-namespace/per-domain cache in 0bcb5908, so a burst of browse-rail requests reuses the aggregate instead of re-reading ~6 lists per namespace each time. The read-filter still runs first, so unreadable scopes are never fetched or cached. A store-level count query stays the longer-term optimisation.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
rocketstack-matt: Fixed in dc11759f — the drill-down now derives namespaces/domains from the `namespaceCounts`/`domainCounts` props Hub already fetched, removing the two redundant requests and the snapshot divergence.

calm-hub/src/main/java/org/finos/calm/resources/NamespaceResource.java
rocketstack-matt: Fixed in 0bcb5908 — extracted the auth-bypass resolution into a shared `ReadableScope.resolve`; Namespace/Domain/SearchResource now all delegate to it, so the bypass rule lives in one place.

rocketstack-matt:
Also addressed the off-diff `/visualizer` note — added a redirect from `/visualizer` to `/` in `App.tsx` (dc11759f) so old bookmarks land on the hub rather than a blank page.

All ten points are in dc11759f (calm-hub-ui) and 0bcb5908 (calm-hub). Full `calm-hub-ui` suite (974) and `calm-hub` `verify` both green (coverage gate met).

jpgough-ms (approved):

PR #2762, opened 2026-06-30
feat(calm-hub-ui): namespace card-browse page with segmented type tabs

rocketstack-matt:
> 🚧 **Draft — stacked on #2761 (phase 1 of the CALM Hub redesign).** Ready for review once #2761 is merged; I'll then rebase this onto `main` so it shows only its own changes. Until then GitHub shows the **cumulative** diff (this phase plus the not-yet-merged prior one).

Part of #2754.

## Description

Phase 2 of the CALM Hub redesign — fills the namespace browse page (the rail's destination) with the card-browse UI. Fixes the "4-level tree doesn't scale / no metadata scent" and "empty types look broken" problems.

- **`SegmentedTypeTabs`** — resource-type tabs with count badges: active (blue underline), resting, and **dimmed-but-visible zero-count** (so "empty" reads differently from "broken"). The mobile drill-down's resource-type level gains the same per-type counts.
- **`ItemCard` grid** — a 3-up card grid: type-tinted striped thumbnail, name, 2-line-clamped description, a `TypeBadge`, and the mono "N versions" scent. Whole card links to the item detail.
- **`EmptyState`** — shown in the grid when a type has zero items.
- **`versionCount` (backend)** — added to the architecture/pattern/flow/standard namespace summaries, computed O(1) from the already-loaded versions map (no extra query), so cards can show the version count.

**Deferred (noted for reviewers):** `ItemCard` is a `<button>`, so open-in-new-tab / middle-click aren't supported yet — a version-less link route was deferred to avoid reopening the phase-1 routing contract.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

## Checklist


jpgough-ms (commented):
Automated review — 2 correctness issues, plus systemic duplication across the parallel Architecture/Flow/Pattern/Standard code paths (see inline comments).

**Off-diff notes:**
- `useNamespaceItems.ts:51` — `Promise.all` fetches all 6 resource types on every namespace mount (pre-existing code, unchanged by this PR), but the new segmented-tabs UI only ever renders one type at a time (`NamespacePage.tsx:126`). Tab-count badges come from a separate, cheap endpoint, so this isn't load-bearing — 5/6 of the fetched data is discarded on first load.
- Only `MongoStandardIntegration.java`/`NitriteStandardIntegration.java` assert `versionCount` end-to-end in the HTTP response body. The equivalent Architecture/Pattern/Flow integration tests don't, despite `versionCount` being populated identically for all four types — a regression that drops it from just those three responses would pass CI.

calm-hub-ui/src/hub/Hub.tsx
```diff
@@ -52,7 +53,14 @@ export default function Hub() {
     // never re-fires and there is no in-flight fetch to cancel on a dependency change. (Unlike
     // useNamespaceItems, whose fetch effect re-runs per namespace and so needs a cancel guard.)
     useEffect(() => {
-        countsService.fetchNamespaceCounts().then(setNamespaceCounts).catch(() => setNamespaceCounts([]));
+        countsService
+            .fetchNamespaceCounts()
+            .then(setNamespaceCounts)
+            .catch(() => setNamespaceCounts([]))
```
jpgough-ms: Counts fetch and item-grid fetch are independent, uncoordinated data sources. If `fetchNamespaceCounts()` rejects here (silently caught), `namespaceCountsLoaded` still flips true on line 63, so the active tab renders a '0' badge while `useNamespaceItems`' separate fetch still populates the grid with real items below it — a visible contradiction with no surfaced error.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
```diff
@@ -91,8 +91,11 @@ public List<NamespaceArchitectureSummary> getArchitecturesForNamespace(String na
             String description = architectureDoc.getString("description");
             if (name == null) name = "Architecture " + archId;
             if (description == null) description = "";
+            // Count versions from the already-in-memory sub-document (O(1), no extra query).
+            Document versions = (Document) architectureDoc.get("versions");
```
jpgough-ms: Unguarded cast: `(Document) architectureDoc.get("versions")`. No schema validation constrains this field's type. A non-Document value (legacy/corrupted data) throws `ClassCastException` here, which propagates unmapped and 500s the entire `GET /namespaces/{ns}/architectures` response — not just the offending record. Same pattern in `MongoFlowStore.java:82`, `MongoPatternStore.java:82`, `MongoStandardStore.java:71`, and identically in all 4 Nitrite stores (`NitriteDocument.get(String, Class)` also throws on mismatch, so this isn't Mongo-specific).

calm-hub/src/main/java/org/finos/calm/domain/namespaces/NamespaceResourceSummary.java
```diff
@@ -9,11 +9,13 @@ public class NamespaceArchitectureSummary {
     private String name;
     private String description;
     private Integer id;
+    private int versionCount;
```
jpgough-ms, line 12: This class is byte-for-byte identical to `NamespaceFlowSummary`, `NamespacePatternSummary`, and `NamespaceStandardSummary` bar the name. The `versionCount` computation added here is also copy-pasted into all 8 Mongo/Nitrite store implementations. `org.finos.calm.store.util.VersionKeySelector` already exists for this kind of versions-key logic (used by `MongoControlStore`/`NitriteControlStore`) — extend it instead of hand-duplicating the null-guard-and-count snippet 8 times.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
```diff
@@ -93,6 +93,31 @@ export function MobileNavMenu({ namespaceCounts, domainCounts, onClose }: Mobile
         (d: string) => domainCounts.find((c) => c.domain === d)?.controlCount,
         [domainCounts]
     );
+    // Per-type count for a namespace's resource-type drill-down level, mirroring
+    // the desktop namespace page's segmented type tabs.
+    const typeCount = useCallback(
```
jpgough-ms, line 100: `typeCount`'s switch re-implements `NamespacePage.tsx`'s `COUNT_FIELD: Record<NamespaceResourceType, keyof NamespaceCounts>` (lines 33-40 there) by hand. The `Record` fails to compile if a type is missing; this switch's `default: return undefined` doesn't — it'll silently drop mobile counts for any future resource type with no build error. Export `COUNT_FIELD` from `resource-type-meta.ts` and use it here instead.

calm-hub-ui/src/hub/components/namespace-page/resource-type-meta.ts
```diff
@@ -0,0 +1,69 @@
+import { colors } from '../../../theme/colors.js';
+
+/** The six resource types shown on the namespace browse page, in tab order. */
+export type NamespaceResourceType =
```
jpgough-ms: `NamespaceResourceType` hand-declares 6 of `TypeInUI`'s 7 literals (`navigation-loaders.ts:7`) instead of `Exclude<TypeInUI, 'Controls'>`. Same failure mode as the MobileNavMenu duplication flagged elsewhere in this review: if `TypeInUI` gains a new browsable type, nothing forces this union to follow — the type just won't appear as a tab, with no compiler signal.

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
```diff
@@ -1,48 +1,164 @@
...
+    Architectures: 'architectures',
+    Patterns: 'patterns',
+    Flows: 'flows',
+    Standards: 'standards',
+    ADRs: 'adrs',
+    Interfaces: 'interfaces',
+};
+
+/** Resolves the `?type=` param to a valid browse type, defaulting deterministically. */
+function useActiveType(counts: NamespaceCounts | undefined): {
```
jpgough-ms: `useActiveType` (lines 43-89) is self-contained — takes only `counts`, returns `{ active, select, tabs }` — and belongs in its own file, same as `useNamespaceItems.ts` next to it. `calm-hub-ui/AGENTS.md`'s Component Decomposition rule: 'When a component grows beyond roughly 150 lines or contains logically distinct UI regions, extract those regions into their own component files.' This file is 166 lines.

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
```diff
@@ -1,48 +1,164 @@
...
+                    console.warn('No openable version for %s; nothing to navigate to', id);
+                }
+            } catch (err) {
+                console.error('Failed to open %s', id, err);
+            }
+        },
+        [namespace, calmService, adrService, navigate]
+    );
+
+    const activeItems = groups.find((g) => g.type === active)?.items ?? [];
```
jpgough-ms: `activeItems` isn't memoized, unlike `tabs`/`defaultType`/`active` (useMemo) and `select`/`openItem` (useCallback) above it in this same file. Minor — `groups` is small — but worth a `useMemo` for consistency, since `NamespacePage` re-renders on unrelated `Hub` state changes (e.g. sidebar toggle).

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoStandardStore.java
```diff
@@ -67,10 +67,14 @@ public List<NamespaceStandardSummary> getStandardsForNamespace(String namespace)
         List<NamespaceStandardSummary> namespaceStanadardSummary = new ArrayList<>();
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 9b17656d — renamed to `namespaceStandardSummary`. That method was rewritten anyway: the count snippet is now `VersionKeySelector.versionCount(...)` (see the dedup below).

calm-hub-ui/src/hub/Hub.tsx
rocketstack-matt: Fixed in c67104d0 — a failed counts fetch now sets a separate `namespaceCountsFailed` flag, and `activeNamespaceCounts` stays `undefined` on failure. The tabs then render resting (no badge) rather than a dimmed 0 contradicting the independently-loaded item grid.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
rocketstack-matt: Fixed in 9b17656d — the `versions` cast is now instanceof-guarded (`rawVersions instanceof Document`) in all 4 Mongo + 4 Nitrite stores, so a non-map value counts as 0 instead of a ClassCastException 500-ing the whole list.

calm-hub/src/main/java/org/finos/calm/domain/namespaces/NamespaceResourceSummary.java
rocketstack-matt: Fixed in 9b17656d — merged the 4 identical summaries into `NamespaceResourceSummary` (domain/namespaces), and the null-guard+count snippet is now `VersionKeySelector.versionCount(Set)` shared by all 8 stores.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
rocketstack-matt: Fixed in c67104d0 — exported `COUNT_FIELD` from resource-type-meta and use it here; the hand switch with its silent `default` is gone.

calm-hub-ui/src/hub/components/namespace-page/resource-type-meta.ts
rocketstack-matt: Fixed in c67104d0 — `NamespaceResourceType = Exclude<TypeInUI, 'Controls'>` now, so a new `TypeInUI` member forces a tab (the `Record`s keyed on it won't compile otherwise).

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
rocketstack-matt: Fixed in c67104d0 — extracted `useActiveType` into its own file (`useActiveType.ts`).

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
rocketstack-matt: Fixed in c67104d0 — `activeItems` is memoized now.

jpgough-ms (approved):

PR #2763, opened 2026-06-30
feat(calm-hub-ui): refine diagram minimap, node legibility and version timeline

rocketstack-matt:
> 🚧 **Draft — stacked on #2762 (phase 2 of the CALM Hub redesign).** Ready for review once #2762 is merged; I'll then rebase this onto `main` so it shows only its own changes. Until then GitHub shows the **cumulative** diff (this phase plus the not-yet-merged prior ones).

Part of #2754.

## Description

Phase 3 of the CALM Hub redesign — refines the existing React Flow diagram view (node/edge semantics unchanged). Fixes the oversized minimap, illegible node labels, and the unclear timeline.

- **MiniMap** — constrained to a fixed 132×84 panel with a low mask opacity, a clear viewport rect, padded clear of the timeline, and made **toggleable** (was an oversized near-solid-blue block overflowing the canvas).
- **Node-label legibility** — labels stay legible after `fitView` and scale with zoom (a zoom floor), instead of shrinking to nothing on dense graphs.
- **Version timeline** — relabelled to **"Browse versions"** with an explanatory line and a version pill, surfacing the existing cross-view behaviour (clicking a moment re-renders the Diagram **and** JSON at that version). Single-version items collapse the strip. This reuses the existing `TimelineBar`, not a rewrite.

## Type of Change
- [x] ✨ New feature (non-breaking change which adds functionality)

## Affected Components
- [x] CALM Hub UI (`calm-hub-ui/`)

## Testing
- [x] I have tested my changes locally
- [x] I have added/updated unit tests
- [x] All existing tests pass

## Checklist
- [x] My commits follow the [conventional commit format](https://www.conventionalcommits.org/)
- [x] I have updated documentation if necessary
- [x] I have added tests for my changes (if applicable)
- [x] My changes follow the project's coding standards

PR #2764, opened 2026-06-30
feat(calm-hub-ui): first-run landing, search disambiguation and dropzone polish

rocketstack-matt:
> 🚧 **Draft — stacked on #2763 (phase 3 of the CALM Hub redesign).** Ready for review once #2763 is merged; I'll then rebase this onto `main` so it shows only its own changes. Until then GitHub shows the **cumulative** diff (this phase plus the not-yet-merged prior ones).

Part of #2754.

## Description

Phase 4 of the CALM Hub redesign — replaces the ~75% blank landing canvas and disambiguates global search.

- **First-run landing** — a heading + sub-copy, 4 catalogue **stat tiles** (Namespaces / Architectures / Patterns / Controls, from the real counts the Hub already holds — not hardcoded), and a "Browse the catalogue" strip of real sample items spanning architectures **and** patterns.
- **Global search disambiguation** — each result now shows its **namespace** chip, so two identically-named items can be told apart.
- **Active-state unification (problem #8)** — the diagram view-tab pill (Diagram / JSON / Deployments) and the Admin active tabs are re-tokenised to the one interaction blue, replacing the inconsistent per-surface highlights.
- **Visualizer dropzone polish** — a bordered dropzone with an icon, an accepted-file-type hint, and error handling (`accept` + a guarded parse) for the standalone upload.

**Note for reviewers:** there is no recency API, so the landing intentionally shows a "Browse the catalogue" sample (architectures + patterns) rather than "recently updated". The standalone Visualizer page itself was removed in phase 1; the hardened dropzone lives on the shared `Drawer`.

## Type of Change

## Affected Components

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-hub-ui/src/hub/Hub.tsx
+++ b/calm-hub-ui/src/hub/Hub.tsx
@@ -5,6 +5,7 @@ import { ExploreRail } from './components/explore-rail/ExploreRail.js';
 import { MobileNavMenu } from './components/tree-navigation/MobileNavMenu.js';
 import { NamespacePage } from './components/namespace-page/NamespacePage.js';
 import { DomainPage } from './components/domain-page/DomainPage.js';
+import { FirstRunLanding } from './components/first-run-landing/FirstRunLanding.js';
 import { useResourceFromRoute } from './hooks/useResourceFromRoute.js';
 import { useIsMobile } from '../hooks/useMediaQuery.js';
 import { Data, Adr } from '../model/calm.js';
@@ -195,7 +196,9 @@ export default function Hub() {
 
     // Route decides the content pane. A loaded resource (including an in-place
     // control/interface selected from the domain/namespace page) takes precedence
-    // over the route-driven page so its detail view shows.
+    // over the route-driven page so its detail view shows. With nothing loaded and
+    // no namespace/domain route (i.e. `/`), the first-run landing fills what was
+    // the ~75% blank canvas (redesign problem #7).
     const content =
         isDetailRoute || controlData || interfaceData || adrData || data ? (
             detailContent
@@ -204,11 +207,11 @@ export default function Hub() {
         ) : activeDomain ? (
             <DomainPage domain={activeDomain} controlCount={domainControlCount} onControlLoad={handleControlLoad} />
         ) : (
-            // Dedicated landing arm: nothing loaded and no browse route active. Kept separate
-            // from detailContent so it never renders DocumentDetailSection with undefined data.
-            <div className="flex-1 flex items-center justify-center text-[14px] text-base-content/50">
-                Select a namespace or control domain from the Explore rail to begin.
-            </div>
+            <FirstRunLanding
+                namespaceCounts={namespaceCounts}
+                domainCounts={domainCounts}
+                countsLoaded={namespaceCountsLoaded}
+            />
         );
 
     return (
--- a/calm-hub-ui/src/hub/Hub.tsx
+++ b/calm-hub-ui/src/hub/Hub.tsx
@@ -38,6 +38,7 @@ export default function Hub() {
     // zero, so consumers must render them as unknown rather than a misleading 0.
     const [namespaceCountsFailed, setNamespaceCountsFailed] = useState(false);
     const [domainCounts, setDomainCounts] = useState<DomainControlCount[]>([]);
+    const [domainCountsLoaded, setDomainCountsLoaded] = useState(false);
     const isMobile = useIsMobile();
 
     // Route-first content selection (redesign problem #4): the same <Hub/> element
@@ -71,7 +72,11 @@ export default function Hub() {
             // unknown (loading)" from "known zero" — an absent namespace after a
             // successful fetch is genuinely zero, not still loading.
             .finally(() => setNamespaceCountsLoaded(true));
-        countsService.fetchDomainCounts().then(setDomainCounts).catch(() => setDomainCounts([]));
+        countsService
+            .fetchDomainCounts()
+            .then(setDomainCounts)
+            .catch(() => setDomainCounts([]))
+            .finally(() => setDomainCountsLoaded(true));
     }, [countsService]);
 
     useEffect(() => {
@@ -210,7 +215,9 @@ export default function Hub() {
             <FirstRunLanding
                 namespaceCounts={namespaceCounts}
                 domainCounts={domainCounts}
-                countsLoaded={namespaceCountsLoaded}
+                // Both must settle: the tiles include a Controls total from domainCounts,
+                // so gating on namespace counts alone would flash a 0 for Controls.
+                countsLoaded={namespaceCountsLoaded && domainCountsLoaded}
             />
         );
 
--- a/calm-hub-ui/src/hub/Hub.tsx
+++ b/calm-hub-ui/src/hub/Hub.tsx
@@ -215,9 +215,11 @@ export default function Hub() {
             <FirstRunLanding
                 namespaceCounts={namespaceCounts}
                 domainCounts={domainCounts}
-                // Both must settle: the tiles include a Controls total from domainCounts,
-                // so gating on namespace counts alone would flash a 0 for Controls.
-                countsLoaded={namespaceCountsLoaded && domainCountsLoaded}
+                // Ready only once both fetches have settled AND the namespace fetch
+                // didn't fail: the tiles include a Controls total from domainCounts (so
+                // namespace-only gating would flash a 0 for Controls), and a failed
+                // namespace fetch is unknown, not zero — hold the placeholder in both cases.
+                countsLoaded={namespaceCountsLoaded && domainCountsLoaded && !namespaceCountsFailed}
             />
         );
 
--- a/calm-hub-ui/src/hub/components/first-run-landing/CatalogueCard.tsx
+++ b/calm-hub-ui/src/hub/components/first-run-landing/CatalogueCard.tsx
@@ -1,0 +1,58 @@
+import { Link } from 'react-router-dom';
+import { colors } from '../../../theme/colors.js';
+import { redesignTokens } from '../../../theme/redesign-tokens.js';
+import { TypeBadge } from '../namespace-page/TypeBadge.js';
+import { getResourceTypeColors } from '../namespace-page/resource-type-meta.js';
+import type { CatalogueHighlight } from './useCatalogueHighlights.js';
+
+interface CatalogueCardProps {
+    item: CatalogueHighlight;
+}
+
+/**
+ * A compact catalogue-highlight card for the landing's "Browse the catalogue"
+ * strip: a type-tinted striped thumbnail, the item name, a resource-type
+ * {@link TypeBadge} and a mono namespace chip. Thumbnail tint and badge are
+ * driven by the item's own type, so architectures and patterns read distinctly.
+ *
+ * It links to the item's namespace page (not its detail route): the detail route
+ * needs the latest version, which is only resolvable via a per-item fetch — an
+ * N+1 the landing must avoid. The namespace page resolves the version on click.
+ */
+export function CatalogueCard({ item }: CatalogueCardProps) {
+    const { accent, tint } = getResourceTypeColors(item.type);
+    const stripes = `repeating-linear-gradient(135deg, ${tint}, ${tint} 7px, ${accent}20 7px, ${accent}20 14px)`;
+    const tabParam = item.type === 'Patterns' ? 'patterns' : 'architectures';
+
+    return (
+        <Link
+            to={`/namespace/${item.namespace}?type=${tabParam}`}
+            data-testid="catalogue-card"
+            className="group relative block rounded-[12px] overflow-hidden bg-base-100 no-underline hover:-translate-y-0.5 hover:shadow-md"
+            style={{
+                border: `1px solid ${colors.redesign.border}`,
+                boxShadow: redesignTokens.shadow.card,
+                transition: redesignTokens.transition,
+            }}
+        >
+            <div style={{ height: 64, background: stripes }} />
+            <div className="p-[14px]">
+                <div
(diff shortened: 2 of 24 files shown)


calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -31,21 +32,40 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
...
-    const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop });
+    // Clear any prior error as soon as a new drag begins, so a fresh attempt
+    // starts from a clean slate.
+    const onDragEnter = useCallback(() => setDropError(undefined), []);
+
+    const { getRootProps, getInputProps, isDragActive } = useDropzone({
+        onDrop,
+        onDragEnter,
+        accept: { 'application/json': ['.json'] },
+    });
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 7a4bd71e — added an `onDropRejected` handler that sets `dropError`, so a file rejected by the accept filter now surfaces the accepted-types message in the empty state instead of `onDrop` returning early with no feedback.

calm-hub-ui/src/hub/Hub.tsx
```diff
@@ -204,11 +207,11 @@ export default function Hub() {
...
-            // Dedicated landing arm: nothing loaded and no browse route active. Kept separate
-            // from detailContent so it never renders DocumentDetailSection with undefined data.
-            <div className="flex-1 flex items-center justify-center text-[14px] text-base-content/50">
-                Select a namespace or control domain from the Explore rail to begin.
-            </div>
+            <FirstRunLanding
+                namespaceCounts={namespaceCounts}
+                domainCounts={domainCounts}
+                countsLoaded={namespaceCountsLoaded}
+            />
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 7a4bd71e — added a `domainCountsLoaded` flag and pass `namespaceCountsLoaded && domainCountsLoaded` to FirstRunLanding, so the tiles (including the Controls total from `domainCounts`) hold the placeholder until both fetches settle.

calm-hub-ui/src/hub/Hub.tsx
```diff
@@ -204,11 +212,13 @@ export default function Hub() {
...
-            // from detailContent so it never renders DocumentDetailSection with undefined data.
-            <div className="flex-1 flex items-center justify-center text-[14px] text-base-content/50">
-                Select a namespace or control domain from the Explore rail to begin.
-            </div>
+            <FirstRunLanding
+                namespaceCounts={namespaceCounts}
+                domainCounts={domainCounts}
+                // Both must settle: the tiles include a Controls total from domainCounts,
+                // so gating on namespace counts alone would flash a 0 for Controls.
+                countsLoaded={namespaceCountsLoaded && domainCountsLoaded}
```
jpgough-ms: `countsLoaded` ignores `namespaceCountsFailed`, so a failed counts fetch renders as a literal "0" instead of an unknown/loading state. If `fetchNamespaceCounts()` rejects, `namespaceCounts` becomes `[]` and `namespaceCountsFailed` becomes `true`, but `namespaceCountsLoaded` still flips `true` in `.finally()`. This gate only checks the loaded flags, not the failed flag, so `StatTile` renders "0" for Namespaces/Architectures/Patterns — misrepresenting a backend failure as an empty catalogue. Line 171 in this same file already guards `activeNamespaceCounts` against exactly this case (`if (!namespaceCountsLoaded || namespaceCountsFailed) return undefined;`); consider this gate doing the same.

Separately: with two independent `loaded` booleans now ANDed here, a future third count source needs another `useState` plus another `&&` term — easy to add a source and forget the flag, silently reintroducing the flash-to-zero bug above.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -31,21 +32,47 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
...
+    // so surface it here rather than failing silently.
+    const onDropRejected = useCallback(() => {
+        setDropError('That file type isn’t supported — drop a CALM JSON file (architecture / pattern).');
+    }, []);
+
+    const { getRootProps, getInputProps, isDragActive } = useDropzone({
+        onDrop,
+        onDragEnter,
+        onDropRejected,
+        accept: { 'application/json': ['.json'] },
```
jpgough-ms: `accept: { 'application/json': ['.json'] }` filters by file extension/MIME before `onDrop` ever runs. A user dropping valid CALM JSON saved with a different extension (`.calm`, `.txt`, none) is routed straight to `onDropRejected` and never gets parsed — a regression from the previous unrestricted behavior. Traced through react-dropzone's `fileAccepted`/`attr-accept` logic: files failing this check never reach `acceptedFiles`.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -31,21 +32,47 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
...
     }, []);
 
-    const { getRootProps, getInputProps, isDragActive } = useDropzone({ onDrop });
+    // Clear any prior error as soon as a new drag begins, so a fresh attempt
+    // starts from a clean slate.
+    const onDragEnter = useCallback(() => setDropError(undefined), []);
+
+    // A file rejected by `accept` (wrong type) never reaches onDrop's acceptedFiles,
+    // so surface it here rather than failing silently.
+    const onDropRejected = useCallback(() => {
```
jpgough-ms: `onDropRejected` discards react-dropzone's `FileRejection[]` argument and always shows the same "file type isn't supported" message, regardless of the actual rejection reason (wrong type vs. too many files vs. any future size limit).

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -245,16 +272,7 @@ export function Drawer({ data, onItemSelect, decorators: decoratorsProp }: Drawe
...
-                    {isDragActive ? (
-                        <p>Drop your file here ...</p>
-                    ) : (
-                        <p>
-                            {'Drag and drop your file here or '}
-                            <span className="border-b border-dotted border-black pb-1">Browse</span>
-                        </p>
-                    )}
-                </div>
+                <DropzoneEmptyState isDragActive={isDragActive} error={dropError} />
```
jpgough-ms, line 287: `dropError` only renders via `DropzoneEmptyState`, which is only mounted in the `!hasContent` branch. If a user drops a bad file while an architecture/pattern is already loaded, `dropError` is set but `hasContent` never clears — the error is set in state and never shown.

calm-hub-ui/src/hub/components/first-run-landing/CatalogueCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * {@link TypeBadge} and a mono namespace chip. Thumbnail tint and badge are
+ * driven by the item's own type, so architectures and patterns read distinctly.
+ *
+ * It links to the item's namespace page (not its detail route): the detail route
+ * needs the latest version, which is only resolvable via a per-item fetch — an
+ * N+1 the landing must avoid. The namespace page resolves the version on click.
+ */
+export function CatalogueCard({ item }: CatalogueCardProps) {
+    const { accent, tint } = getResourceTypeColors(item.type);
+    const stripes = `repeating-linear-gradient(135deg, ${tint}, ${tint} 7px, ${accent}20 7px, ${accent}20 14px)`;
```
jpgough-ms: Duplicates `ItemCard.tsx`'s striped-thumbnail formula, card shell (border/shadow/hover), and mono meta-chip styling verbatim rather than reusing/parameterizing it. The two have already drifted (64px vs. 96px thumbnail height, unexplained). `CatalogueHighlight`'s data shape is compatible with `ItemCardProps`, so this looks reuseable via an optional `href`/`versionCount` prop.

calm-hub-ui/src/hub/components/first-run-landing/useCatalogueHighlights.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        let cancelled = false;
+        setLoading(true);
+
+        Promise.all(
+            targets.map((namespace) =>
+                Promise.all([
+                    service
+                        .fetchArchitectureSummaries(namespace)
```
jpgough-ms: A namespace qualifies for probing if *either* `architectures > 0` or `patterns > 0`, but both `fetchArchitectureSummaries` and `fetchPatternSummaries` are then called unconditionally — wasting a request per namespace when one count is 0, using data the hook already has to skip it.

calm-hub-ui/src/hub/components/first-run-landing/useCatalogueHighlights.ts
```diff
@@ -0,0 +1,113 @@
...
+                        .catch(() => [] as CatalogueHighlight[]),
+                    service
+                        .fetchPatternSummaries(namespace)
+                        .then((summaries) => toHighlights(summaries, namespace, 'Patterns'))
+                        .catch(() => [] as CatalogueHighlight[]),
+                ]).then(([architectures, patterns]) => [...architectures, ...patterns])
+            )
+        ).then((perNamespace) => {
+            if (cancelled) return;
+            setHighlights(perNamespace.flat().slice(0, MAX_HIGHLIGHTS));
```
jpgough-ms, line 115: `fetchArchitectureSummaries`/`fetchPatternSummaries` have no limit/pagination param and return the full namespace list, even though only `MAX_HIGHLIGHTS = 3` items are ever kept — downloads/parses the whole catalogue per namespace just to show up to 3 cards.

jpgough-ms (commented):
## Review notes

Two off-diff findings that don't have a line in this PR's diff hunks, noted here instead of as inline comments:

- **`calm-hub-ui/src/components/navbar/ExplorerSearch.tsx`** is a near-duplicate of `GlobalSearchBar.tsx` (same `FlatResult`, `TYPE_LABELS`, `flattenResults`), used live in `MobileNavMenu.tsx`. This PR's namespace-disambiguation fix (the chip distinguishing same-named results across namespaces) was applied to `GlobalSearchBar` but not to `ExplorerSearch` — the ambiguity bug this PR fixes still exists in the mobile explorer search.

- **`FirstRunLanding` stays mounted underneath the mobile-nav overlay** (`Hub.tsx`, the `content` render) rather than being hidden/unmounted while it's open. Since `isMobileNavOpen` defaults to `true`, at the default Cypress viewport (1000×660, under the 1023px mobile breakpoint) both the mobile nav's "Namespaces" row and the landing page's "Namespaces" stat tile are on screen simultaneously — this breaks `cypress/e2e/home.cy.ts`'s `cy.findByText('Namespaces').should('exist')` (line 11) with a "multiple elements found" error.

calm-hub-ui/src/hub/Hub.tsx
rocketstack-matt: Fixed in a031b2f7 — the gate now also checks `!namespaceCountsFailed` (mirroring the line-171 `activeNamespaceCounts` guard), so a failed namespace fetch holds the placeholder instead of rendering 0.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
rocketstack-matt: Fixed in a031b2f7 — removed the `accept` filter entirely. You're right that it rejected valid CALM JSON saved with a non-`.json` extension; `onDrop` already parses and surfaces `dropError` on non-JSON, so validation lives there. This also reverts the `onDropRejected` I'd added — it was papering over the regression.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
rocketstack-matt: Moot after removing the `accept` filter in a031b2f7 — `onDropRejected` is gone.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
rocketstack-matt: Fixed in a031b2f7 — `dropError` now also renders as an alert banner in the has-content branch, so a bad drop over already-loaded content is visible.

calm-hub-ui/src/hub/components/first-run-landing/CatalogueCard.tsx
rocketstack-matt: Fixed in a031b2f7 — `CatalogueCard` now reuses `ItemCard` (parameterised with optional `href`/`meta`/`thumbnailHeight`/`testId`), removing the duplicated shell and the 64/96px drift.

calm-hub-ui/src/hub/components/first-run-landing/useCatalogueHighlights.ts
rocketstack-matt: Fixed in a031b2f7 — the hook encodes which types each namespace has in the memo key and skips fetching a type whose count is 0, so a namespace with only architectures no longer fires a pattern request.

calm-hub-ui/src/hub/components/first-run-landing/useCatalogueHighlights.ts
rocketstack-matt: Agreed — a real fix needs a `limit`/pagination param on the summary endpoints (backend), so I've opened #2780 to track it. For now it's bounded to `MAX_NAMESPACES` namespaces, so the over-fetch is capped but not eliminated.

jpgough-ms (approved):

Issue #2099, opened 2026-02-02
CALM CLI Server Mode additional validation


Budlee:
### Target Project:
CALM CLI

### Description of Feature:
Currently the CALM Validation of pattern instances is done against patterns that already exist in the project directory.
It should be possible to pass in a pattern instance and the pattern to the validation in order for it to be validated. 

### User Stories:
As a developer, I want to validate a CALM instance against a custom pattern that I provide so that I am assured that the instance of the pattern conforms to the pattern.

### Current Limitations:
Currently all pattern instances need to be defined up front and cannot be calculated on the fly.

### Proposed Implementation:
Add an additional parameter in the request body of the payload to the CLI server mode to allow the upload of a pattern or patterns (required due to the hierarchical status of CALM patterns) that the CLI server mode can validate against. 

### Alternatives Considered:
CALM Hub is not available everywhere yet. The CLI is very accessible. 

### Testing Strategy:
The CLI server mode will be tested with a new pattern instance and a new pattern that is not already within the CALM project, and the CLI server mode should be able to successfully validate against this. The test would be unit tested and integration tested. 

### Documentation Requirements:
Additional documentation required would be to include the new request parameter for the CLI server mode. 

### Implementation Checklist:
- [ ] Design reviewed and approved
- [ ] Implementation completed
- [ ] Tests written and passing
- [ ] Documentation updated

### Additional Context:
...add any other context, diagrams, mockups, or screenshots about the feature request here...

rocketstack-matt:
@Budlee is this issue now complete?

PR #2772, opened 2026-07-02
feat(calm-server): calm server validation against a custom pattern

aayush-razdan:
## Description
Addresses https://github.com/finos/architecture-as-code/issues/2099 - "Currently the CALM Validation of pattern instances is done against patterns that already exist in the project directory. It should be possible to pass in a pattern instance and the pattern to the validation in order for it to be validated."

This is now available through the endpoint /calm/validate/with-pattern

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [x] CALM Server (`calm-server/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


calm-server/src/server/routes/validation-route.ts
```diff
@@ -27,6 +70,7 @@ export class ValidationRouter {
 
     private initializeRoutes(router: Router) {
         router.post('/', this.validateSchema);
+        router.post('/with-pattern', this.validateWithPattern);
```
Budlee: Can we not fold this into the '/' request? 

Doesn't the validateWithPattern logic not map to the validateSchema. The difference is that the if the request body also has a pattern included then you will also verify the instance against the pattern? 

I don't think we need to separate the two

calm-server/src/server/routes/validation-route.ts
```diff
@@ -91,5 +225,10 @@ class ErrorResponse {
 }
 
 class ValidationRequest {
-    architecture: string;
+    architecture!: string;
 }
+
+class ValidationWithPatternRequest {
+    architecture!: string;
+    pattern!: string;
```
Budlee: Here is my thinking to the previous comment, we just have ValidationRequest and the pattern is an optional field

Budlee (changes requested):
I think that we do not need an extra route and can fold this into a single validate endpoint that needs to be extended. 
This should not change compatibility as current usage will be checking the architecture conforms to the schema. If the optional pattern is applied then the follow on logic to check it adheres will then apply.

calm-server/src/server/routes/validation-route.ts
```diff
@@ -52,9 +88,28 @@ export class ValidationRouter {
...
+            return res.status(400).type('json').send(new ErrorResponse('The architecture must be a JSON object'));
+        }
+
+        if (hasNonArrayRelationships(architecture)) {
+            this.logger.error('The "relationships" field in the architecture is not an array');
+            return res.status(400).type('json').send(new ErrorResponse('The "relationships" field in the architecture must be an array'));
+        }
+
+        if (hasPattern(req.body.pattern)) {
+            return this.validateWithPattern(architecture, req.body.pattern, res);
```
Budlee: I believe that having this return statement to validateWithPattern shifts from the functions purpose as the return is a core part of the processing and is now midway through the code and not a top-to-bottom function as I read through it

Budlee (changes requested):

Budlee (approved):

rocketstack-matt (approved):

Issue #492, opened 2024-10-17
Validate Detailed Architectures


jpgough-ms:
It would also be good to validate detailed architectures (and the pattern if it uses one)

_Originally posted by @jpgough-ms in https://github.com/finos/architecture-as-code/issues/443#issuecomment-2397194711_

rocketstack-matt:
Do we need to follow detailed architectures, assuming referenced architectures will have been validated in their own right? If we did this, how many hops do you go and could you get stuck in an infinite loop?

No one has actually asked for this so closing until. someone comes back with a specific use case which needs it.

Issue #443, opened 2024-10-07
CALM CLI Validation of Control Configurations against Control Requirements


jpgough-ms:
### Description of Problem:

In the October schema, we have introduced the concept of control requirements (a JSON schema following the `control-requirement` base schema). The specific control requirement schema use a `control-configuration` to specify the JSON document that meets the schema.

For example: 

```
"controls": {
        "cbom": {
            "description": "Control requirements for delivering patterns",
            "requirements": [
              {
                "control-requirement": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/control-example/pre-prod-review-specification.json",
                "control-config": "https://raw.githubusercontent.com/finos/architecture-as-code/main/calm/control-example/pre-prod-review-configuration.json"
              }
            ]
          }
      }
```

The validator should JSON schema validate the control-configuration against the control-requirement.

jpgough-ms:
It would also be good to validate detailed architectures (and the pattern if it uses one)

rocketstack-matt:
> It would also be good to validate detailed architectures (and the pattern if it uses one)

Can we add that as a separate issue @jpgough-ms ? I think it's useful, but perhaps we can live without it for v1?

willosborne:
Looking to pick this up now.

rocketstack-matt:
Closing — implemented by the validation engine redesign in #2805 (merged 2026-07-16): control configurations are now JSON-schema-validated against their control requirements (shared/src/commands/validate/validate-controls.ts), and detailed architectures are loaded and validated recursively (node-details rule). Follow-ups for edge cases are tracked in #2821/#2822/#2827.

PR #2778, opened 2026-07-03
feat(shared): validate node details and controls 

willosborne:
## Description
Implements #492 and #443 

- Recursively validate node detailed-architecture against the pattern if provided
- Adds fork mechanism to schema dir to allow easy splitting without needing to reload all the doc loaders
- Validate controls against their requirement if provided
- Can load in-line schemas from the pattern file if the requirement URI is a #/json/path style ref

## Type of Change

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


shared/src/commands/validate/validate.ts
```diff
@@ -275,10 +297,37 @@ async function validateArchitectureOnly(architecture: object, schemaDirectory: S
...
+    }
+
     logger.debug(`Returning validation outcome with ${jsonSchemaValidations.length} JSON schema validations, errors: ${errors}`);
     return new ValidationOutcome(jsonSchemaValidations, spectralResultForArchitecture.spectralIssues, errors, warnings);
 }
 
+const validateArchitectureDispatch: ArchitectureValidator = (arch, pat, dir, dbg, visited) =>
+    pat !== undefined
+        ? validateArchitectureAgainstPattern(arch, pat, dir, dbg, visited)
+        : validateArchitectureOnly(arch, dir, dbg);
```
rocketstack-matt, line 330: Cycle detection is silently disabled whenever a node in a detailed-architecture chain has no discoverable pattern. `validateArchitectureDispatch` drops the `visited` set here when routing to `validateArchitectureOnly` (line 266), which doesn't even accept a `visitedUrls` parameter — it allocates a fresh empty `Set<string>()` for its own nested `validateNodeDetails` call (line 311), disconnected from whatever the caller had accumulated.

I reproduced this: a single node whose `detailed-architecture` points to itself, with no `required-pattern`/`$schema`, crashes the process with a V8 heap OOM instead of returning a cycle-detected error — the exact failure mode raised when #492 was originally closed. No test exercises the real dispatch wiring (`validate-node-details.spec.ts` only unit-tests `validateNodeDetails` with a mocked `recursiveValidator`), so this is invisible to CI.

Fix: add `visitedUrls: Set<string> = new Set<string>()` to `validateArchitectureOnly`'s signature, thread it into its `validateNodeDetails` call instead of `new Set<string>()`, and pass `visited` through in the dispatch's no-pattern branch (`validateArchitectureOnly(arch, dir, dbg, visited)`).

shared/src/commands/validate/validate-node-details.ts
```diff
@@ -0,0 +1,229 @@
...
+        'node-details-validation',
+        'error',
+        message,
+        path,
+        undefined, undefined, undefined, undefined, undefined,
+        'architecture'
+    );
+}
+
+function toErrorMessage(error: unknown): string {
```
rocketstack-matt: `toErrorMessage` is now defined identically three times: here, in `validate-controls.ts:308`, and the pre-existing private one in `validate.ts:444`. Worth consolidating into a shared util (non-blocking nit).

rocketstack-matt (changes requested):
Control-config validation against control-requirement schemas (#443) looks solid and well tested. The recursive detailed-architecture validation (#492) has a real infinite-recursion bug that reproduces a heap OOM crash — see inline comment for details and a fix.

shared/src/commands/validate/validate-node-details.ts
```diff
@@ -0,0 +1,229 @@
...
+async function validateNodeDetail(
+    node: CalmNode,
+    nodeIdx: number,
+    schemaDirectory: SchemaDirectory,
+    debug: boolean,
+    recursiveValidator: ArchitectureValidator,
+    visitedUrls: Set<string>
+): Promise<NodeDetailResult> {
+    const logger = initLogger(debug, 'validate-node-details');
+    const archUrl = node.details?.detailedArchitecture?.reference;
```
LeighFinegold, line 88: Small consistency thought here. `node.details.detailedArchitecture` / `requiredPattern` are `ResolvableAndAdaptable<CalmCoreSchema, CalmCore>` in `calm-models` (and `CalmControlDetail.requirement` / `configUrl` are `Resolvable`), so the platform already has a first-class notion of "a reference that knows how to fetch itself".

I get why maybe we don't use `.value` directly - validation needs the raw document to run JSON-Schema + Spectral against, and `ResolvableAndAdaptable` adapts to a `CalmCore` and discards the raw schema, so `.value` isn't what we want and a `toSchema()` round-trip would validate normalised output rather than the user's input. That's fair.

Rather than reading `.reference` and running a parallel load path, could we still drive resolution through the `Resolvable` (its `dereference(resolver)` takes any `(url) => Promise<S>`, so a `SchemaDirectory`backed resolver slots straight in)? Even if validation ultimately needs the raw JSON, keeping "how a reference is resolved" owned by the model, rather than re-deriving it per consumer.

shared/src/commands/validate/validate-node-details.ts
LeighFinegold: there's already a canonical `getErrorMessage` in `shared/src/error-utils.ts` (identical logic), consumed by `dereference-visitor`, `file-system-document-loader`, the resolver, and the template modules.

shared/src/commands/validate/validate-controls.ts
```diff
@@ -0,0 +1,312 @@
...
+    controlId: string;
+    requirements: CalmControlDetail[];
+}
+
+/**
+ * Load all the controls from an architecture and parse to a uniform definition.
+ * @param calmCore Parsed CALM architecture
+ * @returns List of control contexts in standard form.
+ */
+function collectControlContexts(calmCore: CalmCore): FlatControlContext[] {
```
LeighFinegold, line 227: `collectControlContexts` walks controls across top-level/nodes/relationships/flows via `Object.entries(x.controls.data)` + `control.requirements.forEach(d => d.requirement.reference)`. `shared/src/docify/graphing/control-registry.ts` (`ControlRegistry`) already does the nodes/relationships/flows enumeration with the same shape.

The outputs differ (validation needs path prefixes + top-level controls; the registry builds requirement/config maps for rendering), so they can't be identical - but the *traversal* is duplicated. A single shared "iterate all controls with their model path" helper that both this validator and `ControlRegistry` consume would keep control locations enumerated in exactly one place

shared/src/commands/validate/validate.ts
LeighFinegold: Bit of background/context here: `DereferencingVisitor` (`shared/src/model-visitor/`) started life as a *sample* of how to visit the CALM model - a proof-of-concept, not a hardened traversal. It shows its origins: no visited set (so it isn't cycle-safe), and it swallows resolution failures as `console.warn`. So I'm explicitly not saying "just reuse the visitor as-is" - for validation we need cycle safety and failures surfaced as `ValidationOutput`, neither of which it does today.

But that's the opportunity rather than a blocker: instead of adding a third hand-rolled walk (this recursive validator, alongside the visitor and `ControlRegistry`, none of which share traversal), we could invest in making the model-visitor a first-class citizen - a single cycle-safe walk over the model, with the per-consumer behaviour (dereference / validate / index) layered on top as visitor implementations. That turns the sample into real infrastructure and means the cycle gap @rocketstack-matt flagged (the pattern-less dispatch branch dropping `visited`) is handled once, structurally, rather than re-solved in each caller. The immediate must-fix is still that `visited` gap; this is the direction I'd like the detailed-architecture walk to converge on.

LeighFinegold:
One additional constraint I noticed: this change turns on recursive validation for **all** detailed
architectures. Is that really what we want globally, or is it only the right default for the CLI use
case? For something like CalmHub validating on upload, always-on deep traversal may not be desirable.

As discussed, I've raised a separate PR https://github.com/LeighFinegold/architecture-as-code/pull/26  on my fork that addresses what I think my points on this PR were, and also reiterates that we need to properly restate the design and assumption of the validation CLI endpoint (although I think there are still open questions)

LeighFinegold:
> One additional constraint I noticed: this change turns on recursive validation for **all** detailed architectures. Is that really what we want globally, or is it only the right default for the CLI use case? For something like CalmHub validating on upload, always-on deep traversal may not be desirable.
> 
> As discussed, I've raised a separate PR [LeighFinegold#26](https://github.com/LeighFinegold/architecture-as-code/pull/26) on my fork that addresses what I think my points on this PR were, and also reiterates that we need to properly restate the design and assumption of the validation CLI endpoint (although I think there are still open questions)

Caught up with @willosborne  briefly on Thursday and we agreed that perhaps we should converge on the suggestion PR approach but close out early next week Monday

willosborne:
implemented by #2805

PR #2785, opened 2026-07-06
CALMHub/CLI smoke tests

willosborne:
## Description
Adds a suite of tests to run the CLI against a real instance of CALMHub, verifying state after each command.

Testing paths:
- `calm generate`/`calm validate` flow, both using local files and referencing things in CALMHub
- `calm hub` suite including push, pull, and the auto-bump features
- `calm workspace suite` including the `check`, `bump` and `push --fail-if-modified` commands.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [x] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


willosborne:
Smoke tests do everything they're meant to: https://github.com/finos/architecture-as-code/actions/runs/28792578706/job/85374749971?pr=2785

willosborne:
@jpgough-ms would be good to get your 2 cents on this :)

cli/src/test_helpers/cli-runner.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Packs the CLI package at `repoRoot` (cli/) via `npm pack`, installs the tarball into a fresh
+ * temp dir, and returns a runner bound to the installed `calm` binary. Requires the CLI to
+ * already be built (`npm run build:cli`) - `npm pack` reads package.json's `files: ["dist/"]`.
+ */
+export function installPackedCli(repoRoot: string, prefix: string): CliInstall {
+    const tempDir = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+
+    const tgzName = execSync('npm pack', { cwd: repoRoot }).toString().trim();
```
rocketstack-matt: This duplicates the pack/install-into-tempdir flow already inline in `cli/src/cli.e2e.spec.ts` (lines 34-51) instead of refactoring that spec to use this new shared helper. Worth consolidating so there's one copy of the npm-pack logic to maintain — could be a quick follow-up if not in scope here.

cli/smoke/sanity.smoke.spec.ts
```diff
@@ -0,0 +1,21 @@
+import path from 'path';
+import { afterAll, beforeAll, describe, expect, test } from 'vitest';
+import { installPackedCli, type CliInstall } from '../src/test_helpers/cli-runner';
+
+// Proves the smoke glob + packed-CLI harness work without needing Docker.
```
rocketstack-matt: Nit: this comment says the file proves the harness works "without needing Docker", but `vitest.smoke.mts`'s `globalSetup` always brings up Docker/CalmHub for the whole run regardless of which spec is targeted, so running just this file still requires Docker.

rocketstack-matt (commented):
A couple of non-blocking notes. Nice to see this validated against a live hub already.

cli/smoke/sanity.smoke.spec.ts
willosborne: yeah this comment needs to go lol. the smoke tests definitely need docker.

willosborne: removed

cli/src/test_helpers/cli-runner.ts
willosborne: de-duplicated this by reworking cli.e2e.spec.ts to use the new packed runner tool

cli/smoke/hub-crud.smoke.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        expect(await api.listMappings(NS, TYPE)).toContain(MAPPING);
+        expect(stdout).toContain(MAPPING);
+    });
+
+    test('pull architecture writes a file matching the stored document', async () => {
+        const out = path.join(cli.tempDir, 'pulled.json');
+        await cli.run(['hub', 'pull', 'architecture', '--namespace', NS, '-m', MAPPING, '--ver', '1.0.0', '-o', out, '-c', SMOKE_HUB_URL]);
+        const pulled = readJson(out);
+        const stored = await api.getDocument(NS, TYPE, MAPPING, '1.0.0');
+        expect(canonicalEqual(JSON.stringify(pulled), JSON.stringify(stored))).toBe(true);
```
jpgough-ms: `canonicalEqual` is called on `JSON.stringify(pulled)`/`JSON.stringify(stored)` instead of the parsed objects. `canonicalize()` only recurses into objects/arrays and returns strings unchanged, so this degrades to strict string equality — defeating the key-order-independent comparison `canonicalEqual` exists for. Every other call site in the repo (`workspace/push.ts:99`, `workspace/bump.ts:97`, `hub-commands.ts:196`) passes parsed objects; suggest the same here: `canonicalEqual(pulled, stored)`.

cli/vitest.config.mts
```diff
@@ -10,14 +10,15 @@ const v8CoverageSettings: CoverageV8Options = {
         lines: 75,
         statements: 75
     },
-    exclude: ['test_fixtures/**', '*.config.ts'],
+    exclude: ['test_fixtures/**', '*.config.ts', 'smoke/**'],
```
jpgough-ms, line 13: `cli-runner.ts` isn't in this exclude list, but it's only exercised by `cli/smoke/*.smoke.spec.ts`, which is excluded from the default test run (line 21). CI already shows it at 0% coverage, pulling the branches metric to 85.05% against an 85% threshold. Add `src/test_helpers/cli-runner.ts` here before it tips over.

cli/smoke/global-setup.ts
```diff
@@ -0,0 +1,41 @@
...
+    try {
+        compose(['up', '-d']);
+        await waitForHub(SMOKE_HUB_URL);
+    } catch (err) {
+        try {
+            compose(['logs']);
+        } catch {
+            // Best-effort diagnostics; never let a logs failure skip teardown.
+        }
+        compose(['down', '-v']);
```
jpgough-ms: `compose(['down', '-v'])` here isn't guarded, unlike the `compose(['logs'])` call above it. If teardown itself throws (docker daemon busy, container in a bad state), that exception replaces the original `err` and `throw err` on the next line never runs — the diagnostic "CalmHub not ready" message is lost in favor of an opaque compose-down failure, and cleanup may not have completed.

cli/src/test_helpers/cli-runner.ts
```diff
@@ -0,0 +1,47 @@
...
+    run(args: string[], opts?: { cwd?: string; env?: NodeJS.ProcessEnv }): ResultPromise;
+    cleanup(): void;
+}
+
+/**
+ * Packs the CLI package at `repoRoot` (cli/) via `npm pack`, installs the tarball into a fresh
+ * temp dir, and returns a runner bound to the installed `calm` binary. Requires the CLI to
+ * already be built (`npm run build:cli`) - `npm pack` reads package.json's `files: ["dist/"]`.
+ */
+export function installPackedCli(repoRoot: string, prefix: string): CliInstall {
```
jpgough-ms, line 18: Two things worth a look here: (1) no try/finally — a failure partway through (e.g. a flaky `npm install`) leaves the temp dir uncleaned and can strand an untracked `.tgz` inside `cli/`, since every caller's `afterAll(() => cli?.cleanup())` becomes a silent no-op once `cli` stays undefined. (2) it's invoked once per spec file (4 call sites) with `fileParallelism: false` forcing them serial — that's 4 full pack+install cycles from scratch instead of one shared install, adding avoidable wall-clock time to every run.

cli/smoke/harness/hub-ready.ts
```diff
@@ -0,0 +1,24 @@
...
+export async function waitForHub(
+    baseUrl: string,
+    timeoutMs = 120_000,
+    intervalMs = 3_000
+): Promise<void> {
+    const deadline = Date.now() + timeoutMs;
+    let lastErr: unknown;
+    while (Date.now() < deadline) {
+        try {
+            const res = await fetch(`${baseUrl}/q/swagger-ui`);
```
jpgough-ms: `fetch()` here has no `AbortController`/timeout. A stalled (not refused) connection blocks the awaited call indefinitely, so the `while (Date.now() < deadline)` check never gets re-evaluated — the function can hang well past its own `timeoutMs`, surfacing as a generic vitest hookTimeout instead of the intended "CalmHub not ready after Nms" message. Lower-probability than a plain ECONNREFUSED, but worth wiring `AbortSignal.timeout()` into the fetch.

cli/smoke/harness/hub-ready.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Polls the CalmHub Swagger UI until it responds 200 or the timeout elapses.
+ * /q/swagger-ui is available as soon as Quarkus finishes starting, regardless
+ * of DB profile (unlike /q/health/ready). Mirrors calm-hub/smoke-test.sh.
+ */
+export async function waitForHub(
```
jpgough-ms, line 6: This duplicates the readiness-poll loop already in `calm-hub/smoke-test.sh` (same rationale: `/q/health/ready` unreliable, poll `/q/swagger-ui` instead) — the doc comment even says it mirrors that script. Worth sharing the logic (or at least the endpoint/rationale) in one place so a future readiness-contract change doesn't require updating both independently.

cli/smoke/harness/fixtures.ts
```diff
@@ -0,0 +1,29 @@
...
+    namespace: string,
+    type: string,
+    mapping: string,
+    version: string,
+    baseUrl: string = SMOKE_HUB_URL
+): string {
+    return `${baseUrl}/calm/namespaces/${namespace}/${type}/${mapping}/versions/${version}`;
+}
+
+export function writeJson(filePath: string, obj: object): void {
```
jpgough-ms: `writeJson`/`readJson`/`patchJson` duplicate near-identical helpers already at the bottom of `cli/src/cli.e2e.spec.ts` (`mkdirSync`+`JSON.stringify` / `JSON.parse`+`readFileSync`). Since this PR already extracted `installPackedCli` into a shared `cli-runner.ts` and pointed `cli.e2e.spec.ts` at it, consider doing the same for these three functions to avoid two copies drifting.

cli/smoke/fixtures/genval/conference-signup.pattern.json
```diff
@@ -0,0 +1,463 @@
+{
```
jpgough-ms, line 1: This fixture is 463 lines (5 nodes, 4 relationships, per-node security controls, k8s metadata) but the 4 tests that use it only assert file existence, a non-empty `nodes` array, and `hasErrors === false` — none inspect node types, controls, the deployed-in relationship, or metadata. A much smaller fixture would drive the same generate/validate/push paths with less surface for drift. Not blocking, just flagging the maintenance cost.

jpgough-ms (commented):
Automated review of the CLI/CalmHub smoke test suite. One correctness bug in a test assertion, a coverage-exclude gap that's already at a 0.05-point margin in CI, a couple of error-handling gaps in the setup/teardown harness, and some reuse/simplification notes. Note: the `installPackedCli` duplication between `cli-runner.ts` and `cli.e2e.spec.ts` I initially flagged has already been fixed in the latest push (`01d354e`) — not repeating it here.

cli/smoke/hub-crud.smoke.spec.ts
willosborne: Good spot

cli/vitest.config.mts
willosborne: This is fixed by changes pointed out by Matt - to dedupe code

cli/smoke/global-setup.ts
willosborne: Good spot, fixed

cli/smoke/harness/hub-ready.ts
willosborne: Fixed

cli/smoke/harness/hub-ready.ts
willosborne: One is bash and one is TS - can't really share the code here without just shelling-out. IMO a readiness-contract should be an unlikely occurrence and I think keeping this readiness loop in both the setup script and the TS is good to avoid confusing failures

cli/smoke/harness/fixtures.ts
willosborne: fixed

cli/src/test_helpers/cli-runner.ts
willosborne: 1) fixed

2) this is a tradeoff of per-test isolation vs execution speed. currently the smoke tests are still a lot faster than the calmhub e2e suite, so this won't become the blocker. if it does get too slow we can implement this change.

cli/smoke/fixtures/genval/conference-signup.pattern.json
willosborne: i would argue this should not be a minimal example, but rather a more realistic pattern, given this is an e2e/smoke style test. 
however, this is going to be updated once node details validation goes in, so i can take a look at trimming it down later as part of that work.

jpgough-ms (approved):

PR #2788, opened 2026-07-07
feat(calm-hub): paginate namespace summary endpoints (limit/offset)

### LINKED ISSUE #2780: calm-hub: paginate/limit namespace summary endpoints for catalogue highlights
Follow-up from #2764 (part of #2754).

`useCatalogueHighlights` (the first-run landing's "Browse the catalogue" strip) fetches full namespace summary lists via `fetchArchitectureSummaries`/`fetchPatternSummaries` but keeps only `MAX_HIGHLIGHTS = 3` items. Those service methods — and the underlying `GET /calm/namespaces/{ns}/architectures|patterns` endpoints — have no `limit`/pagination parameter, so the whole catalogue is downloaded and parsed per probed namespace just to render up to 3 cards.

**Ask:** add a `limit` (and ideally offset/pagination) parameter to the namespace summary endpoints + the `CalmService` methods, and have `useCatalogueHighlights` request only what it renders.

Raised by @jpgough-ms in review of #2764. Currently bounded only by `MAX_NAMESPACES`, so the over-fetch is capped but not eliminated.

rocketstack-matt:
## Description
`useCatalogueHighlights` (the first-run "Browse the catalogue" strip) fetched the full
architecture/pattern summary lists per namespace via `CalmService` but kept only
`MAX_HIGHLIGHTS = 3` — the underlying `GET {ns}/architectures` and `{ns}/patterns` endpoints
had no `limit`, so the whole catalogue was downloaded to render up to 3 cards.

This adds an optional `limit` (and `offset`) query parameter to those namespace summary
endpoints and threads it through the store layer and the `CalmService` TS methods, and has
`useCatalogueHighlights` request only what it renders.

- **Mongo** limits at the query layer via a `$slice` projection (only when a `limit` is supplied).
  **Nitrite** has no array-slice projection, so it applies the window in memory via a shared
  `SummaryPagination` helper.
- `$slice` cannot express offset-without-limit and the hook only ever sends a limit, so the slice
  is applied only when `limit != null`; offset is honoured only alongside a limit.
- Fully backward-compatible: with no `limit`/`offset` the response body and request URL are unchanged.

Follow-up to #2764; closes #2780.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

TypeScript: `vitest run` → 1104 passed (99 files); lint 0 errors; build succeeds.
Java: `./mvnw -pl calm-hub -am clean install` → 2138 unit tests pass, JaCoCo ≥90%/class met, BUILD
SUCCESS. `MongoArchitectureIntegration` (integration profile) → 8 tests pass, including one proving
the Mongo `$slice` projection returns a sliced array against real MongoDB.

## Checklist
- [ ] I have updated documentation if necessary (n/a — the API self-documents via OpenAPI `@Operation`)


calm-hub-ui/src/service/calm-service.tsx
```diff
@@ -5,6 +5,21 @@ import { getAuthHeaders } from '../authService.js';
...
+ */
+function summaryPageParams(
+    limit?: number,
+    offset?: number
+): { limit?: number; offset?: number } | undefined {
+    const params: { limit?: number; offset?: number } = {};
+    if (limit !== undefined) params.limit = limit;
+    if (offset !== undefined) params.offset = offset;
+    return Object.keys(params).length > 0 ? params : undefined;
+}
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 0ecf44f1 — `summaryPageParams` now returns `undefined` when no `limit` is supplied (regardless of `offset`), and only includes `offset` alongside a `limit`. Added tests asserting an offset-only call sends no params, matching the backend contract.

calm-hub/src/main/java/org/finos/calm/store/util/SummaryPagination.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param <T>    the summary element type
+     * @return the requested window, or {@code items} unchanged when {@code limit} is {@code null}
+     */
+    public static <T> List<T> paginate(List<T> items, Integer limit, Integer offset) {
+        if (limit == null) {
+            return items;
+        }
+        int size = items.size();
+        int from = Math.min(offset == null ? 0 : Math.max(offset, 0), size);
+        int to = Math.min(from + Math.max(limit, 0), size);
```
jpgough-ms: `from + limit` overflows for large `limit` (no `@Max` bounds it) — e.g. `limit=2147483647, offset=1` wraps `to` negative and `items.subList(from, to)` throws `IllegalArgumentException`, surfacing as an unhandled 500 in standalone/Nitrite mode. Clamp with `Math.min(limit, size - from)` before adding, or use `long` arithmetic before casting back to `int`.

jpgough-ms (commented):
Automated review pass — 9 inline comments plus one off-diff note below.

**Off-diff note:** `NitriteArchitectureStore.getArchitecturesForNamespace` (~line 84) and the pattern-store equivalent still build a `NamespaceResourceSummary` for every document in the namespace before `SummaryPagination.paginate` trims the result. For `?limit=1` against a namespace with thousands of resources, this does full-list extraction work (including `VersionKeySelector.versionCount()` per item) only to discard nearly all of it. Consider slicing the raw document list to `[offset, offset+limit)` before the per-item extraction loop.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
```diff
@@ -72,11 +72,25 @@ public MongoArchitectureStore(MongoDatabase database, MongoCounterStore counterS
...
-        Document namespaceDocument = architectureCollection.find(Filters.eq("namespace", namespace)).first();
+        // When a limit is requested, push the slice down to Mongo via a $slice projection so only the
+        // requested window of the architectures array is returned rather than the whole list. No limit
+        // → no projection → full list (unchanged behaviour). $slice needs a limit, so an offset without
+        // a limit is not expressible here and is only honoured alongside a limit.
+        Bson filter = Filters.eq("namespace", namespace);
+        Document namespaceDocument = (limit == null)
+                ? architectureCollection.find(filter).first()
+                : architectureCollection.find(filter)
+                        .projection(Projections.slice("architectures", offset == null ? 0 : offset, limit))
```
jpgough-ms: Negative `offset` is passed straight into `Projections.slice`, whose Mongo semantics treat a negative skip as "count from the end of the array" — diverging from `SummaryPagination.paginate` (used by the Nitrite stores), which clamps negative offset to 0. Currently unreachable via REST (`@Min(0)` on the query param), but nothing in the store interface itself prevents a negative value reaching either backend inconsistently. Consider clamping here too, or validating at the interface boundary.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteArchitectureStore.java
```diff
@@ -90,8 +96,10 @@ public List<NamespaceResourceSummary> getArchitecturesForNamespace(String namesp
             architectureSummaries.add(summary);
         }
 
-        LOG.debug("Retrieved {} architectures for namespace '{}'", architectureSummaries.size(), namespace);
-        return architectureSummaries;
+        // Nitrite has no array-slice projection, so apply the limit/offset window in memory.
+        List<NamespaceResourceSummary> page = SummaryPagination.paginate(architectureSummaries, limit, offset);
+        LOG.debug("Retrieved {} architectures for namespace '{}'", page.size(), namespace);
```
jpgough-ms: This `LOG.debug` used to log the total match count (`architectureSummaries.size()`); it now logs the paginated page size instead. Any log-based monitoring inferring namespace size from this line will silently start reporting page size instead of total count whenever a caller supplies `limit` (e.g. the frontend's first-run highlights, which now always does). Consider logging both counts, e.g. `"Retrieved {} of {} architectures...", page.size(), architectureSummaries.size()`.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
```diff
@@ -100,8 +106,10 @@ public List<NamespaceResourceSummary> getPatternsForNamespace(String namespace)
             }
         }
 
-        LOG.debug("Retrieved {} patterns for namespace '{}'", patternSummaries.size(), namespace);
-        return patternSummaries;
+        // Nitrite has no array-slice projection, so apply the limit/offset window in memory.
+        List<NamespaceResourceSummary> page = SummaryPagination.paginate(patternSummaries, limit, offset);
+        LOG.debug("Retrieved {} patterns for namespace '{}'", page.size(), namespace);
```
jpgough-ms: Same issue as the architecture store: `LOG.debug` now reports the paginated page size instead of the total match count, silently changing what this log line means for anyone monitoring namespace size.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
```diff
@@ -72,11 +72,25 @@ public MongoArchitectureStore(MongoDatabase database, MongoCounterStore counterS
...
         if (!namespaceStore.namespaceExists(namespace)) {
             throw new NamespaceNotFoundException();
         }
 
-        Document namespaceDocument = architectureCollection.find(Filters.eq("namespace", namespace)).first();
+        // When a limit is requested, push the slice down to Mongo via a $slice projection so only the
+        // requested window of the architectures array is returned rather than the whole list. No limit
+        // → no projection → full list (unchanged behaviour). $slice needs a limit, so an offset without
+        // a limit is not expressible here and is only honoured alongside a limit.
+        Bson filter = Filters.eq("namespace", namespace);
```
jpgough-ms, line 83: The `$slice`-projection ternary (and its comment) is duplicated verbatim in `MongoPatternStore`, and the `@QueryParam`/`@Min` annotation block is duplicated between `ArchitectureResource` and `PatternResource`. A shared helper — e.g. a `findWithOptionalSlice(collection, filter, arrayField, limit, offset)` method, plus a `@BeanParam` pagination object for the two resources — would keep the four copies in sync as the contract evolves.

calm-hub-ui/src/service/calm-service.tsx
```diff
@@ -5,6 +5,29 @@ import { getAuthHeaders } from '../authService.js';
...
+/**
+ * Build the optional `limit`/`offset` query params for the namespace summary endpoints.
+ * Returns `undefined` when no `limit` is supplied so the request URL is unchanged (and the
+ * backend returns the full list) — preserving backward-compatible behaviour.
+ *
+ * `offset` is only included alongside a `limit`: the backend applies it only with a limit
+ * ($slice can't express offset-without-limit) and ignores an offset-only request, so sending
+ * one would be a silent no-op that misleads callers into thinking paging is happening.
+ */
+function summaryPageParams(
```
jpgough-ms: `summaryPageParams()` solves "only include defined query params" with a plain object and manual undefined-checks, while `fetchDecoratorValues` (line 354, same file) already solves the identical problem via `URLSearchParams`. Two conventions for the same problem in one file — consider reusing the existing pattern instead.

calm-hub/src/main/java/org/finos/calm/store/ArchitectureStore.java
```diff
@@ -11,6 +11,16 @@
...
     List<NamespaceResourceSummary> getArchitecturesForNamespace(String namespace) throws NamespaceNotFoundException;
+
+    /**
+     * Retrieve architecture summaries for a namespace, optionally limited and offset.
+     *
+     * @param namespace the namespace to retrieve architectures for
+     * @param limit     the maximum number of summaries to return, or {@code null} for no limit
+     * @param offset    the number of summaries to skip, or {@code null} for none
+     * @return a (possibly limited) list of architecture summaries
+     */
```
jpgough-ms, line 30: Pagination is added as two positional `Integer` parameters plus a second overloaded method, rather than a single `PageRequest`/`Pagination` value object. The next query capability (sort, filter, cursor paging) will force another overload or parameter across both interfaces, both backend implementations, and every call site/mock — this diff's shape repeats each time. Worth introducing a `PageRequest(limit, offset, ...)` type now while there's only one capability to migrate.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteArchitectureStoreShould.java
```diff
@@ -173,6 +173,57 @@ public void testGetArchitecturesForNamespace_whenArchitecturesExist_returnsArchi
         verify(mockNamespaceStore, atLeastOnce()).namespaceExists(NAMESPACE);
     }
 
+    @Test
+    public void testGetArchitecturesForNamespace_appliesLimitAndOffsetInMemory() throws NamespaceNotFoundException {
```
jpgough-ms: New test methods use camelCase (`testGetArchitecturesForNamespace_appliesLimitAndOffsetInMemory`); `calm-hub/AGENTS.md` requires `snake_case` test method names. Matches this file's pre-existing style rather than introducing a new deviation, but worth aligning with the newly-added tests in `TestArchitectureResourceShould.java` in this same PR, which do follow the convention.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitritePatternStoreShould.java
```diff
@@ -123,6 +123,53 @@ public void testGetPatternsForNamespace_whenPatternsExist_returnsPatternSummarie
         assertThat(result.get(1).getVersionCount(), is(1));
     }
 
+    @Test
+    public void testGetPatternsForNamespace_appliesLimitAndOffsetInMemory() throws NamespaceNotFoundException {
```
jpgough-ms: Same naming-convention gap as the architecture-store test file: new camelCase test methods against a `snake_case` requirement in `calm-hub/AGENTS.md`.

calm-hub/src/main/java/org/finos/calm/store/util/SummaryPagination.java
rocketstack-matt: Fixed in 22f3a13d — the window now lives in `PageRequest.apply`, which computes the upper bound with `long` arithmetic (`from + max(limit, 0L)`) before clamping to size, so a huge `limit` returns a clamped window instead of throwing. New overflow test covers it.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
rocketstack-matt: Fixed in 22f3a13d — both Mongo stores now go through `MongoResourceSlice.findNamespaceDoc`, which uses `PageRequest.normalizedOffset()` (clamps negative → 0) for the `$slice`, matching the Nitrite path. Tests assert `slice(field, 0, n)` for a negative offset.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteArchitectureStore.java
rocketstack-matt: Fixed in 22f3a13d — now logs `"Retrieved {} of {} architectures…"` (page size + total match count).

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
rocketstack-matt: Fixed in 22f3a13d — same for patterns: logs page-of-total now.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoArchitectureStore.java
rocketstack-matt: Fixed in 22f3a13d — extracted `MongoResourceSlice.findNamespaceDoc` (shared `$slice`/plain-find) and a `@BeanParam PaginationQueryParams` bean carrying the `@Min` annotations + `toPageRequest()`; both Mongo stores and both resources use them, so the four copies are gone.

calm-hub-ui/src/service/calm-service.tsx
rocketstack-matt: Fixed in f572ee3f — `summaryPageParams` is now `summaryPageQuery`, building the query string via `URLSearchParams` like `fetchDecoratorValues`; the tests assert the request URL. Behaviour preserved (no params without a limit; offset only alongside a limit).

calm-hub/src/main/java/org/finos/calm/store/ArchitectureStore.java
rocketstack-matt: Done in 22f3a13d — introduced a `PageRequest(limit, offset)` record; the store interfaces now take `(namespace, PageRequest)`, with the no-arg kept as a `default` delegating to `PageRequest.UNPAGED` so the MCP tools / CountsService callers are untouched. A future sort/cursor capability extends `PageRequest` rather than adding another overload.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteArchitectureStoreShould.java
rocketstack-matt: Fixed in 22f3a13d — the new Nitrite store test methods are renamed to snake_case (matching `calm-hub/AGENTS.md` and the resource tests).

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitritePatternStoreShould.java
rocketstack-matt: Fixed in 22f3a13d — same snake_case rename for the pattern store tests.

jpgough-ms (approved):

Issue #2770, opened 2026-07-02
Standardize CALM document type list in calm-models


willosborne:
## Problem

The set of CALM document types (`architecture`, `pattern`, `flow`, etc.) is currently defined/duplicated independently in several places across the monorepo instead of coming from a single shared source of truth in `calm-models`. For example:

- `CalmDocumentType` union / `CALM_DOCUMENT_TYPES_LIST` in `shared/src/document-loader/document-loader.ts`
- Similar type lists referenced in `shared/src/commands/validate/validation-enrichment.ts` and `shared/src/commands/diff/diff.ts`
- CLI usage in `cli/src/cli.ts` and `cli/src/command-helpers/diff.ts`
- Hub UI navigation loaders (`calm-hub-ui/src/hub/components/tree-navigation/navigation-loaders.ts`)

Each of these maintains its own list/union of valid document types, which is duplicative and risks drifting out of sync — e.g. adding a new document type (or renaming one) requires hunting down and updating every copy, and there's no guarantee they all agree.

## Proposal

Standardize on a single canonical list/type of CALM document types defined once in `calm-models`, and have all other packages (`shared`, `cli`, `calm-hub-ui`, etc.) import and reuse it rather than redeclaring their own lists.

## Acceptance Criteria

- [ ] A single source of truth for CALM document types exists in `calm-models`
- [ ] Duplicate lists/unions elsewhere in the monorepo are removed and replaced with imports of the shared definition
- [ ] No behavior change for existing document type handling

Issue #2716, opened 2026-06-22
Validate uploaded documents in CalmHub via CALM server sidecar


willosborne:
### Target Project:
CalmHub

### Description of Feature:
Add a configurable endpoint to CalmHub for an instance of the calm server. That will be used to validate all uploaded architectures, patterns and so on.
Also include CLI server in CalmHub image as a sidecar by default.

### User Stories:
As an architect, I want a guarantee that my patterns and architectures in CalmHub are valid.

### Current Limitations:
No validation in the Java framework.

### Proposed Implementation:
- CalmHub validation service that calls calm server
- Call this on pattern, architecture, control, etc. upload
- Use the $schema if present to validate these documents
- Include calm server binary in docker image and start it alongside calmhub
- Possible escape hatch header to skip validation eg X-CalmHub-Skip-Validation: true

### Alternatives Considered:
- Validation library in java - rejected as it means a complete rewrite of all CLI logic in Java + ongoing maintenance
- Validate in CLI - this is a soft guarantee as users can still use API endpoints directly to upload invalid architectures to CalmHub

### Testing Strategy:
- Integration testing against real calm server
- Unit testing with mocked calm server responses

### Documentation Requirements:
Docs on how to configure calmhub to use this server in case folks can't use default docker image

### Implementation Checklist:
- [ ] Design reviewed and approved
- [ ] Implementation completed
- [ ] Tests written and passing
- [ ] Documentation updated
- [ ] Relevant workflows updated (if needed)
- [ ] Performance impact assessed

### Additional Context:
- #689 — Expose the CALM CLI as a RESTful API (origin of the "CLI server as a local sidecar" pattern).
- #2051 — CALM Server: extracted the validation server into the standalone `@finos/calm-server` package (the server this proposal would run as a sidecar).

jpgough-ms:
@willosborne 

> Also include CLI server in CalmHub image as a sidecar by default.

I don't think we need to include it in the image? Would using `docker-compose` be better here. I'm thinking from a security perspective I may not want that layer available and exposed

Other than that I am good with this issue.

willosborne:
Good point. Guess that means we need another image to publish for the calm server, if we don't already. But I can absolutely see value in publishing that as a standalone component @jpgough-ms

jpgough-ms:
I think it makes sense to publish calm server as a docker image

LeighFinegold:
[...]

Issue #2566, opened 2026-06-04
Questions on CALM Validations


chandralanka:
# Questions on CALM Validations
We are working with both standard CALM validations and custom validations we've built internally for our organization's specific requirements. Our goal is to enforce both sets of validations at two points in the development workflow: when a developer checks in code to a feature branch, and when a pull request is merged.
We have a few questions on this:

**Validation Server & Custom Extensions** — CALM includes a validation server, but we'd like to understand how we can integrate our custom validations into it. Is extending the validation server with custom rules something that is currently supported, or is it on the roadmap?
**Recommendations** — Are there any best practices or architectural recommendations for applying CALM validations (both standard and custom) at the feature branch check-in and PR merge stages?

LeighFinegold:
When you say CALM exposes Validation Server, are you referring to the Calm Studio product? 

The CLI exposes `calm validate` which could be hooked into to run on your calm files as part of the build process. 
https://github.com/finos/architecture-as-code/issues/2360 was raised as a request for future enhancement to expose default github actions to do such things.

In relation to custom validations, @willosborne, @jpgough-ms and I have talked about how calm validation can be evolved whereby users can supply their own custom rules. That said  `calm validate` would be enhanced to take into account standards you want your architecture to adhere to.

@willosborne - probably worth us typing up a technical visual on how we want to show how calm validate is being designed to evolve.

willosborne:
Absolutely. Custom validations has come up before, and there is definitely a lot of interest. 
At the moment, custom validations beyond just the JSON Schema structural checks are done via Spectral. We're proposing a rework of some of these rules to use a better, more structured validation system. But I think it's very likely we'll leave Spectral in the suite of checks.

Potential extensibility could be
- extra Spectral rules, operating direct on the json
- extra validation rules that operate on the parsed calm model, once we figure out how to make that work for the default validations.

I think it's likely that the extension point will be a plugin - a JS plugin - that's configured via env variables or your ~/.calm.json (or a startup config object in the case of the CLI server). This is the approach we're taking for authentication.

rocketstack-matt:
I assume the question relates to https://github.com/finos/architecture-as-code/tree/main/calm-server

rocketstack-matt:
Closing — the questions were answered above (custom validations today are done via Spectral, with a rework of the validation system in progress, and calm-server covers the server-mode workflow). Feel free to reopen or raise a new issue if anything is still unclear.

PR #2790, opened 2026-07-08
feat(shared): Validation Engine Redesign with articulated assumptions

LeighFinegold:
## Description

Follow-up to @willosborne's PR #2778 (finos/architecture-as-code#2778). I believe this starts to
address the concerns raised on that PR and the assumptions we have been making about the validation
framework, by reworking recursive detailed-architecture and control validation into a cycle-safe,
resolver-driven design.

Summary of the design (see
[`shared/src/commands/validate/ValidationTechnicalSpecification.md`](shared/src/commands/validate/ValidationTechnicalSpecification.md)
for detail):

- Validation phases are modelled as first-class `ValidationRule`s run by a `ValidationEngine`, keeping
  the existing `ValidationOutcome` contract (output buckets, ordering, error/warning flags).
- A single cycle-safe, resolver-driven `ModelWalker` owns model traversal, dereferencing and cycle
  detection. This fixes the pattern-less dispatch branch dropping the `visited` set (the self-reference
  heap OOM flagged on #2778).
- Reference resolution is driven through the model's `Resolvable` / `CalmReferenceResolver` seam, with
  a `CachingTrackingResolver` giving fetch-at-most-once and validate-once/cycle-safety in one place.
- Node-details and control validation reuse that seam; error/message helpers are consolidated onto the
  shared `getErrorMessage`.
- `ValidationTechnicalSpecification.md` also records the open questions the framework does not yet
  settle: default recursive traversal vs a caller-selected resolution policy (same always-on concern
  raised on #2778), custom/organisation-specific rules per document, a typed CALM document input
  instead of four positional params, and how validation should differ between the CLI and CalmHub.

This is intended as a discussion PR: it restates the validation design and surfaces the remaining
unknowns rather than claiming to close them.

## Design detail (from the Validation Technical Specification)

> The full write-up lives in
> [`shared/src/commands/validate/ValidationTechnicalSpecification.md`](shared/src/commands/validate/ValidationTechnicalSpecification.md).
> The key parts are inlined below so the design is reviewable directly on the PR.

### Reference resolution, caching and cycle safety

The CALM model is *lazily* dereferenced: a `ResolvableAndAdaptable<CalmCoreSchema, CalmCore>` only
calls `resolve(ref)` when a caller chooses to dereference it, and the resolve function returns the
**raw** `CalmCoreSchema` before it is adapted to `CalmCore`. Node-details validation needs that raw
document (for Spectral + JSON-Schema), so the resolve seam is the natural place to centralise:

- **caching**: a reference is fetched at most once; repeat requests return the cached raw document;
- **tracking**: every attempted reference is recorded, giving dedupe ("validate each
  detailed-architecture once") and cycle safety without each caller keeping its own `Set`.

`CachingTrackingResolver` is a **decorator** that `implements CalmReferenceResolver`, wrapping any
inner resolver. This is the single resolver interface used by both the model dereference path and
validation:

```mermaid
classDiagram
    class CalmReferenceResolver {
        <<interface>>
        +canResolve(ref) boolean
        +resolve(ref) Promise
    }
    class CachingTrackingResolver {
        -delegate: CalmReferenceResolver
        -cache: Map~string, unknown~
        -seen: Set~string~
        +canResolve(ref) boolean
        +resolve(ref) Promise
        +has(ref) boolean
        +get(ref) unknown
        +markSeen(ref) void
        +resolvedReferences: ReadonlySet~string~
    }
    class SchemaDirectoryReferenceResolver {
        -schemaDirectory: SchemaDirectory
        +canResolve(ref) boolean
        +resolve(ref) Promise
    }
    CalmReferenceResolver <|.. CachingTrackingResolver
    CalmReferenceResolver <|.. SchemaDirectoryReferenceResolver
    CachingTrackingResolver o-- CalmReferenceResolver : decorates
```

A reference is marked *seen* before the underlying load is awaited, so a failed load still counts as
seen (a sibling referencing the same failing URL is not retried), matching the `visitedUrls.add`
-before-load behaviour originally implemented in PR #2778
(finos/architecture-as-code#2778). Only successful loads populate the value cache. In validation the
decorator wraps a `SchemaDirectoryReferenceResolver` (which loads architecture documents via
`schemaDirectory.loadDocument(ref, 'architecture')`); in docify the **caller** (`TemplateProcessor`)
composes the decorator around its resolver (`Mapped` -> `Composite`) so repeated references are
fetched once. Both `DereferencingVisitor` and `ModelWalker` are agnostic: they dereference through
whatever `CalmReferenceResolver` they are given; caching/tracking is purely the caller's composition
choice.

### Resolution policy (design intent)

Whether a given reference is resolvable is governed by a `ReferenceResolutionPolicy`, consulted in
`canResolve`. The policy is a constructor input to the resolver adapter; the `CalmReferenceResolver`
interface is unchanged. Two variants:

- **shallow (default):** `canResolve` returns `false` for `detailed-architecture` document
  references, so a shallow run never fetches or recurses into sub-architectures.
- **deep (opt-in):** `canResolve` allows those references, so the node-details rule resolves and
  validates each referenced sub-architecture.

Because only the node-details rule uses `ValidationContext.references` (schemas load via
`SchemaDirectory.getSchema` directly), the policy affects recursion alone. Child validation contexts
reuse the same resolver, so the policy is inherited at every level of recursion. This is design
intent; the policy type, its resolver wiring and the entry-point option are pending implementation.

### Open questions

These are recorded in the specification as design intent, not settled behaviour:

- **Default traversal and resolution policies.** By default, validation recurses into a node's
  `detailed-architecture`. This rewrite still suffers from the same always-on concern raised on
  finos/architecture-as-code#2778; making it a caller-selected resolution policy is worth considering.
- **Custom rules versus the default traversal.** Organisation-specific rules may apply only to
  specific documents or document types, which the generic traversal does not model. It is an open
  question how a caller such as CalmHub tells the calm-server which rules to run, and how differing
  rule sets across a large organisation are catered for (see
  finos/architecture-as-code#2716 and #2566).
- **Typed CALM document input.** `validate()` currently takes four loosely-typed positional inputs and
  infers a mode; a single typed CALM document from `calm-models` (finos/architecture-as-code#2770)
  would be cleaner.
- **CLI versus CalmHub.** The CLI is user-driven and would default to recursive (deep) traversal;
  CalmHub validates on upload and would decide when a deep run is performed.

### DIFF
--- a/calm-models/src/model/resolvable.ts
+++ b/calm-models/src/model/resolvable.ts
@@ -1,3 +1,11 @@
+/**
+ * A resolvable node in the CALM model — either a plain {@link Resolvable} or a
+ * {@link ResolvableAndAdaptable}. Both expose `reference`, `isResolved` and `value`.
+ */
+export type AnyResolvable =
+    | Resolvable<unknown>
+    | ResolvableAndAdaptable<unknown, unknown>;
+
 export class Resolvable<T> {
     constructor(
         public reference: string,
--- a/shared/src/commands/validate/ValidationTechnicalSpecification.md
+++ b/shared/src/commands/validate/ValidationTechnicalSpecification.md
@@ -1,0 +1,333 @@
+# Validation Technical Specification
+
+Technical design for CALM document validation in `shared` (`@finos/calm-models` model +
+`shared/src/commands/validate`). Validation is a phased pipeline built from **`ValidationRule`**
+units executed by a **`ValidationEngine`**. `validate(...)` builds a `ValidationContext` and calls
+`ValidationEngine.validate`, which runs the registered rules and aggregates their results into a
+single `ValidationOutcome`. Rules come in two families over a shared context: Spectral-backed
+(JSONPath over raw JSON) and model/schema-backed (AJV / typed `CalmCore`). Recursive resolution of a
+node's `detailed-architecture` is opt-in and resolver-controlled: it is governed by a reference
+resolution policy that defaults to shallow, so top-level validation never implicitly fetches
+sub-architectures.
+
+## 1. Objectives
+
+- **O1: One authoritative validator in `shared`.** Validation is a library concern, not a CLI
+  concern. The CLI, CALM Hub (upload), CalmStudio, and any future tool call the same
+  `validate(...)` entry point so guarantees cannot be bypassed by hitting an API directly.
+- **O2: Phased validation.** Run cheap/broad linting (Spectral) and structural schema validation
+  (JSON Schema) as distinct phases, plus semantic/model checks (controls, node-details recursion).
+  A failure in one phase still lets other phases report, so users get a complete picture.
+- **O3: Uniform, machine-readable output.** Every phase emits `ValidationOutput` items aggregated
+  into a single `ValidationOutcome` with stable `code`, `severity`, JSON-pointer `path`, and
+  `source`. Downstream (JUnit/JSON/pretty formatters, Hub UI) depends on this shape.
+- **O4: Cycle- and reference-safe traversal.** Detailed architectures and control configs are
+  references that can form cycles; traversal must terminate and must not re-fetch or re-adapt more
+  than necessary.
+- **O5: Extensibility.** Adding a new check is registering a new `ValidationRule` with the engine;
+  the entry point does not change.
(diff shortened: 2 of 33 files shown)


LeighFinegold:
Meant to raise on local to start off with

Issue #2806, opened 2026-07-14
feat(calm-hub-ui): real thumbnails for architecture/pattern cards, type icons for non-visual resources


rocketstack-matt:
### Target Project:
`calm-hub-ui`, with supporting changes to `calm-hub` and `calm-server`.

### Description of Feature:
Replace the placeholder card backgrounds used across CALM Hub's browse/explore surfaces with content that actually represents each resource:

- **Visual assets (architectures, patterns):** a real thumbnail image of the resource's diagram, rendered the same way it looks in the Hub's own visualizer, instead of today's striped gradient placeholder.
- **Non-visual assets (flows, standards, ADRs, interfaces):** a distinguishing type icon on the card, matching the pattern Controls already use, plus their description shown where it isn't already (flows/standards/interfaces already show one; ADRs currently have none — closing that gap is left as a follow-up to avoid a backend schema change here).

### User Stories:
- As a CALM Hub user browsing a namespace, I want to see a recognizable thumbnail of an architecture or pattern's diagram on its card, so I can identify it without opening it.
- As a CALM Hub user browsing flows, standards, ADRs or interfaces, I want a distinguishing type icon on the card (as Controls already have), so I can tell resource types apart at a glance in a mixed grid or search-result list.

### Current Limitations:
- Every architecture/pattern card shows the same per-type-colored striped `repeating-linear-gradient` background (`calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx`) — there is no thumbnail/image field anywhere in the data model (frontend or backend), so no per-item visual signal exists today.
- Only Controls render a type icon (`IoShieldCheckmarkOutline`, hardcoded in `calm-hub-ui/src/hub/components/domain-page/ControlCard.tsx`) via `ItemCard`'s existing `thumbnailIcon` slot. Flows, standards, ADRs and interfaces render no icon.
- Flows, standards and interfaces already show a description on their cards; ADRs do not, because neither the frontend `AdrSummary` nor the backend `NamespaceAdrSummary` carry a description field at all today (the full ADR document does have `contextAndProblemStatement`, but it's never surfaced through the summary endpoint the browse grid uses).

### Proposed Implementation:

**Phase 1 — Type icons (frontend only, no schema changes)**
[...]

PR #2809, opened 2026-07-14
feat(calm-hub): diagram thumbnails and type icons on browse cards

rocketstack-matt:
## Description

Implements the first two phases of #2806: real diagram thumbnails on architecture/pattern browse cards, and type icons for the non-visual resource types.

**Commit 1 — resource type icons (`calm-hub-ui` only).** A central registry (`src/theme/resource-type-meta.ts`) now maps each resource type to its icon alongside the existing label/colour metadata: Flows, Standards, ADRs and Interfaces get a type glyph on their browse cards (the pattern Controls already used — its hardcoded shield migrated into the registry), and every search surface (navbar dropdown, explorer, intro search, full results page) shows the icon on its type-group headers.

**Commit 2 — diagram thumbnails rendered via calm-server.**
- **calm-server** gains an internal `POST /calm/render/thumbnail`: it drives a local headless Chrome (via `@finos/calm-shared`'s existing `launchBrowser`) against the Hub UI's new chrome-free `/#/render` route, injects the document with `addInitScript`, waits for the diagram's `data-render-ready` signal, and returns a PNG clipped to the graph's content bounds. The clip is expanded toward the browse card's aspect ratio so the card's `object-cover` never crops away nodes — a two-node architecture shows both nodes, not just the edge between them.
- **calm-hub** stores thumbnails base64-per-version in both Nitrite and Mongo stores and serves them via `GET .../versions/{version}/thumbnail` and a latest-version `GET .../{id}/thumbnail` (`image/png`, private cache). Rendering is triggered asynchronously on every architecture/pattern write — a render failure never affects the write — and a GET for a missing thumbnail attempts a synchronous single-flight render (with a short failure cache) before 404ing, so failed renders self-heal on next browse. The whole pipeline is **disabled unless `calm.render.service-url` is configured**, and requires a deployment where the UI and thumbnail GETs are reachable without interactive auth (standalone / no-auth / proxy-auth / public-read) — documented in `application.properties` and `calm-hub/AGENTS.md`.
- **calm-hub-ui** adds the chromeless `/#/render` route, `ItemCard` thumbnail support with graceful fallback to the existing striped placeholder, capped-width auto-fill card grids (namespace, domain), and the search results page now renders the same browse cards.
- **Bundled fix:** the SPA shell (`/`, `/index.html`) was served `public, immutable, max-age=86400`, pinning browsers to a stale bundle for up to a day after every upgrade; it is now `no-cache` while hashed `/assets/*` keep immutable caching.

Open follow-ups from #2806 deliberately not in this PR: ADR descriptions (needs a backend schema change), auth-aware thumbnail loading for secure OIDC deployments, and thumbnail backfill for existing versions (they render on first browse via the self-healing GET).

Note on `package-lock.json`: beyond the `playwright-core` addition for calm-server, the diff includes lockfile sync of pre-existing drift (cli/calm-server version fields already bumped by release commits) and removal of stale extraneous entries.

### Screenshots

**Architecture cards with rendered thumbnails (namespace browse):**
architecture cards with diagram thumbnails

**Type icons on non-visual resource cards (flows):**
flow cards with type icons

**Search results using the shared browse cards:**
search results rendered as browse cards

**Small architectures render minimap-style (aspect-bounded clip):**
two-node architectures showing both nodes in thumbnails

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [x] CALM Server (`calm-server/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Testing
- [x] All existing tests pass

Verified end-to-end against a local standalone deployment (calm-hub + calm-server + system Chrome): eager render on write, self-healing render on GET-miss with single-flight dedup and failure caching, graceful degradation with calm-server down (writes succeed, cards fall back to the striped placeholder), and thumbnail regeneration via PUT updates. Suites: calm-hub-ui 1248 tests, calm-server 73 tests (coverage gate passing), shared 963 tests, calm-hub 2199 tests.

## Checklist

### DIFF
--- a/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
+++ b/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
@@ -1,9 +1,16 @@
-import { useState, type ReactNode } from 'react';
+import { useId, useState, type ReactNode } from 'react';
 import { Link } from 'react-router-dom';
 import { colors } from '../../../theme/colors.js';
 import { redesignTokens } from '../../../theme/redesign-tokens.js';
 import { TypeBadge } from './TypeBadge.js';
 import { type CardResourceType, getResourceTypeColors, getResourceTypeIcon } from '../../../theme/resource-type-meta.js';
+
+/**
+ * Shared responsive card grid for the browse surfaces (namespace, domain, search
+ * results). `min(280px,100%)` keeps a single column from overflowing containers
+ * narrower than the 280px card minimum.
+ */
+export const CARD_GRID_CLASS = 'grid gap-4 grid-cols-[repeat(auto-fill,minmax(min(280px,100%),1fr))]';
 
 interface ItemCardProps {
     name: string;
@@ -23,7 +30,13 @@ interface ItemCardProps {
      * When unset the footer falls back to the "N versions" scent or {@link customId}.
      */
     meta?: string;
-    /** Thumbnail header height in px. The tall default gives rendered diagram thumbnails enough vertical space that the cover-fit crop stays modest. */
+    /**
+     * Thumbnail header height in px. An explicit value always wins; otherwise the
+     * default is 160 when {@link thumbnailUrl} is set (rendered diagram thumbnails
+     * need the vertical space so the cover-fit crop stays modest) and 96 when not
+     * (icon/stripe-only headers shouldn't burn it). Every browse surface shows one
+     * type at a time, so the per-type default can't misalign a row.
+     */
     thumbnailHeight?: number;
     /**
      * URL of a rendered diagram thumbnail (architectures/patterns). When set, an
@@ -74,7 +87,7 @@ export function ItemCard({
     customId,
     versionCount,
     meta,
-    thumbnailHeight = 160,
+    thumbnailHeight,
     thumbnailUrl,
     thumbnailIcon,
     active,
@@ -87,6 +100,10 @@ export function ItemCard({
     // changed thumbnailUrl naturally retries without an effect to reset state.
     const [failedThumbnailUrl, setFailedThumbnailUrl] = useState<string | undefined>(undefined);
     const showThumbnailImage = thumbnailUrl !== undefined && thumbnailUrl !== failedThumbnailUrl;
+
+    // Explicit prop wins; otherwise diagram-thumbnail cards get the tall header and
+    // icon/stripe-only cards the short one (see the thumbnailHeight prop JSDoc).
+    const headerHeight = thumbnailHeight ?? (thumbnailUrl !== undefined ? 160 : 96);
 
     const { accent, accentText, tint } = getResourceTypeColors(type);
     // Registry-derived type glyph for the thumbnail; an explicit thumbnailIcon
@@ -129,6 +146,16 @@ export function ItemCard({
               ? `${versionCount} ${versionCount === 1 ? 'version' : 'versions'}`
               : customId;
 
+    // The activation element's accessible name is just the item name; the description
+    // paragraph and footer chip become its accessible description via aria-describedby
+    // so screen readers announce them with the control.
+    const descriptionId = useId();
+    const chipId = useId();
+    const describedByIds = [description ? descriptionId : undefined, chip !== undefined ? chipId : undefined]
+        .filter(Boolean)
+        .join(' ');
+    const describedBy = describedByIds === '' ? undefined : describedByIds;
+
     return (
         <article
             className="group relative rounded-[12px] overflow-hidden bg-base-100 hover:-translate-y-0.5 hover:shadow-md"
@@ -142,7 +169,7 @@ export function ItemCard({
         >
             <div
                 className={!showThumbnailImage && headerIcon ? 'flex items-center justify-center' : undefined}
-                style={{ height: thumbnailHeight, background: stripes }}
+                style={{ height: headerHeight, background: stripes }}
             >
                 {showThumbnailImage ? (
                     // Rendered diagram thumbnail on top of the stripe background; on
@@ -166,7 +193,13 @@ export function ItemCard({
             </div>
             <div className="p-[14px]">
                 {href ? (
-                    <Link to={href} data-testid={testId} className={activationClass} style={{ color: colors.redesign.ink }}>
+                    <Link
+                        to={href}
+                        data-testid={testId}
+                        aria-describedby={describedBy}
+                        className={activationClass}
+                        style={{ color: colors.redesign.ink }}
+                    >
                         {name}
                     </Link>
                 ) : (
@@ -176,6 +209,7 @@ export function ItemCard({
                         // Only forwarded when `active` is defined, so plain browse cards
                         // (which don't pass it) carry no aria-pressed and aren't toggles.
                         aria-pressed={active}
+                        aria-describedby={describedBy}
                         onClick={onActivate}
                         className={activationClass}
                         style={{ color: colors.redesign.ink }}
@@ -185,6 +219,7 @@ export function ItemCard({
                 )}
                 {description && (
                     <p
+                        id={descriptionId}
                         className="text-[12px] leading-[1.45] mt-[5px] mb-[11px] line-clamp-2"
                         style={{ color: colors.redesign.mutedAlt }}
                     >
@@ -197,6 +232,7 @@ export function ItemCard({
                     <TypeBadge type={type} />
                     {chip !== undefined && (
                         <span
+                            id={chipId}
                             className="font-mono-jb text-[10.5px] ml-auto truncate"
                             style={{ color: colors.redesign.mutedAlt }}
                         >
--- a/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
+++ b/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
@@ -3,7 +3,7 @@ import { Link } from 'react-router-dom';
 import { colors } from '../../../theme/colors.js';
 import { redesignTokens } from '../../../theme/redesign-tokens.js';
 import { TypeBadge } from './TypeBadge.js';
-import { type CardResourceType, getResourceTypeColors } from './resource-type-meta.js';
+import { type CardResourceType, getResourceTypeColors, getResourceTypeIcon } from '../../../theme/resource-type-meta.js';
 
 interface ItemCardProps {
     name: string;
@@ -26,8 +26,10 @@ interface ItemCardProps {
     /** Thumbnail header height in px (default 96; landing highlights use a shorter one). */
     thumbnailHeight?: number;
     /**
-     * Optional glyph centred in the striped thumbnail (e.g. a shield for a control).
-     * Namespace cards leave the header a plain stripe; only control-style cards set it.
+     * Overrides the glyph centred in the striped thumbnail. Pass a non-null node
+     * to override; when unset or `null` the type's registry icon
+     * ({@link getResourceTypeIcon}) is used, and types without one
+     * (Architectures / Patterns) keep the plain stripe.
      */
     thumbnailIcon?: ReactNode;
     /**
@@ -72,7 +74,21 @@ export function ItemCard({
     href,
     onActivate,
 }: ItemCardProps) {
-    const { accent, tint } = getResourceTypeColors(type);
+    const { accent, accentText, tint } = getResourceTypeColors(type);
+    // Registry-derived type glyph for the thumbnail; an explicit thumbnailIcon
+    // prop still wins so callers can override the header treatment.
+    const TypeIcon = getResourceTypeIcon(type);
+    const headerIcon =
+        thumbnailIcon ??
+        (TypeIcon ? (
+            <TypeIcon
+                size={30}
+                data-testid="thumbnail-type-icon"
+                // The type's own accent in its text/icon role, so the glyph
+                // can't drift if the type colours are adjusted independently.
+                style={{ color: accentText, opacity: 0.55 }}
+            />
+        ) : undefined);
     // Selected treatment mirrors the diagram's selected-node style: an interaction-blue
     // outline + elevated shadow while the item's detail panel is open. Uses the brand
     // interaction blue (not the type accent) so selection reads consistently. Only the
@@ -111,10 +127,10 @@ export function ItemCard({
             }}
         >
             <div
-                className={thumbnailIcon ? 'flex items-center justify-center' : undefined}
+                className={headerIcon ? 'flex items-center justify-center' : undefined}
                 style={{ height: thumbnailHeight, background: stripes }}
             >
-                {thumbnailIcon}
+                {headerIcon}
             </div>
             <div className="p-[14px]">
                 {href ? (
--- a/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
+++ b/calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
@@ -1,4 +1,4 @@
-import { type ReactNode } from 'react';
+import { useState, type ReactNode } from 'react';
 import { Link } from 'react-router-dom';
 import { colors } from '../../../theme/colors.js';
 import { redesignTokens } from '../../../theme/redesign-tokens.js';
@@ -23,8 +23,15 @@ interface ItemCardProps {
      * When unset the footer falls back to the "N versions" scent or {@link customId}.
      */
     meta?: string;
-    /** Thumbnail header height in px (default 96; landing highlights use a shorter one). */
+    /** Thumbnail header height in px. The tall default gives rendered diagram thumbnails enough vertical space that the cover-fit crop stays modest. */
     thumbnailHeight?: number;
+    /**
+     * URL of a rendered diagram thumbnail (architectures/patterns). When set, an
+     * `<img>` fills the thumbnail header on top of the stripe background; if the
+     * image fails to load (e.g. no thumbnail rendered yet — the endpoint 404s)
+     * the header falls back to exactly the stripe(+icon) rendering.
+     */
+    thumbnailUrl?: string;
     /**
      * Overrides the glyph centred in the striped thumbnail. Pass a non-null node
      * to override; when unset or `null` the type's registry icon
@@ -67,13 +74,20 @@ export function ItemCard({
     customId,
     versionCount,
     meta,
-    thumbnailHeight = 96,
+    thumbnailHeight = 160,
+    thumbnailUrl,
     thumbnailIcon,
     active,
     testId = 'item-card',
     href,
     onActivate,
 }: ItemCardProps) {
+    // Tracks the thumbnail URL that failed to load, so the header falls back to
+    // the stripe(+icon) treatment; keying on the URL (not a boolean) means a
+    // changed thumbnailUrl naturally retries without an effect to reset state.
+    const [failedThumbnailUrl, setFailedThumbnailUrl] = useState<string | undefined>(undefined);
+    const showThumbnailImage = thumbnailUrl !== undefined && thumbnailUrl !== failedThumbnailUrl;
+
     const { accent, accentText, tint } = getResourceTypeColors(type);
     // Registry-derived type glyph for the thumbnail; an explicit thumbnailIcon
     // prop still wins so callers can override the header treatment.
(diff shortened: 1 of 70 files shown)


calm-server/src/server/routes/render-route.ts
```diff
@@ -0,0 +1,221 @@
...
+ * to a very wide thin strip, and the browse card's `object-cover` would crop away
+ * both nodes leaving only the middle of the edge. Expanding the box toward the
+ * card header's minimum aspect (~1.75:1) adds empty canvas margin instead, so the
+ * cover crop trims margin — never the diagram.
+ *
+ * Exported for unit tests only. Runtime-safe to export: the function is fully
+ * self-contained (no captured bindings), which `page.evaluate` requires anyway
+ * to serialize it into the page.
+ */
+export function computeDiagramClip(): { x: number; y: number; width: number; height: number } | null {
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in f7a7ff1e — added minimal local `declare const document/window` shapes scoped to the in-page function, keeping DOM libs out of the Node tsconfig. One nuance: the build (tsup/esbuild) doesn't typecheck, so compilation wasn't actually failing — but the declares keep `tsc --noEmit` and editors clean.

calm-hub/src/main/java/org/finos/calm/services/ThumbnailService.java
```diff
@@ -0,0 +1,311 @@
...
+            @ConfigProperty(name = "calm.render.timeout-ms", defaultValue = "20000") long timeoutMs,
+            @ConfigProperty(name = "calm.render.failure-cache-ttl-ms", defaultValue = "60000") long failureCacheTtlMs) {
+        this(renderServiceUrl.orElse(""), uiBaseUrl, timeoutMs, failureCacheTtlMs, HttpClient.newBuilder()
+                .connectTimeout(Duration.ofSeconds(10))
+                .build());
+    }
+
+    /** Visible-for-testing constructor allowing an injected {@link HttpClient}. */
+    ThumbnailService(String renderServiceUrl, String uiBaseUrl, long timeoutMs, long failureCacheTtlMs, HttpClient httpClient) {
+        this.renderServiceUrl = renderServiceUrl == null ? "" : renderServiceUrl.trim();
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in f7a7ff1e — trailing slashes are stripped in the constructor, mirroring calm-server's uiBaseUrl normalization, with a unit test that fails against the double-slash path.

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
```diff
@@ -295,6 +304,117 @@ public Response getArchitectureTimeline(
...
+        try {
+            store.storeThumbnail(architecture, bytes);
+        } catch (NamespaceNotFoundException | ArchitectureNotFoundException | ArchitectureVersionNotFoundException e) {
+            logger.debug("Could not store rendered thumbnail for [{}]", architecture, e);
+        }
+    }
+
+    private Response thumbnailNotFoundResponse() {
+        return Response.status(Response.Status.NOT_FOUND).entity("No thumbnail available").build();
+    }
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in f7a7ff1e — the 404 helper now sets text/plain explicitly.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
```diff
@@ -243,6 +252,119 @@ public Response updateVersionedPattern(
...
+        try {
+            store.storeThumbnail(pattern, bytes);
+        } catch (NamespaceNotFoundException | PatternNotFoundException | PatternVersionNotFoundException e) {
+            logger.debug("Could not store rendered thumbnail for [{}]", pattern, e);
+        }
+    }
+
+    private Response thumbnailNotFoundResponse() {
+        return Response.status(Response.Status.NOT_FOUND).entity("No thumbnail available").build();
+    }
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in f7a7ff1e — same fix as the architecture resource: the 404 helper now sets text/plain explicitly.

calm-hub/src/main/java/org/finos/calm/services/ThumbnailService.java
```diff
@@ -0,0 +1,313 @@
...
+            return Optional.empty();
+        }
+        try {
+            return renderSingleFlight(key, documentType, documentJson, storeCallback).future().join();
+        } catch (Exception e) {
+            logger.debug("Thumbnail render failed for [{}]", key, e);
+            return Optional.empty();
+        }
+    }
+
```
jpgough-ms, line 195: `renderSync` ignores the caller's `documentJson`/`storeCallback` whenever a render for the same key is already in-flight — it just joins the existing future. A PUT that updates content while a stale GET-triggered render is still running gets the old (pre-update) render back, and that stale image is then pinned client-side for 5 minutes via `Cache-Control: private, max-age=300`. The write path (`triggerRender`) already guards this exact race by chaining a follow-up render when it isn't the owner — see the coverage at `TestThumbnailServiceShould.java:385-411` — `renderSync` needs the same guard.

calm-hub-ui/src/hub/components/namespace-page/useNamespaceItems.ts
```diff
@@ -29,6 +35,20 @@ const summaryToItem = (s: ResourceSummary): NamespaceItem => ({
...
+ * thumbnail URL. Built on the numeric-ID storage API (the same base path the
+ * summary fetches use), so it uses the summary's numeric `id`, not the slug.
+ */
+const summaryToThumbnailItem = (
+    s: ResourceSummary,
+    namespace: string,
+    typePath: 'architectures' | 'patterns'
+): NamespaceItem => ({
+    ...summaryToItem(s),
+    thumbnailUrl: `/api/calm/namespaces/${encodeURIComponent(namespace)}/${typePath}/${encodeURIComponent(String(s.id))}/thumbnail`,
```
jpgough-ms: Every `ItemCard`'s `thumbnailUrl` points at the synchronous on-demand render endpoint with no client- or server-side concurrency cap. First browse of an unpopulated namespace fires one blocking ~20s render per item, each spawning its own headless browser in calm-server (no pooling — see `render-route.ts`), risking hub request-thread exhaustion and multi-second page stalls until every item has a cached thumbnail. Consider a request cap, an async placeholder + poll pattern, or server-side batch triggering instead of one render call per card.

calm-server/src/server/routes/render-route.ts
```diff
@@ -0,0 +1,235 @@
...
+            // Injected before any page script runs, so the UI's render route sees
+            // the document on first mount (see RenderView in calm-hub-ui).
+            await page.addInitScript(
+                (data) => {
+                    (window as unknown as Record<string, unknown>)['__CALM_RENDER_DATA'] = data;
+                },
+                { documentType, document }
+            );
+            // Navigation bounded by the request timeout, not Playwright's 30s default.
+            await page.goto(`${base}/#/render`, { timeout: timeoutMs });
```
jpgough-ms: `page.goto` and `page.waitForSelector` are each budgeted the full `timeoutMs` sequentially (worst case ~2× `timeoutMs`), but `ThumbnailService`'s HTTP client only waits `timeoutMs + 5000ms` total. A render that's healthy but slow across its two steps (e.g. 15s + 15s) can total 30s server-side and still get aborted client-side at ~25s — then have that failure cached for `calm.render.failure-cache-ttl-ms` (default 60s), returning 404 for a render that would have succeeded. Bound both waits against one shared deadline, or raise the client-side grace to cover the worst case.

calm-hub-ui/src/hub/components/search-results/SearchResultsPage.tsx
```diff
@@ -62,14 +64,28 @@ export function SearchResultsPage({ searchService }: SearchResultsPageProps) {
         return () => controller.abort();
     }, [query, service]);
 
+    // Groups the registry doesn't know are dropped rather than rendered with a
+    // guessed card type; the model's GroupedSearchResults keys make this a no-op
```
jpgough-ms, line 70: Results are now filtered to groups present in `SEARCH_GROUP_TYPE`, but `totalCount` is still computed from the unfiltered `flattenResults(results)`. If the backend ever returns a group key the frontend registry doesn't recognize (e.g. a new resource type shipped ahead of the UI), those results vanish with no error while the header still reports the original count — "N results" with fewer than N cards rendered.

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
```diff
@@ -89,7 +89,7 @@ export function NamespacePage({ namespace, counts }: NamespacePageProps) {
                         message={`No ${getResourceTypeMeta(active).pluralLabel} in this namespace yet`}
                     />
                 ) : (
-                    <div className="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 gap-4">
+                    <div className="grid gap-4 grid-cols-[repeat(auto-fill,minmax(280px,1fr))] max-w-[1240px]">
```
jpgough-ms: The explicit responsive grid (`grid-cols-1` / `md:grid-cols-2` / `lg:grid-cols-3`) was replaced with `grid-cols-[repeat(auto-fill,minmax(280px,1fr))]`, dropping the guaranteed single-column layout below the 768px breakpoint. Any viewport whose content area is narrower than 280px forces a column wider than the container, causing horizontal overflow. Same pattern repeats in `DomainPage.tsx:92` and `SearchResultsPage.tsx:142`.

calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
```diff
@@ -65,14 +74,35 @@ export function ItemCard({
     customId,
     versionCount,
     meta,
-    thumbnailHeight = 96,
+    thumbnailHeight = 160,
```
jpgough-ms: Default `thumbnailHeight` moved from 96px to 160px and applies to every card type — including Flows, Standards, ADRs, Interfaces, and Controls, none of which render an actual thumbnail image. Browsing those tabs now shows a much taller empty/icon-only header per card, pushing content down and reducing visible cards per screen with no visual benefit.

calm-hub-ui/src/hub/components/search-results/SearchResultsPage.tsx
```diff
@@ -117,46 +133,27 @@ export function SearchResultsPage({ searchService }: SearchResultsPageProps) {
                         {groups.map(([type, items]) => (
                             <section key={type}>
                                 <div
-                                    className="font-mono-jb text-[11px] uppercase tracking-[0.1em] mb-3"
+                                    className="flex items-center gap-1.5 font-mono-jb text-[11px] uppercase tracking-[0.1em] mb-3"
                                     style={{ color: colors.redesign.faintAlt }}
                                 >
+                                    {renderGroupIcon(type)}
                                     {TYPE_LABELS[type] ?? type} ({(items as SearchResult[]).length})
```
jpgough-ms, line 135: Each result changed from a single `<button>` whose accessible name included item name, namespace, and description to an `ItemCard` whose activation button wraps only the item name. Screen-reader users now hear just the bare name per result — two items both named "Payment Flow" in different namespaces are announced identically, losing the context that disambiguated them.

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
```diff
@@ -295,6 +304,119 @@ public Response getArchitectureTimeline(
...
+            @PathParam("version") @Pattern(regexp = VERSION_REGEX, message = VERSION_MESSAGE) String version) {
+        Architecture architecture = new Architecture.ArchitectureBuilder()
+                .setNamespace(namespace)
+                .setId(architectureId)
+                .setVersion(version)
+                .build();
+
+        return architectureThumbnailResponse(architecture);
+    }
+
```
jpgough-ms, line 326: The full thumbnail flow (~90 lines: `architectureThumbnailResponse`, `renderArchitectureThumbnailOnDemand`, `triggerThumbnailRender`, `storeThumbnailQuietly`, `thumbnailNotFoundResponse`) is duplicated near-verbatim in `PatternResource.java` with only the type swapped. A future fix to cache headers, the 404 body, or the single-flight key format applied to one and not the other leaves architectures and patterns with divergent thumbnail behavior. Worth factoring into `ThumbnailService` or a shared helper.

calm-hub-ui/src/visualizer/components/reactflow/ArchitectureGraph.tsx
```diff
@@ -67,9 +79,15 @@ function readMinimapHidden(): boolean {
...
 }
 
-export function ArchitectureGraph({ jsonData, onNodeClick, onEdgeClick, viewportKey }: ArchitectureGraphProps) {
+export function ArchitectureGraph({ jsonData, onNodeClick, onEdgeClick, viewportKey, chromeless = false }: ArchitectureGraphProps) {
     const isMobile = useIsMobile();
 
+    // Chromeless render mode: true once nodes are measured and the first paint has
+    // settled (RenderReadySignal), surfaced as data-render-ready on the wrapper so
+    // calm-server's headless browser knows the graph is safe to screenshot.
+    const [renderReady, setRenderReady] = useState(false);
```
jpgough-ms: `renderReady`/`markRenderReady` state, the parsed-flag empty-state wrapper, and `CHROMELESS_FIT_VIEW_OPTIONS` are duplicated near-identically between `ArchitectureGraph.tsx` and `PatternGraph.tsx`. A fix to render-ready signalling applied to only one leaves the other's thumbnails cropped or timing out on empty documents. Consider extracting a shared `useChromelessRender()` hook.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
```diff
@@ -306,56 +308,132 @@ public Pattern updatePatternForVersion(Pattern pattern) throws NamespaceNotFound
...
+            lock.unlock();
+        }
 
         LOG.info("Updated version '{}' for pattern {} in namespace '{}'",
                 pattern.getMongoVersion(), pattern.getId(), pattern.getNamespace());
         return pattern;
     }
 
+    @Override
+    public void storeThumbnail(Pattern pattern, byte[] png) throws NamespaceNotFoundException, PatternNotFoundException, PatternVersionNotFoundException {
```
jpgough-ms, line 399: `storeThumbnail`'s find/mutate/splice/update sequence duplicates `updatePatternForVersion` immediately above it. A fix to the splice logic (e.g. the pattern-not-found fallthrough) needs to land in both places; missing one leaves it able to silently write to the wrong document or drop the update. Worth factoring a shared "update this pattern document" helper.

jpgough-ms (commented):
Automated high-effort review: 14 findings independently verified, 10 posted inline below (7 correctness, 3 cleanup/duplication). 4 lower-impact duplication findings were dropped by the report cap and are not posted here: icon-render snippet (`renderGroupIcon`) duplicated across `ExplorerSearch.tsx`/`GlobalSearchBar.tsx`/`IntroSearchBar.tsx`/`SearchResultsPage.tsx`; `ErrorResponse` class duplicated between `render-route.ts` and `validation-route.ts`; thumbnail URL template duplicated between `useNamespaceItems.ts` and `SearchResultsPage.tsx`; grid class string duplicated across `NamespacePage.tsx`/`DomainPage.tsx`/`SearchResultsPage.tsx`.

calm-hub/src/main/java/org/finos/calm/services/ThumbnailService.java
rocketstack-matt: Partially addressed in a1b5a3a9 — documented rather than changed, deliberately. The read path doesn't chase because concurrent GETs for one missing thumbnail are the common case single-flight exists for; a per-joiner chase would fire N renders. The write path's chase (which you point to) already converges the store, so the exposure is one client's cache window (≤5 min) in a seconds-wide same-version-update race. The trade-off is now spelled out in renderSync's Javadoc — happy to revisit if you think it's the wrong call.

calm-hub-ui/src/hub/components/namespace-page/useNamespaceItems.ts
rocketstack-matt: Fixed in a1b5a3a9 — renders are now bounded by a semaphore (calm.render.max-concurrent-renders, default 2). Only render owners take permits; when saturated the attempt returns empty immediately WITHOUT entering the failure cache, so the next view retries rather than being 404-pinned. The card images are also loading="lazy", so off-screen cards don't fire requests at all.

calm-server/src/server/routes/render-route.ts
rocketstack-matt: Fixed in a1b5a3a9 — goto and waitForSelector now share a single deadline, so the combined server-side budget stays within timeoutMs and inside the client's timeoutMs+5s wait.

calm-hub-ui/src/hub/components/search-results/SearchResultsPage.tsx
rocketstack-matt: Fixed in a1b5a3a9 — totalCount is now computed from the filtered groups.

calm-hub-ui/src/hub/components/namespace-page/NamespacePage.tsx
rocketstack-matt: Fixed in a1b5a3a9 — minmax(min(280px,100%),1fr) removes the sub-280px overflow, and the grid class is now a single shared constant (CARD_GRID_CLASS) across the three pages. One note: dropping the guaranteed single column below 768px was deliberate — cards are width-capped so extra viewport becomes extra columns, and two ~300px columns at ~600px is the intended behaviour.

calm-hub-ui/src/hub/components/namespace-page/ItemCard.tsx
rocketstack-matt: Fixed in a1b5a3a9 — the header is 160px only when a thumbnail URL is present and 96px otherwise. Rows can't misalign because every browse surface shows a single type at a time.

calm-hub-ui/src/hub/components/search-results/SearchResultsPage.tsx
rocketstack-matt: Fixed in a1b5a3a9 — the activation element now carries aria-describedby referencing the description and the namespace chip, so same-named results are announced with their disambiguating context. Applies to all card surfaces, not just search.

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
rocketstack-matt: Fixed in a1b5a3a9 — extracted ThumbnailEndpointSupport: response building, the on-demand render flow and the 404 shape live once, with both resources delegating through small functional seams. Both resource test suites pass unchanged.

calm-hub-ui/src/visualizer/components/reactflow/ArchitectureGraph.tsx
rocketstack-matt: Fixed in a1b5a3a9 — shared chromeless module (fit options + useChromelessRender hook) and a ChromelessEmptyState wrapper, used by both graphs.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
rocketstack-matt: Fixed in a1b5a3a9 — both Nitrite stores now route the version-update and storeThumbnail writes through one per-store mutate helper; lock scope and each caller's exception mapping are unchanged.

rocketstack-matt:
The four duplication findings noted in the review body (dropped by the report cap) are also addressed in a1b5a3a9: `renderGroupIcon` → shared `SearchGroupIcon` component; `ErrorResponse` → shared `error-response` module in calm-server; the thumbnail URL template → `latestThumbnailUrl()` helper; the grid class → `CARD_GRID_CLASS` constant.

jpgough-ms (approved):

markscott-ms (changes requested):
I'm not overly keen on the calm-server dependency by CalmHub (particularly with its requirement for headless Chrome or Edge) - and this PR doesn't add calm-server to the CalmHub container.

@rocketstack-matt says he'll refactor.

PR #2820, opened 2026-07-15
Automatically validate architectures after bump + interactively ask for change type

willosborne:
## Description
- After a `workspace bump`, re-validate architectures in case they no longer work against the pattern
- Do the same for `workspace check` as this is a CI control
- Ask the user what change type they're making when they bump.

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


cli/src/command-helpers/workspace/bump.ts
```diff
@@ -17,6 +17,22 @@ export { canonicalEqual };
...
 const logger: Logger = initLogger(false, 'workspace');
 
+/**
+ * Rewrites `$id`/`title` for a version bump without the empty-description default that
+ * `updateDocumentMetadata` applies for CalmHub push normalisation (see hub-commands.ts, where it
+ * mirrors how CalmHub stores a description-less document). Workspace documents are pushed raw
+ * (see push.ts) and validated locally, so injecting `description: ''` into a document that never
+ * had one would fail the `architecture-has-no-empty-string-properties` spectral rule.
+ */
+function bumpDocumentContent(raw: string, metadata: DocumentMetadata): string {
```
markscott-ms, line 30: bumpDocumentContent calls JSON.parse up to three times per document (once for the hasOwnProperty check, once inside updateDocumentMetadata, once more to delete description). Not a correctness problem, just avoidable overhead — a single parse/patch/stringify would do.

Can we reduce this?

cli/src/command-helpers/workspace/post-bump-validate.ts
```diff
@@ -0,0 +1,60 @@
...
+        let content: object;
+        try {
+            content = JSON.parse(await readFile(filePath, 'utf8'));
+        } catch {
+            results.push({ id, filePath, type, passed: false, errorCount: -1 });
+            continue;
+        }
+
+        try {
+            const outcome = type === 'architecture'
```
markscott-ms: The post-bump validation never actually checks an architecture against its pattern, despite that being the PR's headline stated purpose ("re-validate architectures in case they no longer work against the pattern").

In cli/src/command-helpers/workspace/post-bump-validate.ts:36-38:

```
const outcome = type === 'architecture'
    ? await validate(content, undefined, undefined, schemaDir, false)
    : await validate(undefined, content, undefined, schemaDir, false);
```

Passing `undefined` for the pattern argument when validating an architecture puts `validate()` into 'architecture-only' mode (`shared/src/commands/validate/validate.ts:154-155`). Per that mode's own JSON-schema rule docstring (`shared/src/commands/validate/rules/json-schema-rule.ts:19-20`):

▎ "architecture-only: discover the latest CALM core schema and validate against it... does NOT [check the pattern]"

So this only runs generic CALM-core-schema/spectral checks — it can never catch an architecture that has drifted out of conformance with its own pattern after the pattern was bumped, which is exactly the regression this feature exists to catch.

The real CLI path (`cli/src/command-helpers/validate.ts` → `loadArchitectureAndPattern` → `loadPatternFromDocumentIfPresent`, `shared/src/document-loader/loading-helpers.ts:49-66`) resolves the architecture's `$schema` to its pattern and passes both into `validate()`, which does check conformance (architecture-with-pattern mode). `runPostBumpValidation` should do the same — load the pattern via `$schema` (using the same `WorkspaceDocumentLoader`/`schemaDir.getSchema` machinery already wired up) and pass it as the second validate() argument for architectures.

This is made worse by the failure message the feature prints:

```
1 document(s) failed validation:
  my-arch (2 error(s)) — run `calm validate -a .calm-workspace/files/my-arch.json` to see full output
```

The suggested manual command is strictly more thorough than the automated check that produced this message — so a document can pass workspace bump/workspace check silently and still fail calm validate -a, giving users false confidence right when the feature is supposed to catch that exact class of regression. No test (unit or smoke) exercises an architecture/pattern mismatch, so this gap isn't caught by CI either.

*Suggested fix*: in `post-bump-validate.ts`, for `type === 'architecture'`, load the pattern the same way `loadPatternFromDocumentIfPresent` does (via the architecture's `$schema` and `schemaDir.getSchema`/`docLoader.loadMissingDocument`) and pass it as the pattern argument to validate().

markscott-ms (changes requested):

cli/src/command-helpers/workspace/post-bump-validate.ts
willosborne: Fixed in b19b2648. `runPostBumpValidation` now resolves the architecture's pattern via its `$schema` using `loadPatternFromDocumentIfPresent` (the same machinery the real `calm validate -a` path uses) and passes it as the second `validate()` argument, so it runs in architecture-with-pattern mode. It falls back to architecture-only only when no pattern can be resolved from `$schema`.

Added coverage: unit tests assert the pattern is loaded from `$schema` and forwarded to `validate()` (and the architecture-only fallback), and `workspace.smoke.spec.ts` now asserts `workspace check` fails with a validation error when a document violates its schema — an end-to-end guard against regressing back to architecture-only mode.

cli/src/command-helpers/workspace/bump.ts
willosborne: Fixed in b19b2648. `bumpDocumentContent` now parses and stringifies once: it patches `$id`/`title` on a single parsed object and only touches `description` when it was already present (preserving the no-empty-description behaviour). It no longer routes through `updateDocumentMetadata`, so there are no redundant parse passes.

cli/AGENTS.md
```diff
@@ -63,9 +63,8 @@ npm run copy-workspace-templates # Copy `workspace new` templates from src/comma
...
-    - `workspace rm [id]` — removes a document from the active workspace manifest; prompts interactively for the id if omitted.
+    - `workspace check` — CI/PR gate: exits non-zero if any tracked document changed on disk relative to CalmHub but wasn't version-bumped, or if any architecture/pattern in the workspace fails validation. Prints a pass/fail validation summary alongside the bump check.
+    - `workspace bump` — prompts interactively for the bump type (MAJOR/MINOR/PATCH) for each directly-changed document (default from config or MINOR). Cascade-bumped dependents are also prompted with the trigger's increment as the default; `--inherit-change-type` suppresses cascade prompts and auto-inherits silently. `--major`/`--patch` skip all prompts. After bumping, repoints all references to the new `$id`s and prints a post-bump validation summary (informational only). Idempotent: re-bumping an already-bumped doc is a no-op until it's pushed.
```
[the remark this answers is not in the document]
willosborne: Fixed in b19b2648 — AGENTS.md now reads `--major`/`--minor`/`--patch` skip all prompts, matching the command wiring and README.

markscott-ms (approved):
Looks great now.

Issue #2853, opened 2026-07-22
Embed CALM diagrams in Docusaurus MDX — @calmstudio/docusaurus-plugin ("CALM as Mermaid replacement")


gjs-opsflo:
## Problem

Teams document architectures in Docusaurus sites today with hand-drawn Mermaid diagrams that drift from reality. CALM files are the source of truth, but there is no way to render them inline in docs pages. Community feedback (2026-07-09) asked for CALM diagrams natively embeddable in docs sites, replacing Mermaid.

## Proposal

A Docusaurus plugin so this works in any MDX page:

```mdx
import { CalmDiagram } from '@calmstudio/docusaurus-plugin'

<CalmDiagram src="./architectures/loan-approval.calm.json" />
```

- Build-time static SVG (SEO, no-JS, instant paint; light+dark baked, follows site theme)
- Hydrates client-side to the existing `@calmstudio/diagram` web component (zoom/pan/tooltips/flow animation)
- Broken `.calm.json` fails the docs build with the file path — drift caught in CI, like broken Mermaid
- Props: `src`, `data`, `theme`, `flow`, `interactive`

## Ship path (small sequential PRs)

1. Package `@calmstudio/docusaurus-plugin` under `calm-studio/packages/` — includes a fix in `@calmstudio/diagram`: it previously only understood a legacy flat `relationship-type` shape, so canonical CALM documents rendered with no edges
2. Root `docs/` site dogfoods it on a real architecture (`calm/architecture/calm-1.json`)
3. Follow-ups: `showValidation` prop; Docify integration as a separate small plugin (same pattern — relates to #2577, #2079)

Implementation is complete and tested; PRs to follow this issue.

Issue #2851, opened 2026-07-22
Office Hours - 2026-07-23


markscott-ms:
## Date

Thu 23 Jul 2026 - 1030 EST / 1530 UK

## Zoom info

Join Zoom Meeting
https://zoom-lfx.platform.linuxfoundation.org/meeting/98579493833?password=2af32413-a2a4-4ada-b064-ae8544c64c85

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

Office Hours is a meeting for maintainers to get together to discuss active development and upcoming priorities.

- [x] CALM plugin to use React Flow (Calm Hub) based tech stack @byrash 
- [x] Advent of CALM - do we still need to keep it following docs renovations?
- [x] https://github.com/finos/architecture-as-code/issues/2857
- [x] finos/calm-governance#2 - motion is carried with 13 positive votes as of 3.36pm LNT (out of possible 18)
- [ ]  finos/architecture-as-code#2859 @YoofiTT96

byrash:
Hello @markscott-ms we have attempted to convert CALM plugin to use React Flow (Calm Hub) based tech stack. Was wondering if you can add us to agenda for tomorrow to demo the new plugin. Thank You.

markscott-ms:
# CALM plugin to use React Flow (Calm Hub) based tech stack

* Generally lovely
[...]

Issue #1998, opened 2026-01-02
Refactoring Proposal: Split shared module into calm-core and calm-commands


LeighFinegold:
### Target Project
shared

### Description of Feature

Split the existing shared module into two distinct modules:

1. `@finos/calm-core` - Core utilities with zero dependencies on higher-level packages (logger, consts, schema-directory, resolvers, model-visitors)
2. `@finos/calm-commands` - Command implementations (validate, generate, docify, template-processing) that depend on `calm-core`, `calm-widgets`, and `calm-models`

This refactoring resolves a **circular dependency anti-pattern** where `shared` depends on `calm-widgets`, but `calm-widgets` needed to duplicate logger logic because it cannot depend on `shared`. The split enables a proper dependency hierarchy and unlocks future extensibility.

### Current Limitations

| Problem | Evidence |
|---------|----------|
| **Circular dependency risk** | `shared/package.json` depends on `@finos/calm-widgets`, but `calm-widgets` cannot import from `shared` |
| **Duplicated logger code** | `shared/src/logger.ts` (Winston-based, 78 lines) vs `calm-widgets/src/widget-logger.ts` (custom interface, 49 lines) - two separate implementations |
| **Re-export indirection** | `shared/src/index.ts` re-exports `setWidgetLogger` from `calm-widgets` to hide the dependency complexity |
| **Unclear module boundaries** | Commands (validate, generate, docify) are mixed with core utilities (logger, consts, schema-directory) in one package |
| **Blocking future React integration** |  docify cannot depend on a shared utilities module, preventing clean integration of React components |

### Proposed Implementation

**New Package Structure:**

```
calm-core/                    # NEW: Zero dependencies on other CALM packages
├── src/
│   ├── logger.ts             # Unified logger (absorbs widget-logger)
│   ├── consts.ts
│   ├── schema-directory.ts
│   ├── resolver/             # Reference resolvers
│   ├── model-visitor/        # Dereference visitor etc.
│   └── util.ts
└── package.json              # deps: winston, loglevel (no @finos/* deps)

calm-commands/                # NEW: Command implementations
├── src/
│   ├── commands/
│   │   ├── validate/
│   │   └── generate/
│   ├── docify/
│   ├── template/
│   └── document-loader/
└── package.json              # deps: @finos/calm-core, @finos/calm-widgets, @finos/calm-models
```

**Dependency Graph (after refactor):**

```mermaid
graph TD
    subgraph Foundation Layer
[...]

PR #2854, opened 2026-07-22
feat(calm-suite): @calmstudio/docusaurus-plugin — embed CALM diagrams in Docusaurus MDX

gjs-opsflo:
## Description

New package **`@calmstudio/docusaurus-plugin`** (under `calm-studio/packages/`) — embed CALM architecture diagrams directly in Docusaurus MDX pages, the same way Mermaid diagrams embed today, but driven by real, validatable `.calm.json` files:

```mdx
import { CalmDiagram } from '@calmstudio/docusaurus-plugin'

<CalmDiagram src="./architectures/loan-approval.calm.json" />
```

How it works:
- An async webpack loader pre-renders each `.calm.json` to static SVG at build time (light + dark variants) — SEO-friendly, works without JavaScript, instant paint, follows the site theme via `html[data-theme]` CSS.
- A remark plugin rewrites relative `src` paths into hoisted imports, so authors write plain paths.
- The React `<CalmDiagram>` component SSRs the static SVG and lazily upgrades client-side to the existing `@calmstudio/diagram` web component (zoom, pan, node tooltips, flow animation). `interactive={false}` keeps the static SVG with zero client JS.
- Invalid or unrenderable `.calm.json` fails the docs build with the offending file path — architecture drift is caught in CI, exactly like broken Mermaid.

Also includes a standalone bug fix in `@calmstudio/diagram`: the renderer only understood a legacy flat `relationship-type` shape, so **canonical CALM documents rendered with zero edges**. It now expands the nested `connects` / `interacts` / `composed-of` / `deployed-in` variants via the `@calmstudio/calm-core` accessors (flat shape kept as fallback), and tolerates schema-valid documents missing top-level `nodes` / `relationships`.

Supporting changes: node-safe `./render` (+CJS) export on `@calmstudio/diagram` so build tools can render without `customElements`; single-file ES bundle preserved for CDN use; CI wiring (release workflow, studio build workflow, and docs workflows build the new package — no-ops until the docs site adopts it in the follow-up PR).

Part of #2853. Follow-up PR dogfoods the plugin on the root `docs/` site with a real architecture (screenshot in the issue).

Notes for reviewers:
- `package-lock.json` includes removal of a stale `"extraneous": true` `calmguard-docs` ghost entry — benign npm tree recompute from registering the new workspace.
- `@calmstudio/diagram` has 17 pre-existing `tsc --noEmit` errors on `main` (legacy flat-shape test files); this PR does not add to them (net −3) and deliberately does not add that package's typecheck to CI.
- `showValidation` prop from the original idea is deferred (needs validation UI in the web component first).

## Type of Change

## Affected Components

- [x] CALM Studio (`calm-studio/`) — new `packages/docusaurus-plugin`, fixes in `packages/web-component`
- [x] CI/CD (`.github/workflows/`)

## Testing

- [x] All existing tests pass

Evidence: `@calmstudio/diagram` 31/31, `@calmstudio/docusaurus-plugin` 21/21 (loader, remark rewrite, SSR render paths, plugin wiring, relationship-edge expansion incl. defensive paths), full monorepo `npm test` exit 0, `npm run lint` 0 errors. End-to-end: root docs site built with the plugin + a real architecture — static SVG present in the emitted HTML for both themes (that adoption ships in the follow-up PR).

## Checklist

- [x] I have updated documentation if necessary (package README with install/config/props/error behavior)

### DIFF
--- a/calm-studio/packages/web-component/src/render/relationshipEdges.ts
+++ b/calm-studio/packages/web-component/src/render/relationshipEdges.ts
@@ -1,0 +1,109 @@
+// SPDX-FileCopyrightText: 2024 CalmStudio contributors - see NOTICE file
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  getConnectsEndpoints,
+  getContainerAndNodes,
+  getActorAndNodes,
+  getRelationshipVariant,
+  type CalmRelationship,
+} from '@calmstudio/calm-core';
+
+/**
+ * A single renderable edge derived from a CALM relationship. One relationship
+ * can expand into zero, one, or many edges depending on its variant (e.g. an
+ * `interacts` relationship with N nodes expands into N edges).
+ */
+export interface DiagramEdge {
+  id: string;
+  source: string;
+  target: string;
+  variant: string;
+}
+
+/** True when `rel` uses the legacy flat CalmStudio shape (never valid CALM). */
+function isFlatRelationship(
+  rel: CalmRelationship
+): rel is CalmRelationship & { source: string; destination: string } {
+  const raw = rel as unknown as Record<string, unknown>;
+  return (
+    typeof raw['relationship-type'] === 'string' &&
+    typeof raw.source === 'string' &&
+    typeof raw.destination === 'string'
+  );
+}
+
+/**
+ * Expand a CALM relationship into zero or more diagram edges.
+ *
+ * Handles both the canonical nested `relationship-type` object (connects,
+ * interacts, composed-of, deployed-in, options) and the legacy flat
+ * CalmStudio shape as a fallback. Defensive: any edge with a missing
+ * endpoint id is skipped rather than emitted with an undefined id, since
+ * ELK crashes or produces garbage on undefined node references.
+ *
+ * @param rel - The CALM relationship to expand
+ * @returns Array of diagram edges (possibly empty)
+ */
+export function relationshipToEdges(rel: CalmRelationship): DiagramEdge[] {
+  const uid = rel['unique-id'];
+
+  if (isFlatRelationship(rel)) {
+    const raw = rel as unknown as Record<string, unknown>;
+    const variant = raw['relationship-type'] as string;
+    if (!rel.source || !rel.destination) return [];
+    return [{ id: uid, source: rel.source, target: rel.destination, variant }];
+  }
+
+  const rt = rel['relationship-type'];
+  const variant = getRelationshipVariant(rt);
+
+  switch (variant) {
+    case 'connects': {
+      const endpoints = getConnectsEndpoints(rel);
+      if (!endpoints || !endpoints.source || !endpoints.destination) return [];
+      return [{ id: uid, source: endpoints.source, target: endpoints.destination, variant }];
+    }
+    case 'interacts': {
+      const actorAndNodes = getActorAndNodes(rel);
+      if (!actorAndNodes || !actorAndNodes.actor) return [];
+      return buildFanOutEdges(uid, actorAndNodes.actor, actorAndNodes.nodes, variant);
+    }
+    case 'composed-of':
+    case 'deployed-in': {
+      const containerAndNodes = getContainerAndNodes(rel);
+      if (!containerAndNodes || !containerAndNodes.container) return [];
+      return buildFanOutEdges(uid, containerAndNodes.container, containerAndNodes.nodes, variant);
+    }
+    case 'options':
+    default:
+      return [];
+  }
+}
+
+/**
+ * Build one edge per target node from a single source (actor/container).
+ * Uses the plain `uid` when there is exactly one target node (preserving
+ * flow-overlay matching against `relationship-unique-id`), else suffixes
+ * with `__${index}` for each node.
+ */
+function buildFanOutEdges(
+  uid: string,
+  source: string,
+  nodes: string[],
+  variant: string
+): DiagramEdge[] {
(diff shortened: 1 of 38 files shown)


gjs-opsflo:
Here is an example of rendering calm-platform.calm.json, the CALM self-description of the CALM platform itself. 

Screenshot 2026-07-22 at 2 56 59 PM

willosborne:
this is very cool, do you think `calm docify` should use this too @LeighFinegold ?

calm-studio/packages/web-component/src/render/relationshipEdges.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (validNodes.length === 1 && firstNode !== undefined) {
+    return [{ id: uid, source, target: firstNode, variant }];
+  }
+  return validNodes.map((target, index) => ({
+    id: `${uid}__${index}`,
+    source,
+    target,
+    variant,
+  }));
+}
```
rocketstack-matt, line 113: Fan-out edges get ids suffixed with `__${index}` (e.g. `rel-x__0`), but flow matching elsewhere still compares against the raw relationship `unique-id`: `flowOverlay.ts`'s `renderFlowOverlay` does `edgeLayouts.get(transition['relationship-unique-id'])`, and `elkRender.ts:192` does `applyFlowDimming(edge.id, activeFlowEdgeIds, true)`. Neither matches a suffixed id, so any flow transition referencing an `interacts`/`composed-of`/`deployed-in` relationship with 2+ target nodes silently renders no animated overlay and dims all of that relationship's edges to 0.3 opacity — no error, and nothing in the test suite exercises this combination.

Confirmed by direct repro against the built package: a 2-target `interacts` relationship referenced by a flow transition produces zero `animateMotion` elements and both fan-out edges at `opacity="0.3"`, vs. a `connects` relationship in the same setup which renders correctly.

Root cause: `DiagramEdge.id` conflates the render id with the source relationship's id, so identity is lost once a relationship fans out to N edges. Suggest adding a `relationshipId` field to `DiagramEdge` (separate from the render `id`) and matching flow transitions against that instead.

rocketstack-matt (changes requested):
Agree this is very cool.

One confirmed bug in the relationship fan-out logic; otherwise builds/tests clean (31/31 diagram, 21/21 docusaurus-plugin) and matches the linked issue's ship path.

Also, presumably the npm package name should be something like `@finos\calm-docusaurus-plugin` as it will need to be under the finos namespace?

LeighFinegold:
@willosborne — yes very cool indeed, and inline with previous discussions:-

- #790
- #1116
- #1998


Nice to see some of the discussions with [@aidanm3341](https://github.com/aidanm3341), [@Thels](https://github.com/Thels) and [@willosborne](https://github.com/willosborne) which was set aside earlier now being recognised as the way forward, and further enhances the eventual consolidation of calm-studio (calm builder tool) and calm-hub concepts. 
My only hope is we now see the convergence of the react controls between the tooling so its all just the same thing. 



@

initcron:
Thanks @rocketstack-matt — both points addressed:

**Fan-out flow bug** (`14bd2eb7`): took exactly your suggested shape — `DiagramEdge` now carries `relationshipId` separate from the render `id`. Flow dimming matches on `relationshipId`, and the overlay animates along **every** render edge a fanned-out relationship expands into (one sequence badge per transition, placed on its first edge). New tests cover the case you reproduced: a 2-target `interacts` referenced by a flow now yields one `animateMotion` per fan-out edge and no false dimming, both at the `relationshipToEdges` unit level and through a full `renderELKDiagram` integration render.

**Naming** (`149a0a4e`): renamed to `@finos/calm-docusaurus-plugin` (assuming the `\` was a typo for `/`) — agreed it belongs in the `@finos` namespace with the rest of the `@finos/calm-*` family. Directory stays under `calm-studio/packages/` for now; happy to relocate if you'd rather it live elsewhere. One thing to confirm on your side: the release workflow publishes with the existing `NPM_TOKEN` — assuming that token has publish rights on the `@finos` npm scope, no further change is needed.

Suites: diagram 33/33 (3 new), plugin 21/21, typecheck clean on the plugin, workflows updated for the new name.

initcron:
Thanks for the feedback @LeighFinegold,  @rocketstack-matt and @willosborne . Agree on convergence of the react controls between the tooling. I have added a [comment to Issue 2857](https://github.com/finos/architecture-as-code/issues/2857#issuecomment-5086969082) and appreciate your inputs on that. What approach should we pick ? A. Add edit feature to Hub  B. Migrate Studio to a web app with same react engine.

initcron:
Heads-up on a related follow-up, kept out of this PR to avoid growing it mid-review: on a branch stacked on this one, `composed-of`/`deployed-in` relationships now render as **nested container boxes** (elkjs hierarchical layout) instead of dashed arrows — with a `containers: 'nested' | 'edges'` opt-out. That addresses the point raised at office hours (#2851) about Mermaid's static render being good at nested containers/basic layouts.

It builds on the `relationshipId` machinery from this PR's review round, so it PRs cleanly only once this merges — I'll open it as its own small PR right after. Happy to share a rendered before/after if useful.

rocketstack-matt (approved):
Both requested changes correctly addressed. `DiagramEdge.relationshipId` separates render identity from source relationship id — every fan-out edge now animates and dims correctly, with tests covering the exact repro. Package renamed to `@finos/calm-docusaurus-plugin`. Three-pass review finds no new defects.

PR #2865, opened 2026-07-23
fix(shared): don't swallow malformed 200 bodies into [] in calm-hub-client

### LINKED ISSUE #2861: calm-hub-client: list-style methods silently return [] instead of surfacing malformed/auth-gateway response bodies
## Summary

Several list-style methods on `CalmHubClient` (`shared/src/hub/calm-hub-client.ts`) extract results with:

```js
const values = response.data?.values ?? [];
```

This correctly returns `[]` for a genuinely empty Hub list (e.g. `{ values: [] }` or `{}`), and any non-2xx response (401/403/500, etc.) already throws correctly via `wrapError` — that part is not broken.

The gap is a 200 response whose body simply doesn't look like a Hub response at all (a string, `null`, or otherwise not a plain object) - the current code can't tell that apart from "no items" and silently returns `[]`.

This matters most for authentication failures that don't originate from CALM Hub itself: CALM Hub's own auth layer always returns a proper 401/403 (confirmed by inspecting `calm-hub`'s security code), but anything sitting in front of it - an SSO gateway, a reverse-proxy `auth_request`, or a custom `AuthPlugin` whose token silently expired - commonly responds `200 OK` with a login page or unrelated JSON instead of a clean 401. Today that's swallowed into `[]` by every list-style method: `listNamespaces`, `listDomains`, `listControls`, `listControlConfigurations`, `getNamespaceMappings`, `getControlRequirementVersions`, `getControlConfigurationVersions`, `getMappedResourceVersions`. A user could see "no namespaces" / "no versions" when the real problem is a broken auth plugin or gateway.

Separately (found while investigating): the 404-swallow special case in three of these methods used `err instanceof AxiosError` to detect a genuine "not found" response. This check is unreliable across module-instance boundaries (reproduced with `axios-mock-adapter` under Vitest, and a real risk in any bundler setup that ends up with more than one loaded copy of `axios`) - `axios.isAxiosError(err)` (already used elsewhere in the same file's `wrapError`) is the robust, duck-typed equivalent.

## Fix

- Added a shared `extractValues` helper that only treats a plain object as "possibly empty" (defaulting to `[]` when `values` is absent); anything else (string/array/other primitive) throws a `HubClientError` instead of silently returning `[]`.
- Wired this into all list-style methods above.
- Replaced the fragile `err instanceof AxiosError` 404 checks with `axios.isAxiosError(err)`.
- Added regression tests in `calm-hub-client.spec.ts` covering: malformed/non-object 200 bodies now throwing, a failing `AuthPlugin.getAuthHeaders()` correctly rejecting rather than resolving to `[]`, and a real 401 from Hub still surfacing as an error.

## Note

`calm-hub-client.ts`'s own error propagation for real 401/403/500 responses was already correct prior to this fix (verified via source review of `calm-hub`'s security layer and the client's `wrapError`). This issue is specifically about the 200-with-unexpected-body gap and the `instanceof` fragility.

willosborne:
## Description
List-style `CalmHubClient` methods (`listNamespaces`, `listDomains`, `listControls`, `listControlConfigurations`, `getNamespaceMappings`, `getControlRequirementVersions`, `getControlConfigurationVersions`, `getMappedResourceVersions`) defaulted to `[]` whenever a 200 response lacked a `values` array. That can't distinguish a genuinely empty Hub list from a 200 response that doesn't look like Hub JSON at all - e.g. an SSO gateway or reverse-proxy returning a login page instead of a real 401 when a credential/`AuthPlugin` fails. Added an `extractValues()` helper so only a plain object is treated as possibly-empty; anything else (string/array/other primitive) now throws a `HubClientError` instead of silently reporting "no results".

Also replaced the `err instanceof AxiosError` 404 checks (in `getControlRequirementVersions`, `getControlConfigurationVersions`, `getMappedResourceVersions`) with `axios.isAxiosError(err)`, matching `wrapError`'s existing duck-typed check - `instanceof` is unreliable across module-instance boundaries (reproduced with `axios-mock-adapter` under Vitest; a real risk in any bundler setup with more than one loaded copy of `axios`).

`calm-hub-client.ts`'s error propagation for genuine 401/403/500 responses was already correct - verified via source review of `calm-hub`'s security layer (always returns proper 401/403, never a disguised 404) and the client's `wrapError`. This PR is specifically about the 200-with-unexpected-body gap and the `instanceof` fragility.

Fixes #2861

## Type of Change

## Affected Components
- [x] Shared (`shared/`)

## Testing
- [x] All existing tests pass

Added regression tests in `calm-hub-client.spec.ts` covering: malformed/non-object 200 bodies now throwing for every affected method, a failing `AuthPlugin.getAuthHeaders()` correctly rejecting rather than resolving to `[]`, and a real 401 from Hub still surfacing as an error. Full `shared` (1069 tests) and `cli` (624 tests) suites pass; `npm run build:shared` and lint are clean.

## Checklist

### DIFF
--- a/shared/src/hub/calm-hub-client.ts
+++ b/shared/src/hub/calm-hub-client.ts
@@ -1,4 +1,4 @@
-import axios, { Axios, AxiosError } from 'axios';
+import axios, { Axios } from 'axios';
 import { AuthPlugin } from '../auth/auth-plugin';
 import { initLogger, Logger } from '../logger';
 import { DocumentMetadata, extractDocumentMetadata, validateDocumentId } from './document-id-utils';
@@ -128,8 +128,7 @@ export class CalmHubClient {
         const endpoint = 'GET /api/calm/namespaces';
         try {
             const response = await this.ax.get('/api/calm/namespaces');
-            const values: HubNamespaceSummary[] = response.data?.values ?? [];
-            return values;
+            return this.extractValues<HubNamespaceSummary>(response.data, endpoint);
         } catch (err) {
             throw this.wrapError(err, endpoint);
         }
@@ -160,7 +159,7 @@ export class CalmHubClient {
         const endpoint = '/calm/domains';
         try {
             const response = await this.ax.get(endpoint);
-            const names: string[] = response.data?.values ?? [];
+            const names = this.extractValues<string>(response.data, `GET ${endpoint}`);
             return names.map(name => ({ name }));
         } catch (err) {
             throw this.wrapError(err, `GET ${endpoint}`);
@@ -178,7 +177,7 @@ export class CalmHubClient {
         const endpoint = `/calm/domains/${domain}/controls`;
         try {
             const response = await this.ax.get(endpoint);
-            return (response.data?.values ?? []) as HubControlSummary[];
+            return this.extractValues<HubControlSummary>(response.data, `GET ${endpoint}`);
         } catch (err) {
             throw this.wrapError(err, `GET ${endpoint}`);
         }
@@ -194,7 +193,7 @@ export class CalmHubClient {
         const endpoint = `/calm/domains/${domain}/controls/${controlName}/configurations`;
         try {
             const response = await this.ax.get(endpoint);
-            return (response.data?.values ?? []) as HubControlSummary[];
+            return this.extractValues<HubControlSummary>(response.data, `GET ${endpoint}`);
         } catch (err) {
             throw this.wrapError(err, `GET ${endpoint}`);
         }
@@ -209,9 +208,9 @@ export class CalmHubClient {
         const endpoint = `/calm/domains/${domain}/controls/${controlName}/requirement/versions`;
         try {
             const response = await this.ax.get(endpoint);
-            return (response.data?.values ?? []) as string[];
+            return this.extractValues<string>(response.data, `GET ${endpoint}`);
         } catch (err) {
-            if (err instanceof AxiosError && err.status === 404) {
+            if (axios.isAxiosError(err) && err.status === 404) {
                 return [];
             }
             throw this.wrapError(err, `GET ${endpoint}`);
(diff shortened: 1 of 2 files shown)


willosborne:
Fixes #2861

shared/src/hub/calm-hub-client.ts
```diff
@@ -396,6 +394,29 @@ export class CalmHubClient {
...
+     * `values` key (or no body at all) as a legitimately empty list. Anything that isn't a
+     * plain object - e.g. a string body - is rejected rather than silently swallowed, since a
+     * 200 response that doesn't look like Hub JSON is more likely an auth gateway/proxy
+     * returning a login page for a failed/expired credential than an actual empty result.
+     * @param data Response body.
+     * @param endpoint Endpoint label for error context.
+     * @returns The extracted values array, or [] when absent.
+     */
+    private extractValues<T>(data: unknown, endpoint: string): T[] {
+        if (data === undefined || data === null) {
```
rocketstack-matt: This treats a `null`/`undefined` body the same as no body (returns `[]`), but #2861's summary explicitly calls out `null` as one of the malformed-body cases that should throw rather than silently succeed — and there's no test for a literal `null` 200 body. Was excluding `null`/`undefined` from the check deliberate (e.g. for legitimate empty responses), or should it fold into the `typeof data !== 'object'` branch above?

rocketstack-matt (commented):
Solid fix — closes the string/array/no-values-key gap cleanly and the `axios.isAxiosError` swap is a good robustness fix. One question inline.

shared/src/hub/calm-hub-client.ts
willosborne: Good catch — that was inconsistent with the issue write-up. Fixed in 4ab40aa: `null` now falls into the throwing branch alongside strings/arrays/other primitives, and only `undefined` (a genuinely absent body) still maps to `[]`. Added a test for a literal `null` 200 body.

markscott-ms:
One question @willosborne - does this definitely work with the CALMHub API endpoints that do return an empty body such as some of the PUT / POST endpoints?

rocketstack-matt (approved):

rocketstack-matt:
Have approved but not merged pending response to @markscott-ms's question

markscott-ms (approved):
Answering myself - this PR introduces the new logic onto GET requests, so PUT/POST etc are unaffected.

Issue #2884, opened 2026-07-26
MongoDB document-per-namespace/domain storage risks hitting the 16MB BSON document limit


markscott-ms:
### Steps to Reproduce:
1. Inspect `MongoArchitectureStore`, `MongoPatternStore`, `MongoAdrStore`, `MongoFlowStore`, `MongoInterfaceStore`, `MongoStandardStore`, `MongoTimelineStore` (namespace-scoped) and `MongoControlStore` (domain-scoped) — each explicitly documents, in its own class-level javadoc, "one document per namespace/domain" holding an array of entities, each with a `versions` map containing the full raw content of every version ever created.
2. Note MongoDB enforces a hard 16MB limit per BSON document, inclusive of all nested arrays/sub-documents.
3. Note CalmHub is intentionally append-only today (no delete); the discussion on #2856 only proposes marking content archived, not removing it — nothing bounds this growth, and nothing currently proposed would either.
4. Calculate real-world exposure: the largest example CALM architecture JSON checked into this repo (`calm-plugins/vscode/test_fixtures/architecture/large.architecture.json`) is ~47KB; a realistic one (`docs/static/calm-example/trading-system.architecture.json`) is ~41KB. At ~40KB/version, a single namespace accumulates roughly **400 version-writes total** (summed across every architecture/pattern/etc. of that type ever created in that namespace) before hitting the 16MB ceiling — plausible for any actively-used, long-lived, shared namespace, and reachable far sooner for real enterprise-scale documents.

### Expected Result:
Writes continue to succeed indefinitely, or fail with a clear, accurate error identifying the real cause so operators can act.

### Actual Result:
- No safeguard exists anywhere. `MongoUpsertPush.pushWithDuplicateRetry` (used creating new architectures/patterns/etc.) only special-cases `DUPLICATE_KEY` and rethrows anything else raw.
- Worse, `MongoArchitectureStore.writeArchitectureToMongo`'s catch block **mismaps** any `MongoWriteException` — including a genuine "document too large" failure — into `ArchitectureNotFoundException`, surfacing to API clients as a misleading **404**, hiding the real cause.
- No global JAX-RS `ExceptionMapper` exists in calm-hub, so anything else falls through to Quarkus's generic, undiagnostic 500.
- Once a namespace/domain's document for a given resource type hits ~16MB, all further writes of that type to that scope fail outright.

### Environment:
[...]

PR #2897, opened 2026-07-27
feat(calm-hub): add calm collection with startup schema migration runner

markscott-ms:
## Description

A spike on the pathway to #2884, based on a brief discussion with @jpgough-ms.

Adds a `calm` collection (Mongo + Nitrite) that tracks the schema/data version applied to the active storage backend — a precursor to the larger storage redesign discussed in #2884, so future migrations have a version marker to gate on.

`SchemaMigrationRunner` drives startup migrations: it reads the current version from `SchemaVersionStore` and runs pending `SchemaMigrationStep` implementations one version at a time. Steps (currently `MongoIndexInitializationStep` and `NamespaceAccessBackfillStep`) are autowired via CDI's `@All List<SchemaMigrationStep>` and indexed by each step's declared `fromVersion()` — adding a future step requires no changes to the runner itself, and two steps declaring the same `fromVersion()` is a fatal, fail-fast configuration error detected both at build time and run time. Deliberate design choice by @markscott-ms .

Since multiple CalmHub instances can start concurrently against the same shared database, `SchemaVersionStore` also holds a `migrationLock` document guarding the run. The lock has **no automatic timeout** — a migration that fails or runs long requires manual operator resolution rather than risking a second instance racing in on a guessed timeout. Deliberate design choice by @markscott-ms.

`SchemaMigrationInProgressFilter` makes every instance reject requests with 503 while the lock is held (the check is cached briefly to avoid a DB round-trip per request, with the runner deliberately waiting slightly longer than that cache window after acquiring the lock before running any real migration step, so every instance's cache is guaranteed to reflect the lock first). Again, deliberate design choice by @markscott-ms . I intend a subsequent PR that will surface the 503 on CALMHub UI in a nice way.

This went through two rounds of `/code-review`, which found and fixed several real issues: a Mongo lock retry that could throw uncaught on a second concurrent-startup race instead of correctly reporting the lock as unacquired; store exceptions escaping `onStart` and aborting Quarkus startup entirely, contradicting the intended graceful-degradation behaviour; a test-mode skip that silently disabled Mongo index creation even under real-MongoDB integration tests, breaking duplicate-key-detection assertions (confirmed and fixed via an empirical before/after TestContainers run); a wall-clock cache-TTL check vulnerable to clock adjustments; and a filter that could 500 instead of failing open on a transient store error.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-hub/src/main/java/org/finos/calm/migration/SchemaMigrationRunner.java
+++ b/calm-hub/src/main/java/org/finos/calm/migration/SchemaMigrationRunner.java
@@ -69,12 +69,17 @@ import java.util.function.LongConsumer;
  * lets an exception escape and abort Quarkus startup.
  *
  * <h2>Test mode</h2>
- * Under {@code @QuarkusTest} ({@code LaunchMode.TEST}), the lock/version-store machinery
- * above is skipped entirely — most test classes don't back {@link SchemaVersionStore} with
- * a store that supports it. Only steps that opt in via
- * {@link SchemaMigrationStep#runInTestMode()} still run (e.g. {@link MongoIndexInitializationStep},
- * matching its pre-migration-framework behaviour of creating indexes on every startup,
- * including real-MongoDB {@code -P integration} tests, which are {@code @QuarkusTest}s too).
+ * Under {@code @QuarkusTest} ({@code LaunchMode.TEST}), this runner is a no-op — most test
+ * classes don't back {@link SchemaVersionStore} with a store that supports the lock/version
+ * machinery (a mocked, unstubbed {@code MongoDatabase}/{@code Nitrite} would fail on first
+ * real use), and {@code LaunchMode.TEST} can't distinguish those from the real-MongoDB
+ * {@code -P integration} tests that could support it. Steps that genuinely need to run
+ * against a real backend in an integration test (e.g. {@link MongoIndexInitializationStep}
+ * creating the unique indexes {@code MongoDomainIntegration} and similar depend on) are
+ * applied directly by that test's own setup instead — see
+ * {@code src/integration-test/java/integration/EndToEndResource.java}, which has a genuine,
+ * live container to run them against before this runner (or anything else in the app) even
+ * starts.
  */
 @ApplicationScoped
 public class SchemaMigrationRunner {
@@ -127,7 +132,7 @@ public class SchemaMigrationRunner {
 
     void onStart(@Observes StartupEvent ev) {
         if (LaunchMode.current() == LaunchMode.TEST) {
-            runTestModeSteps();
+            LOG.debug("Schema migration skipped in test mode");
             return;
         }
         try {
@@ -140,30 +145,6 @@ public class SchemaMigrationRunner {
             // continuing keeps the instance inspectable instead.
             LOG.error("Unexpected failure while checking/running schema migrations — the application will "
                     + "continue starting, but may be unable to serve requests until this is investigated.", e);
-        }
-    }
-
-    /**
-     * In test mode, the lock/version-store machinery is skipped entirely — most
-     * {@code @QuarkusTest} classes don't back {@link SchemaVersionStore} with a store that
-     * supports it (a mocked, unstubbed {@code MongoDatabase}/{@code Nitrite} would NPE). Only
-     * steps that opt in via {@link SchemaMigrationStep#runInTestMode()} still run, each
-     * defensively isolated so one step's test-mode failure can't affect another's.
-     *
-     * <p>Package-private (not {@code private}) so tests can exercise this directly — {@code
-     * onStart} itself only reaches this branch under a real {@code LaunchMode.TEST}, which
-     * plain unit tests constructing this class directly don't run under.</p>
-     */
-    void runTestModeSteps() {
-        for (SchemaMigrationStep step : stepsByFromVersion.values()) {
-            if (!step.runInTestMode()) {
-                continue;
-            }
-            try {
-                step.apply();
-            } catch (RuntimeException e) {
-                LOG.warn("Schema migration step {} failed in test mode (ignored)", step.getClass().getSimpleName(), e);
-            }
         }
     }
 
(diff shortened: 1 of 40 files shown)


calm-hub/src/main/java/org/finos/calm/store/producer/SchemaVersionStoreProducer.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    @Produces
+    @ApplicationScoped
+    public SchemaVersionStore produceSchemaVersionStore() {
+        if ("standalone".equals(databaseMode)) {
+            return standaloneSchemaVersionStore.get();
+        } else {
+            return mongoSchemaVersionStore.get();
+        }
+    }
```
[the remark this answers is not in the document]
markscott-ms: This is implemented in a consistent manner to the store producer classes. Not changing.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
```diff
@@ -61,33 +65,59 @@
...
 
     private final MongoDatabase database;
 
     @ConfigProperty(name = "calm.database.mode", defaultValue = "mongo")
     String databaseMode;
 
-    public MongoIndexInitializer(MongoDatabase database) {
-        this.database = database;
+    public MongoIndexInitializationStep(MongoDatabase database) {
+        this.database = database.withTimeout(OPERATION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
rocketstack-matt: This 3s client-side operation timeout (added in `0e0e4bef` as a test-startup-time fix) also bounds the real production `createIndex` calls below, not just the test-environment connection attempt — the class javadoc itself confirms it bounds "the entire operation... not just query execution." Combined with `createUniqueIndexes()`/`createAuditIndexes()` swallowing all exceptions internally (lines 195, 227) and this step now being schema-version-gated to run exactly once (vs. the old every-startup behaviour), a single timeout or transient failure on a real, non-trivial existing collection silently and permanently disables unique-index enforcement for that deployment's lifetime — the runner sees no exception, advances the version, and never retries this step again. Worth either a longer/configurable timeout for the real (non-test) path, or surfacing this step's internal failures so the migration isn't marked successful when index creation didn't actually happen.

calm-hub/src/main/java/org/finos/calm/migration/SchemaMigrationRunner.java
```diff
@@ -0,0 +1,217 @@
...
+            if (!step.runInTestMode()) {
+                continue;
+            }
+            try {
+                step.apply();
+            } catch (RuntimeException e) {
+                LOG.warn("Schema migration step {} failed in test mode (ignored)", step.getClass().getSimpleName(), e);
+            }
+        }
+    }
```
rocketstack-matt: Iterates `stepsByFromVersion.values()` (a `HashMap`) in arbitrary order, unlike `runPendingMigrations()` which runs strictly ascending by `fromVersion()`. Harmless with only one test-mode step today, but silently drops the ordering guarantee documented above the class the moment a second step opts into `runInTestMode()`.
```suggestion
    void runTestModeSteps() {
        for (int version : new TreeSet<>(stepsByFromVersion.keySet())) {
            SchemaMigrationStep step = stepsByFromVersion.get(version);
            if (!step.runInTestMode()) {
                continue;
            }
            try {
                step.apply();
            } catch (RuntimeException e) {
                LOG.warn("Schema migration step {} failed in test mode (ignored)", step.getClass().getSimpleName(), e);
            }
        }
    }
```
(needs `import java.util.TreeSet;`)

rocketstack-matt (changes requested):
Solid design overall (lock semantics, fail-open filter, graceful startup handling all check out). One finding I'd consider blocking before merge, plus a minor ordering nit.

calm-hub/src/main/java/org/finos/calm/migration/SchemaMigrationRunner.java
markscott-ms: fair nit that I'd noted too but ignored as it's test. Will fix.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
markscott-ms: Fixed in 26963b02: `MongoIndexInitializationStep` no longer catches its own exceptions — `createUniqueIndexes()`/`createAuditIndexes()` now let failures propagate up through `apply()` to `SchemaMigrationRunner`, which already leaves the schema version unadvanced and the migration lock held on a failed step, so it retries on the next startup instead of silently marking the migration successful.

That made the 3s client-side timeout unnecessary — it existed only to bound the case where this step ran under `@QuarkusTest` with no real Mongo available. Instead, this step no longer runs under `@QuarkusTest` at all; the timeout is removed. The 16 integration tests that rely on unique-index enforcement (duplicate-key rejection etc.) now create the indexes directly against their own TestContainers-provisioned Mongo via a new public `createIndexes()` method, called from `EndToEndResource` before the Quarkus app under test even starts.

calm-hub/src/main/java/org/finos/calm/migration/SchemaMigrationRunner.java
markscott-ms: Fixed in 26963b02, though not with the suggested TreeSet — turned out the whole `runInTestMode()`/`runTestModeSteps()` mechanism this ordering bug lived in had zero remaining callers once the sibling comment's fix landed (see the other thread): index creation for integration tests moved out to `EndToEndResource`, so nothing opts into test-mode execution anymore. Removed `runTestModeSteps()`, the `runInTestMode()` interface method, and its only other override entirely rather than keep unused infrastructure around. `runPendingMigrations()` (the real, production ordering path) was never affected by this bug and is unchanged.

rocketstack-matt (approved):
Both prior findings fixed and verified in 26963b02 — exceptions now propagate from MongoIndexInitializationStep instead of being swallowed, and the buggy test-mode step-ordering mechanism was removed entirely rather than patched. Ran the migration/store test suite locally (65 tests, all green) and CI is green.

Issue #2857, opened 2026-07-22
CalmStudio roadmap H2 2026 — positioning in the CALM platform + proposed workstreams & timelines


initcron:
> [!IMPORTANT]
> **Status 2026-07-28:** Direction has moved on from the workstream table below. Based on the discussion here and at office hours, we are consolidating into **CALM Hub with a studio mode and one shared React visualisation stack** — see [this comment](https://github.com/finos/architecture-as-code/issues/2857#issuecomment-5103561004) for the agreed direction and proposed phases. The body will be rewritten as v2 once the phasing is confirmed.

## Why this issue

CalmStudio has been landing as a series of individual PRs (#2731 viz parity, finos/architecture-as-code#2854 Docusaurus embed, …). Before more workstreams open, maintainers should have one place that says where Studio sits in the platform and what is proposed to ship next, with rough timelines — so priorities can be debated at the weekly call rather than PR-by-PR. Child of the consolidation effort (#2600); refines Track C (#2651).

## Where CalmStudio sits in the platform

| Component | Role | Interface to Studio |
|---|---|---|
| `calm/` (spec + schemas) | The language. Single source of truth for what is valid. | Studio reads/writes canonical CALM 1.x only (nested relationship-type; no local divergences) |
| `calm-cli` + `calm workspace` | Local dev loop + CI validation (#2378) | Studio edits the same files a workspace validates; no competing validation logic |
| CALM Hub (+ Hub UI) | Shared registry and system of record; browse/publish/version | Studio is an authoring client of Hub — planned browse/open/publish/diff integration (Track C item) |
| VSCode extensions (`calm-plugins/vscode`, `calmstudio`) | In-editor authoring/validation for text-first workflows | Peer authoring surfaces; same canonical files, no competing formats |
| CalmStudio | **Visual** (diagram-first) authoring/editing surface. Offline desktop editor; no in-app AI; MCP server as sidecar for external agents | — |
| `@calmstudio/diagram` + `@calmstudio/docusaurus-plugin` | Embed/render layer — CALM diagrams in any docs site (#2853, PR finos/architecture-as-code#2854) | Studio's renderer, packaged for reuse outside Studio |
| CALMGuard | Continuous compliance over CALM in CI/CD | Consumes the same architectures Studio authors; governance surface aligns with finos/calm-schema#26 `governance` property |
| `calm-ai` / MCP servers | Agent integration | External agents drive Studio files via the MCP sidecar (#2562 consolidation) |

[...]

PR #2900, opened 2026-07-28
feat(calm-suite): render composed-of and deployed-in as nested containers

gjs-opsflo:
## Description

`composed-of` and `deployed-in` relationships in `@calmstudio/diagram` now render as **nested container boxes** (elkjs hierarchical layout) instead of dashed arrows — the Mermaid-subgraph-style visual raised at office hours (#2851), where it was noted Mermaid's static render handles nested containers and basic layouts well. Follow-up to #2854 (announced there before opening), part of the W2 visualization workstream in #2857.

How it works:
- A pure **containment planner** resolves containment claims into a parent/child forest: first claim wins (document order), cycle-safe, self/unknown-id-safe. Every unsatisfiable claim falls back to the previous dashed-edge rendering, so no information is silently dropped.
- The ELK graph nests member nodes inside container nodes (`elk.hierarchyHandling: INCLUDE_CHILDREN`, root-relative edge coordinates), and containers draw as translucent rounded boxes (node-type-colored border, name + type label) with members laid out inside. Containers keep the same `data-` attributes as leaf nodes, so click-tooltips keep working; flows referencing a nested containment relationship simply have no edge to animate (member nodes still highlight).
- **Opt-out:** `containers: 'nested' | 'edges'` on `renderELKDiagram`, the `<calm-diagram>` element, and the React `<CalmDiagram>` — `'edges'` reproduces the previous output exactly (verified by differential layout probe). Default is `'nested'`.

Notes for reviewers:
- Default-behavior change for the web component (boxes instead of arrows); the opt-out plus 0.x versioning is the mitigation. Happy to flip the default if preferred.
- A CALM document referencing a nonexistent node id in a relationship fails the render loudly (ELK error) — pre-existing behavior, consistent with the fail-the-build error contract of the docs plugin.
- One benign typecheck-fixture complaint added (`summary` property on a flow-transition test fixture) matching the pre-existing convention in `flowOverlay.test.ts`; the package's 17 pre-existing typecheck errors on `main` are untouched.
- Before/after renders of `calm/architecture/calm-1.json` attached below.

## Type of Change

## Affected Components

- [x] CALM Studio (`calm-studio/`) — `packages/web-component` (renderer), `packages/docusaurus-plugin` (prop pass-through + README)

## Testing

- [x] All existing tests pass

Evidence: `@calmstudio/diagram` 52/52 (14 new: planner unit suite incl. double-claim/cycle/self/unknown-id cases, integration: container bounds containment, 3-level nesting, container-as-edge-endpoint, edges-mode parity, flow-over-containment), `@finos/calm-docusaurus-plugin` 22/22, lint 0 errors — all re-verified after rebasing onto post-#2854 `main` with a fresh `npm ci`.

## Checklist

- [x] I have updated documentation if necessary (plugin README props table)

### DIFF
--- a/calm-studio/packages/web-component/src/render/containment.test.ts
+++ b/calm-studio/packages/web-component/src/render/containment.test.ts
@@ -178,12 +178,7 @@ describe('renderELKDiagram nested containers (integration)', () => {
     expect(container).not.toBeNull();
     const boundsRaw = (container as RegExpExecArray)[1] as string;
     const [cx, cy, cw, ch] = boundsRaw.split(',').map(Number) as [number, number, number, number];
-    // Each member node's x/y must fall inside the container bounds.
-    // NOTE FOR IMPLEMENTER: before finalizing this regex, inspect what
-    // renderNodeSvg actually emits (grep the node group markup for 'svc-a' in a
-    // console.log of the svg) — it may use a translate() transform OR x/y attrs
-    // on the group's <rect>. Adapt ONLY the coordinate-extraction regex to the
-    // real markup; keep these numeric containment assertions exactly as written.
+    // Each member node's x/y (from its <rect>) must fall inside the container bounds.
     for (const member of ['svc-a', 'svc-b']) {
       const m = new RegExp(`data-node-id="${member}"[\\s\\S]{0,200}?<rect x="([\\d.]+)" y="([\\d.]+)"`).exec(svg);
       expect(m).not.toBeNull();
@@ -202,6 +197,19 @@ describe('renderELKDiagram nested containers (integration)', () => {
     // composed-of fan-out (2 members) + connects = 3 polylines
     const polylines = (svg.match(/<polyline/g) ?? []).length;
     expect(polylines).toBe(3);
+  });
+
+  it('an unrecognized containers value falls back to nested (never a silent drop)', async () => {
+    // The 'nested' | 'edges' union is not enforced at the DOM-attribute
+    // boundary, so a typo'd attribute must not make containment vanish.
+    const svg = await renderELKDiagram(nestedArch, {
+      theme: 'light',
+      containers: 'banana' as unknown as 'nested',
+    });
+    expect(svg).toContain('data-container-id="sys"');
+    // connects edge still present; no silent drop of the containment relationship
+    const polylines = (svg.match(/<polyline/g) ?? []).length;
+    expect(polylines).toBe(1);
   });
 
   it('doubly-claimed node: one box plus one fallback arrow', async () => {
--- a/calm-studio/packages/web-component/src/render/containment.test.ts
+++ b/calm-studio/packages/web-component/src/render/containment.test.ts
@@ -1,0 +1,142 @@
+// SPDX-FileCopyrightText: 2024 CalmStudio contributors - see NOTICE file
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it, expect } from 'vitest';
+import { planContainment, isNestedContainment, emptyContainmentPlan } from './containment.js';
+import type { CalmRelationship } from '@calmstudio/calm-core';
+
+function composedOf(uid: string, container: string, nodes: string[]): CalmRelationship {
+  return { 'unique-id': uid, 'relationship-type': { 'composed-of': { container, nodes } } };
+}
+function deployedIn(uid: string, container: string, nodes: string[]): CalmRelationship {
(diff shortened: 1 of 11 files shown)


initcron:
This is how it now looks like with nesting added.
 
Screenshot 2026-07-28 at 12 58 54 PM

calm-studio/packages/web-component/src/render/elkRender.ts
```diff
@@ -62,7 +65,7 @@ export async function renderELKDiagram(
   arch: CalmArchitecture,
   options: RenderOptions = {}
 ): Promise<string> {
-  const { theme = 'light', direction = 'DOWN', flow: flowId } = options;
+  const { theme = 'light', direction = 'DOWN', flow: flowId, containers = 'nested' } = options;
```
rocketstack-matt, line 68: An unrecognized `containers` value (e.g. a typo'd attribute on the raw `<calm-diagram>` custom element — the `'nested'|'edges'` union isn't enforced at the DOM-attribute boundary) falls through to `emptyContainmentPlan()` here *and* fails the `containers === 'edges'` check in the edge filter a few lines down, so `composed-of`/`deployed-in` relationships render with no box and no edge. Verified directly: `renderELKDiagram(arch, {containers:'banana'})` produces zero polylines and zero container boxes. That's a silent drop, contradicting this PR's own design guarantee that unsatisfiable claims always fall back to an edge.

Normalizing once here fixes both call sites:
```suggestion
  const { theme = 'light', direction = 'DOWN', flow: flowId, containers: containersOpt } = options;
  const containers = containersOpt === 'edges' ? 'edges' : 'nested';
```

calm-studio/packages/web-component/src/render/containment.test.ts
```diff
@@ -0,0 +1,286 @@
...
+    const container = /data-container-id="sys"[^>]*data-bounds="([\d.,-]+)"/.exec(svg);
+    expect(container).not.toBeNull();
+    const boundsRaw = (container as RegExpExecArray)[1] as string;
+    const [cx, cy, cw, ch] = boundsRaw.split(',').map(Number) as [number, number, number, number];
+    // Each member node's x/y must fall inside the container bounds.
+    // NOTE FOR IMPLEMENTER: before finalizing this regex, inspect what
+    // renderNodeSvg actually emits (grep the node group markup for 'svc-a' in a
+    // console.log of the svg) — it may use a translate() transform OR x/y attrs
+    // on the group's <rect>. Adapt ONLY the coordinate-extraction regex to the
+    // real markup; keep these numeric containment assertions exactly as written.
```
rocketstack-matt: (non-blocking nit) This reads like an unedited implementer note ("grep the node group markup... inspect what renderNodeSvg actually emits") rather than committed-code prose — no comparable pattern exists elsewhere in this file. Worth trimming before merge.

rocketstack-matt (changes requested):
Nested-container rendering is solid and well-tested (verified the 52/52 test claim and the exact typecheck delta myself), but there's a real information-loss bug in how the `containers` option is validated — see inline comment.

gjs-opsflo:
Thanks @rocketstack-matt — both addressed in `ad032f7b` (branch also rebased onto current main):

- **Silent drop**: normalized once at the top (`containers === 'edges' ? 'edges' : 'nested'`) so the containment plan and the edge filter always agree; any unrecognized value now renders nested. Added a regression test for exactly your repro — `containers: 'banana'` must render the container box and keep the connects edge (fails on the previous code with zero polylines, passes now).
- **Leftover implementer note** in the bounds test: trimmed to normal prose.

Suite: 53/53 (one new test), typecheck unchanged.

rocketstack-matt (approved):
Both review comments addressed in ad032f7b (containers normalization + regression test, trimmed comment). Verified the fix and re-ran a fresh review pass — no new issues.

calm-studio/packages/web-component/src/render/elkRender.ts
rocketstack-matt: Fixed in ad032f7b — normalized `containers` once into `containersMode` so the plan and edge filter always agree, plus a regression test for the exact `containers:'banana'` repro.

calm-studio/packages/web-component/src/render/containment.test.ts
rocketstack-matt: Fixed in ad032f7b — trimmed to normal prose.

PR #2899, opened 2026-07-28
feat(calm-hub-ui): show a Keep CALM modal on 503 responses

markscott-ms:
## Description

Adds a global 503 handler to the shared `apiClient` axios interceptor (already used by every service in the app), so any request rejected while CalmHub is temporarily unavailable — e.g. applying a schema migration, per #2884's `SchemaMigrationInProgressFilter` — surfaces the backend's message to the user instead of failing silently or with a raw error.

Follows the existing `AuthErrorModal`/`auth-store` pattern already in this codebase: a small pub-sub `MigrationStore` holds the current message, a hook subscribes to it, and `MigrationErrorModal` renders it — titled **"Keep CALM"** — when present. Falls back to a generic message if the response body isn't usable text.

This is independent of the backend PR that introduces the 503 (#2897) — it just handles whatever message a 503 response body contains, so it works standalone.

It can be merged before or after #2897.

image

## Type of Change
- [x] ✨ New feature (non-breaking change which adds functionality)

## Affected Components
- [x] CALM Hub UI (`calm-hub-ui/`)

## Testing
- [x] I have tested my changes locally
- [x] I have added/updated unit tests
- [x] All existing tests pass

## Checklist
- [x] My commits follow the [conventional commit format](https://www.conventionalcommits.org/)
- [x] I have updated documentation if necessary
- [x] I have added tests for my changes (if applicable)
- [x] My changes follow the project's coding standards

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2902, opened 2026-07-28
refactor(calm-hub-ui): migrate store hooks to useSyncExternalStore

markscott-ms:
## Description

Follow-up to #2899, implementing a non-blocking review comment from @rocketstack-matt.

`useState` + `useEffect` for external-store subscriptions can miss an update that arrives between the initial render's snapshot and the effect establishing the subscription, under React 18 concurrent rendering. [`useSyncExternalStore`](https://react.dev/reference/react/useSyncExternalStore) is what React exposes specifically to close that gap.

The reviewer noted the concern applied codebase-wide rather than being specific to the new hook, since `use-migration-store.ts` was modelled on the pre-existing `use-auth-store.ts`. This PR migrates both.

Also adds direct unit tests for both hooks (via `renderHook`) — previously they were only ever exercised indirectly through the modal component tests, which mock the hook away entirely.

## Type of Change

## Affected Components

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/calm-hub-ui/src/service/utils/use-auth-store.test.ts
+++ b/calm-hub-ui/src/service/utils/use-auth-store.test.ts
@@ -1,5 +1,5 @@
 import { renderHook, act } from '@testing-library/react';
-import { describe, it, expect, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { useAuthError } from './use-auth-store.js';
 import { authStore } from './auth-store.js';
 
@@ -25,14 +25,16 @@ describe('useAuthError', () => {
         expect(result.current).toBe(403);
     });
 
-    it('stops updating after unmount', () => {
-        const { result, unmount } = renderHook(() => useAuthError());
+    it('unsubscribes from the store on unmount', () => {
+        const unsubscribe = vi.fn();
+        const subscribeSpy = vi.spyOn(authStore, 'subscribe').mockReturnValue(unsubscribe);
+
+        const { unmount } = renderHook(() => useAuthError());
+        expect(unsubscribe).not.toHaveBeenCalled();
+
         unmount();
 
-        act(() => {
-            authStore.setAuthError(401);
-        });
-
-        expect(result.current).toBeNull();
+        expect(unsubscribe).toHaveBeenCalledTimes(1);
+        subscribeSpy.mockRestore();
     });
 });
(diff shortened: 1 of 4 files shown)


calm-hub-ui/src/service/utils/use-auth-store.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        const { result, unmount } = renderHook(() => useAuthError());
+        unmount();
+
+        act(() => {
+            authStore.setAuthError(401);
+        });
+
+        expect(result.current).toBeNull();
+    });
+});
```
rocketstack-matt: This doesn't test what it claims. After `unmount()`, `result.current` is frozen regardless of whether the store listener was actually removed — I verified this by making `subscribeToAuthStore`'s cleanup a no-op and this test still passed. Spy on `authStore.subscribe` and assert the returned unsubscribe fn is actually called (also needs `vi` added to the `vitest` import at the top of the file):
```suggestion
    it('unsubscribes from the store on unmount', () => {
        const unsubscribe = vi.fn();
        const subscribeSpy = vi.spyOn(authStore, 'subscribe').mockReturnValue(unsubscribe);

        const { unmount } = renderHook(() => useAuthError());
        unmount();

        expect(unsubscribe).toHaveBeenCalledTimes(1);
        subscribeSpy.mockRestore();
    });
});
```

calm-hub-ui/src/service/utils/use-migration-store.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        const { result, unmount } = renderHook(() => useMigrationError());
+        unmount();
+
+        act(() => {
+            migrationStore.setMigrationError('migrating');
+        });
+
+        expect(result.current).toBeNull();
+    });
+});
```
rocketstack-matt: Same issue as the auth-store test — after `unmount()`, `result.current` can't change whether or not the subscription was actually removed, so this passes even with a broken cleanup (verified by breaking it locally). Spy on `migrationStore.subscribe` instead (also needs `vi` added to the `vitest` import at the top of the file):
```suggestion
    it('unsubscribes from the store on unmount', () => {
        const unsubscribe = vi.fn();
        const subscribeSpy = vi.spyOn(migrationStore, 'subscribe').mockReturnValue(unsubscribe);

        const { unmount } = renderHook(() => useMigrationError());
        unmount();

        expect(unsubscribe).toHaveBeenCalledTimes(1);
        subscribeSpy.mockRestore();
    });
});
```

rocketstack-matt (changes requested):
The unmount tests for both hooks don't actually verify unsubscription — see inline comments.

rocketstack-matt (approved):

PR #2910, opened 2026-07-29
fix(calm-hub): stop test config shadowing the main application.properties

markscott-ms:
## Description

`calm-hub` has an `application.properties` in **both** `src/main/resources` and `src/test/resources`. During tests one shadows the other wholesale, and whichever wins breaks something:

| which file wins | what breaks |
|---|---|
| **main** | tests lose `calm.audit.*=false`, so the audit filter does live Mongo I/O on every mutating request and writes into whatever MongoDB the developer happens to be running |
| **test** | tests lose `calm.auth.enabled` and the `quarkus.http.filter.security.*` block, so `TestCurrentUserAccessResourceShould` (4) and `TestSecurityResponseHeadersShould` (3) fail in a full `verify` while passing in isolation |

Which side wins varies by environment, so this reads as flaky tests. **It isn't flaky** — locally it failed identically on every run, and passed in isolation every time.

This moves the test settings into the main file under `%test.`, leaving a single `application.properties` so no precedence rule has to hold for the config to be correct.

### This also changes what CI has been testing

CI currently passes on the *other* branch of the same coin — the one where the main file wins. That means **CI has been running the unit suite with audit logging enabled**, contrary to the stated intent of that config block (it's deliberately disabled so ~300 resource tests don't depend on live Mongo). Nobody noticed because CI has no MongoDB for the writes to reach. After this change CI runs with audit disabled, as intended.

### Caveat, stated plainly

The [config reference](https://quarkus.io/guides/config-reference#configuration-sources) says each `application.properties` found is a separate `ConfigSource` merged *per property*, which would make this shadowing impossible. But a deliberately **empty** file in `src/test/resources` shadowed the main one just the same, which rules out any per-property conflict. Documented and observed behaviour disagree on Quarkus 3.34.7 and I could not reconcile them; I also can't explain why the shadowing direction differs between CI and a local machine. The one-file arrangement is proposed precisely because it doesn't depend on resolving that — with a single file there is no precedence rule to get wrong.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Commit Message Format ✅
- [x] Conventional commit format (`fix(calm-hub): ...`)

## Testing
- [x] All existing tests pass

Verified under the condition that previously failed — a MongoDB running on `localhost:27017`, no overrides:

- `../mvnw clean verify` → **2381 tests, 0 failures** (previously 7 failures)
- `../mvnw -P integration verify` → **498 integration tests, 0 failures**, matching the pre-change baseline exactly, confirming the `%test.` scoping doesn't strip anything the integration profile relied on
- `calm.auth.enabled` and the security headers now apply, and the run writes **nothing** to the local database

No new tests: this is a configuration fix, and its correctness is exactly the existing suite passing under a condition that previously broke it.

## Checklist


rocketstack-matt (approved):
Consolidating into one `application.properties` with `%test.`/`%standalone.` prefixes matches the pattern already established in this file, and the diff doesn't drop anything that changes test behavior.

Non-blocking nit (outside this diff so I can't anchor it inline): `IntegrationTestProfile.java:29` and `IntegrationTestSecureProfile.java:31` both have comments referencing `src/test/resources/application.properties`, which this PR deletes. Worth a follow-up to point them at the `%test.` block in `src/main/resources/application.properties` instead.

PR #2912, opened 2026-07-29
feat(vscode): #2869 plugin updates

byrash:
## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📚 Documentation update
- [ ] 🎨 Code style/formatting changes
- [ ] ♻️ Refactoring (no functional changes)
- [ ] ⚡ Performance improvements
- [ ] ✅ Test additions or updates
- [ ] 🔧 Chore (maintenance, dependencies, CI, etc.)

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] CALM Hub UI (`calm-hub-ui/`)
- [ ] CALM Server (`calm-server/`)
- [ ] CALM Widgets (`calm-widgets/`)
- [x] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [x] VS Code Extension (`calm-plugins/vscode/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] I have tested my changes locally
- [x] I have added/updated unit tests
- [x] All existing tests pass

## Checklist
- [x] My commits follow the [conventional commit format](https://www.conventionalcommits.org/)
- [x] I have updated documentation if necessary
- [x] I have added tests for my changes (if applicable)
- [x] My changes follow the project's coding standards

Issue #2869, opened 2026-07-23
CALM VS Code Plugin Updates


byrash:
## Feature Proposal
Update CALM VS Code Plugin to use CALM Hub technology ( React Flow ) and allow editing capability of diagrams in VS Code 

### Target Project:
VS Code Plugin

### Description of Feature:

- Edit CALM Diagrams using VS Code CALM Plugin
- Export CALM Diagrams as SVG ( WYSIWYG )
- Workspace level Building Blocks, Standards & Guidelines Sourcing in Plugin
- Validate CALM Architecture Digram for Controls compliance
- Trigger CALM Agent to produce Solution Design Specification using CALM Agent

### User Stories:
As user i wanted to use vs code plugin of CALM to author, manage & maintain CALM Architecture diagrams in VS Code.

### Current Limitations:
Not able to edit CALM Diagrams in VS Code Extention

### Proposed Implementation:
...provide details of the intended implementation approach, including:
- Technical design considerations
- API changes (if applicable)
- Data model changes (if applicable)
- Dependencies on other components

### Alternatives Considered:
...what other approaches were considered and why were they rejected?

### Testing Strategy:
...how will this feature be tested? Include unit, integration, and any other testing considerations...

### Documentation Requirements:
...what documentation will need to be created or updated?

### Implementation Checklist:
- [ ] Design reviewed and approved
- [ ] Implementation completed
- [ ] Tests written and passing
- [ ] Documentation updated
- [ ] Relevant workflows updated (if needed)
- [ ] Performance impact assessed

PR #2930, opened 2026-08-03
feat(vscode): #2869 plugin updates (merged with main + lockfile fix)

LeighFinegold:
## Purpose

This raises byrash's CALM Canvas VS Code plugin rewrite (finos/architecture-as-code#2912) against `main`, merged with the latest `main` and with build fixes on top, so CI is green.

Raised at byrash's request per the discussion on #2912 (https://github.com/finos/architecture-as-code/pull/2912#issuecomment-5160421287 and the follow-up), where a maintainer offered to patch the build on top of the original work and byrash agreed ("Please do the required").

All of byrash's original commits are retained. Created via `git worktree` off the #2912 PR head, merged with latest `main`, with fix-ups added on top. 

## Commits

- byrash's original PR commits (unchanged): CALM Canvas v1.0.0 visual editor
- Merge of latest upstream `main`
- `fix(deps): regenerate lockfile for vscode plugin dependencies`
- `ci(vscode): remove dead screenshots workflow`
- `fix(deps): pin docusaurus runtime packages to 3.10.1`
- `ci(vscode): drop integration-test job for removed harness`
- `fix(vscode): build extension before packaging`

## Fixes

1. **Lockfile out of sync.** The rewrite added new deps to `calm-plugins/vscode/package.json` (`vite`, `@dagrejs/dagre`, `elkjs`, `lucide-react`, `ajv-formats`, and others) but the root `package-lock.json` was never regenerated, so CI's `npm ci` failed. Regenerated via `rm -rf node_modules package-lock.json && npm install` on Node 26, preserving all platform-specific optional binaries (rollup/oxide/swc/esbuild Linux variants intact for `validate-lockfile`).

2. **`docs` build** (`useTitleFormatterContext called outside <TitleFormatterProvider>`). Regenerating the lockfile let `@docusaurus/theme-mermaid` drift to `3.10.2`, pulling a second `@docusaurus/core` + `theme-common` tree that mismatched the `3.10.1` tree used by the docs preset. Added overrides pinning `@docusaurus/core`, `theme-common`, and `theme-mermaid` to `3.10.1` (single consistent tree, matching `main`; `faster` stays `3.10.2` as on `main`).

3. **VS Code packaging** (`Extension entrypoint(s) missing`). The rewrite changed the `package` script to run `vsce package` without building first and dropped the `vscode:prepublish` hook, so `dist/extension.js` did not exist at package time. Restored `npm run build &&` in the `package` script and re-added `vscode:prepublish` (matching pre-rewrite behaviour).

4. **Orphaned CI.** The rewrite deleted `calm-plugins/vscode/screenshots/` and the `test/integration` harness (plus the `test:integration` script), but the workflows still referenced them:
   - Removed the dead `build-vscode-screenshots.yml` workflow, because its `working-directory` (`calm-plugins/vscode/screenshots`) no longer exists so every run failed before doing any work.
   - Removed the `integration-test` matrix job from `build-vscode-extension.yml` and dropped it from `package-and-publish`'s `needs`, because the job invoked the now-deleted `test:integration` script and could only fail with `Missing script`, which also blocked the packaging job that depended on it.

## Verification (Node 26)

- `npm ci`: pass (was failing)
- VS Code plugin build / test / lint: pass (155 tests, 0 lint errors)
- `docs` build: pass
- `package:vscode`: produces `.vsix`
- Full PR CI on the fork: green

_Note: PR only touches `calm-plugins/` plus the root lockfile/overrides and two VS Code CI workflows. Pre-existing docs/calm-guard/calm-studio issues on `main` are unrelated._

Supersedes / continues #2912. Credit for the feature work goes to @byrash.


calm-plugins/vscode/src/webview/App.tsx
```diff
@@ -0,0 +1,1032 @@
...
+            setNodes([]); setEdges([]); setLastParsedArch(null);
+            lastEmittedJson.current = json;
+            return;
+        }
+        try {
+            const arch = JSON.parse(json) as CalmArchitecture;
+            if (!arch || !arch.nodes || arch.nodes.length === 0) {
+                setNodes([]); setEdges([]); setLastParsedArch(null); return;
+            }
+            setLastParsedArch(arch);
```
rocketstack-matt, line 153: `documentControls` defaults to `{}` and is never hydrated from the loaded document's `controls` field here. `flowToCalm` only writes `arch.controls` back out when `documentControls` is non-empty, so any document with solution-level `controls` silently loses them on the first save.
```suggestion
            setLastParsedArch(arch);
            useCanvasStore.setState({ documentControls: ((arch as Record<string, unknown>).controls as Record<string, unknown>) ?? {} });
```

calm-plugins/vscode/src/webview/transforms/calm-editor-transformer.ts
```diff
@@ -0,0 +1,166 @@
...
+        });
+        const relId = (existingRel?.['unique-id'] as string) ?? `containment-${parentId}`;
+        calmRelationships.push({
+            'unique-id': relId,
+            'relationship-type': { [variant]: { container: parentId, nodes: children } },
+        });
+    }
+
+    // Build edge relationships
+    for (const edge of edges) {
```
rocketstack-matt, line 99: This loop builds one relationship per edge and reuses `data.calmRelId` as `unique-id` for all of them. `calm-parser.ts` splits a single multi-target `interacts` (`actor` + `nodes: [B, C]`) into N edges that all share one `calmRelId` — so a document with a multi-target `interacts` relationship round-trips into N relationships with the *same* `unique-id`, which is invalid per the CALM schema. Needs to group edges by `calmRelId` (for the `interacts` variant) back into one relationship with multiple `nodes` before pushing.

calm-plugins/vscode/src/webview/transforms/calm-parser.ts
```diff
@@ -0,0 +1,382 @@
...
+            height: isContainer ? containerHeight : NODE_HEIGHT,
+        };
+
+        nodes.push(node);
+    }
+
+    // Build edges from non-containment relationships
+    interface ConnectsEntry { relId: string; src: string; dst: string; protocol?: string; description?: string; controls?: unknown; lineStyle: string; routing: string; edgeMetadata: Record<string, unknown> }
+    const connectsList: ConnectsEntry[] = [];
+    for (const rel of arch.relationships ?? []) {
```
rocketstack-matt, line 123: Only `connects`, `interacts`, `deployed-in` and `composed-of` are turned into edges here, and `calm-editor-transformer.ts` rebuilds `relationships` purely from edges with no fallback for anything else. A relationship with any other `relationship-type` (e.g. a decorator-style or custom type) never becomes an edge and is silently deleted the first time the document is saved. Needs a passthrough for unrecognized types from `lastParsedArch` when rebuilding.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
```diff
@@ -0,0 +1,534 @@
...
+            workspaceFolder.uri,
+            'building-blocks'
+        );
+        try {
+            await vscode.workspace.fs.stat(buildingBlocksDir);
+        } catch {
+            await vscode.workspace.fs.createDirectory(buildingBlocksDir);
+        }
+
+        const fileUri = vscode.Uri.joinPath(buildingBlocksDir, filename);
```
rocketstack-matt, line 445: `filename` comes straight from a webview message with no validation before being joined into `building-blocks/`. A crafted filename (e.g. `../../foo.ts`) writes `content` outside `building-blocks/`, potentially outside the workspace entirely — arbitrary file write.
```suggestion
        const safeName = filename.replace(/[/\\]/g, '');
        if (!safeName || safeName !== filename || filename.includes('..')) {
            vscode.window.showErrorMessage(`Invalid building block filename: ${filename}`);
            return;
        }
        const fileUri = vscode.Uri.joinPath(buildingBlocksDir, safeName);
```

calm-plugins/vscode/src/webview/transforms/calm-parser.ts
```diff
@@ -0,0 +1,382 @@
...
+                calmId: id,
+                calmType: n['node-type'] ?? 'system',
+                description: n.description ?? '',
+                interfaces: n.interfaces,
+                controls: n.controls,
+                details: n.details,
+                metadata: n.metadata,
+                containmentType: containerVariant.get(id),
+            },
+            ...(parent ? { parentId: parent, extent: 'parent' as const } : {}),
```
rocketstack-matt, line 107: Nodes parsed from the document get `parentId`/`extent: 'parent'` here but no `zIndex`, whereas `containment.ts#makeContainment` (used when a user manually drags a node into a container) explicitly sets `zIndex: depth`. Without that, a node that's already nested when the file loads can sit at/below its parent container's stacking order, so a drag on the child hits the container underneath instead. This matches manual testing against the TraderX architecture from CALM Hub: some contained nodes aren't draggable — dragging them moves the container (e.g. "Web GUI", "TraderX DB") instead. Needs the same depth-based `zIndex` assigned on initial parse, not just on manual re-parenting.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
```diff
@@ -0,0 +1,534 @@
...
+
+        // 1. Try workspace roots (building-blocks are at workspace root level)
+        for (const folder of vscode.workspace.workspaceFolders ?? []) {
+            candidates.push(path.join(folder.uri.fsPath, filePath));
+        }
+
+        // 2. Try relative to current document
+        if (this.currentDocument) {
+            candidates.push(
+                path.resolve(
```
rocketstack-matt, line 259: `drillInto` resolves `filePath` (from `details["detailed-architecture"]` in the loaded document) via `path.join`/`path.resolve` with no containment check, then reads it — a `../../../etc/passwd`-style path in a malicious `.calm.json` reads arbitrary local files and surfaces them in the canvas. Same class of issue as `workspace-asset-service.ts`'s `resolveStandardProse`. Needs the resolved path canonicalized and checked to stay within the workspace/external-assets roots before reading.

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    /**
+     * Resolve the raw markdown prose for a standard/guideline referenced by a
+     * control requirement URL (e.g. `standards/tls-policy.md`). Searches every
+     * workspace root plus the configured external assets path.
+     */
+    async resolveStandardProse(requirementUrl: string): Promise<string | null> {
+        for (const root of this.getRoots()) {
+            const uri = vscode.Uri.joinPath(root, requirementUrl);
```
rocketstack-matt, line 418: Same path-traversal shape as `canvas-panel.ts`'s `drillInto`: `requirementUrl` comes from an untrusted `requirement-url` field in the document's control requirements and is joined with no containment check, so `../` can escape the standards/guidelines roots. Worth fixing both spots the same way.

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
```diff
@@ -0,0 +1,432 @@
...
+            } catch {
+                /* try next root */
+            }
+        }
+        return null;
+    }
+
+    dispose(): void {
+        if (this.debounceTimer) clearTimeout(this.debounceTimer);
+    }
```
rocketstack-matt, line 431: `dispose()` only clears the debounce timer — `this.watchers` (pushed in `registerWatchers`) is never disposed. Since a `WorkspaceAssetService` is created per `CanvasPanel`, every open/close cycle leaks 6×N watchers for the life of the extension, and the leaked watchers keep firing rescans against already-closed panels.
```suggestion
    dispose(): void {
        if (this.debounceTimer) clearTimeout(this.debounceTimer);
        for (const watcher of this.watchers) watcher.dispose();
        this.watchers = [];
    }
```

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
```diff
@@ -0,0 +1,432 @@
...
+    private toRelativeUrl(root: vscode.Uri, file: vscode.Uri): string {
+        return path
+            .relative(root.fsPath, file.fsPath)
+            .split(path.sep)
+            .join('/');
+    }
+
+    private stem(uri: vscode.Uri): string {
+        const base = path.basename(uri.fsPath);
+        return base.replace(/\.calm\.json$/, '').replace(/\.md$/, '');
```
rocketstack-matt, line 295: `stem()` strips `.calm.json` and `.md` but not `.architecture.json`, even though `scanBuildingBlocks` explicitly globs `*.architecture.json` files too. Building blocks sourced from an `.architecture.json` file get an id like `foo.architecture.json` instead of `foo`, inconsistent with their `.calm.json` siblings.
```suggestion
    private stem(uri: vscode.Uri): string {
        const base = path.basename(uri.fsPath);
        return base.replace(/\.(calm|architecture)\.json$/, '').replace(/\.md$/, '');
    }
```

calm-plugins/vscode/src/extension/webview/html-provider.ts
```diff
@@ -0,0 +1,55 @@
...
+    font-src ${cspSource};
+    connect-src ${cspSource};
+  " />
+  <link rel="stylesheet" href="${styleUri}" />
+  <title>CALM Canvas</title>
+</head>
+<body>
+  <div id="root"></div>
+  <script nonce="${nonce}">
+    window.__INITIAL_CALM_JSON__ = ${initialJson};
```
rocketstack-matt, line 48: `JSON.stringify` doesn't escape `</script>`. If the opened document's raw text contains that literal substring anywhere (e.g. inside a description field), it closes this script block early and injects the remainder as HTML into the page. The CSP nonce blocks a new `<script>` from executing, but not the markup injection itself.
```suggestion
    window.__INITIAL_CALM_JSON__ = ${initialJson.replace(/</g, '\\u003c')};
```

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
```diff
@@ -0,0 +1,534 @@
...
+            document
+        );
+        this.panel.reveal(vscode.ViewColumn.Beside);
+    }
+
+    onDispose(callback: () => void): void {
+        this.disposeCallbacks.push(callback);
+    }
+
+    dispose(): void {
```
rocketstack-matt, line 103: `dispose()` calls `this.panel?.dispose()`, which fires `onDidDispose` synchronously and re-enters `dispose()` before `this.panel = undefined` runs. `disposeCallbacks` isn't cleared, so every registered callback fires twice per panel close. Harmless today since the callbacks look idempotent, but fragile — worth guarding with a `private disposed = false` flag checked at the top of the method.

calm-plugins/vscode/src/extension/extension.ts
```diff
@@ -0,0 +1,91 @@
+import * as vscode from 'vscode';
+import { CanvasPanel } from './webview/canvas-panel';
+import { CalmCanvasCodeLensProvider } from './services/codelens-provider';
+
+let canvasPanel: CanvasPanel | undefined;
+let outputChannel: vscode.OutputChannel;
+
+const CALM_FILE_SUFFIXES = [
```
rocketstack-matt, line 8: This narrows file recognition to a fixed suffix list (`.calm.json`, `.architecture.json`, etc.), dropping the old plugin's looser matching (a plain `architecture.json`, or generic JSON under `calm/`, is no longer picked up). This was flagged by @LeighFinegold in the #2912 discussion as one of the "likely minor" parity gaps but wasn't explicitly agreed as an accepted direction before merge — worth a maintainer decision here rather than merging it as an unstated behavior change.

rocketstack-matt (changes requested):
Reviewed the rewrite plus the 5 build-fixup commits. The fixups (lockfile regen, docusaurus pin, dead-workflow removal, integration-test job removal, build-before-package fix) are correct — verified by rebuilding and packaging the extension from this branch myself. The findings below are in the carried-over feature code, mostly data-loss/correctness bugs on save plus a couple of path-handling issues, and one open scope question that wasn't explicitly resolved in the #2912 discussion.

calm-plugins/vscode/src/webview/App.tsx
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/webview/transforms/calm-editor-transformer.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/webview/transforms/calm-parser.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/webview/transforms/calm-parser.ts
byrash: We shall address and push a fix soon. Thank You. As a side note format in calm hub is using draft version schema and in later iteration we will have to support new version when published in's code extension.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/extension/services/workspace-asset-service.ts
byrash: Adding support for .calm.json, .architecture.json, .template.json, .solution.json, .standard.json, .guideline.json types. Whilst we were wondering on how we could potential have node names enforced for consistency. We talk more about that in our WG. Thanks!

calm-plugins/vscode/src/extension/webview/html-provider.ts
byrash: We shall address and push a fix soon. Thank You.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
byrash: It's harmless today since the callbacks are idempotent, but it's fragile, we will introduce a safe dispose option. Thanks!

calm-plugins/vscode/src/extension/extension.ts
byrash: We have introduced this consciously to allow for where plugin should be activate for. Happy to discuss this in WG. Thanks!

byrash:
Hello @LeighFinegold  for fixes to above noted items, We will have to push changes to our repo. I shall uptake your changes into my code and will push into my PR. Once that's done, I can let you know and you can close this PR. Thank You.

byrash:
Hello @rocketstack-matt all the above mentioned issues are addressed and push into https://github.com/finos/architecture-as-code/pull/2912 . @LeighFinegold  you may now close this PR. Thank You.

LeighFinegold:
> Hello @LeighFinegold for fixes to above noted items, We will have to push changes to our repo. I shall uptake your changes into my code and will push into my PR. Once that's done, I can let you know and you can close this PR. Thank You.

Sure. Sounds good to me.

LeighFinegold:
Closing as @byrash now has managed to update his original PR at source.

PR #2932, opened 2026-08-03
feat: support items-based (zero-or-more) candidate catalogs in pattern decisions

### LINKED ISSUE #2859: Support items-based (zero-or-more) candidate catalogs in pattern decisions
### Target Project:
`shared` (Spectral pattern rules) and `calm-hub-ui` (pattern visualizer)

### Description of Feature:
CALM patterns already allow node/relationship candidates to be declared inside a `prefixItems` slot's `oneOf`/`anyOf` (positional tuple JSON Schema), and decisions (`relationship-type.options`) can reference any of those candidates by unique-id. This works well when a decision is "pick exactly one alternative to fill a specific slot."

It does not work for patterns that need to express "zero or more selections, in any combination, from an open catalog of candidate nodes" — e.g. a set of optional add-on nodes where any subset (including none, including skipping the middle one) may be present. JSON Schema already has the right construct for this: `items: { oneOf: [...] }` with `minItems: 0` validates a catalog of candidates matched against however many array elements are actually present, with no positional constraint. This is standard JSON Schema and requires **no change to the CALM meta-schema** — it already validates correctly today.

The gap is entirely in the tooling: pattern validation (Spectral rules in `shared/src/spectral/`) and pattern rendering (`calm-hub-ui`'s `patternTransformer.ts`) only ever look at `prefixItems`, so a pattern authored with `items`-declared candidates validates a subset of what it should, and renders no decision box at all.

### User Stories:
- As a pattern author, I want to declare a set of optional candidate nodes as an `items`-based catalog (rather than one `prefixItems` slot per candidate), so that my pattern accurately expresses "zero or more, any combination" instead of a fixed-position tuple that doesn't match the real architecture.
- As a pattern author, I want `calm validate` to catch duplicate unique-ids and unreferenced nodes inside an `items`-declared catalog, the same way it already does for `prefixItems`, so I get consistent validation regardless of which construct I use.
- As a calm-hub-ui user, I want a decision that references `items`-declared candidates to render as a visible decision box with its prompt and choices, instead of silently rendering nothing.

### Current Limitations:
- `shared/src/spectral/functions/pattern/ids-are-unique.ts` only scans `nodes.prefixItems[*]` / `relationships.prefixItems[*]` / interface `prefixItems[*]` for unique-id duplication — duplicate ids inside an `items.oneOf`/`anyOf` catalog currently pass validation silently.
- `shared/src/spectral/rules-pattern.ts`'s `pattern-nodes-must-be-referenced` rule's `given` selector is scoped to `nodes.prefixItems[*]` only, so an unreferenced `items`-declared node never triggers the existing "must be referenced" warning.
- `calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts`'s `getPrefixItems` helper (and everything downstream of it — `extractNodesFromPattern`, `extractRelationshipsFromPattern`) only ever reads `prefixItems`. A decision's metadata (`parsePatternData`'s `nodeToGroupMap`/`groupOptionsMap`) is only ever *attached* to a pre-existing node-level decision group built from `prefixItems` oneOf/anyOf wrapping — it never creates a new group, so a decision referencing only `items`-declared ids currently computes its metadata and then silently drops it, rendering no box at all.

### Minimal Example:
A pattern with two mandatory nodes (`webapp`, `database`) plus an open catalog of optional add-ons (`cache`, `queue`) declared via `items` instead of `prefixItems`, and a decision that lets an architecture instance pick any subset of those add-ons:

```json
{
  "$schema": "https://calm.finos.org/release/1.0/meta/calm.json",
  "properties": {
    "nodes": {
      "type": "array",
      "prefixItems": [
        {
          "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/node",
          "properties": {
            "unique-id": { "const": "webapp" },
            "node-type": { "const": "service" },
            "name": { "const": "Web Application" },
            "description": { "const": "Front-end web application" }
          }
        },
        {
          "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/node",
          "properties": {
            "unique-id": { "const": "database" },
            "node-type": { "const": "database" },
            "name": { "const": "Database" },
            "description": { "const": "Primary data store" }
          }
        }
      ],
      "items": {
        "oneOf": [
          {
            "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/node",
            "properties": {
              "unique-id": { "const": "cache" },
              "node-type": { "const": "service" },
              "name": { "const": "Cache" },
              "description": { "const": "Optional caching layer" }
            }
          },
          {
            "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/node",
            "properties": {
              "unique-id": { "const": "queue" },
              "node-type": { "const": "service" },
              "name": { "const": "Queue" },
              "description": { "const": "Optional async messaging layer" }
            }
          }
        ]
      },
      "minItems": 2
    },
    "relationships": {
      "type": "array",
      "prefixItems": [
        {
          "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/relationship",
          "properties": {
            "unique-id": { "const": "webapp-to-database" },
            "relationship-type": {
              "const": {
                "connects": {
                  "source": { "node": "webapp" },
                  "destination": { "node": "database" }
                }
              }
            }
          }
        },
        {
          "$ref": "https://calm.finos.org/release/1.0/meta/core.json#/defs/relationship",
          "properties": {
            "unique-id": { "const": "optional-add-ons" },
            "description": { "const": "Choose any optional add-ons for this deployment" },
            "relationship-type": {
              "properties": {
                "options": {
                  "prefixItems": [
                    {
                      "anyOf": [
                        {
                          "properties": {
                            "description": { "const": "Add a caching layer" },
                            "nodes": { "const": ["cache"] },
                            "relationships": { "const": [] }
                          }
                        },
                        {
                          "properties": {
                            "description": { "const": "Add an async messaging queue" },
                            "nodes": { "const": ["queue"] },
                            "relationships": { "const": [] }
                          }
                        }
                      ]
                    }
                  ]
                }
              }
            }
          }
        }
      ]
    }
  }
}
```

`webapp` and `database` are mandatory; `cache` and `queue` are an open catalog — zero, one, or both may appear, in any combination. The `optional-add-ons` decision lets an architecture instance pick either, both, or neither add-on.

**Observed today:**
1. **Validation gap** — duplicate a `unique-id` inside `nodes.items.oneOf` (e.g. give `queue` the id `cache` by mistake) and run `calm validate`: `unique-ids-must-be-unique-in-pattern` does not catch it, because its JSONPath queries only scan `nodes.prefixItems[*]`. The same duplicate inside a `prefixItems` slot is correctly rejected.
2. **Rendering gap** — load this pattern in calm-hub-ui: `webapp`, `database`, and the connecting edge render correctly, but the `optional-add-ons` decision box never appears, and `cache`/`queue` are never drawn at all. `patternTransformer.ts`'s extraction (`getPrefixItems`) only reads `prefixItems`, so the `items`-declared candidates and the decision's prompt/choices are silently dropped rather than rendered.

### Proposed Implementation:
- **`ids-are-unique.ts`**: extend the three JSONPath queries to also cover `items.oneOf[*]`/`items.anyOf[*]` equivalents for nodes, relationships, and node interfaces.
- **`rules-pattern.ts`**: widen the `given` selector on `pattern-nodes-must-be-referenced` to also walk `nodes.items.oneOf[*]`/`anyOf[*]`. No function change needed — `node-has-relationship.ts`'s function body already scans the whole document unscoped.
- **`patternTransformer.ts`**:
  - Extend node/relationship extraction to also read `items.oneOf`/`anyOf` candidates (in addition to the existing `prefixItems` path), producing `ExtractedNode`/`ExtractedRelationship` entries the same way `prefixItems` candidates do today.
  - In `parsePatternData`, change the decision-to-group mapping from "look up the first existing group that matches" to: fold any decision-referenced ids not already in a group into the matching group (if one exists), or create a new decision group from the decision's own id list if none of its referenced ids match an existing group. Only fold in/create using ids that resolve to a real extracted node — ids that don't resolve to anything (e.g. a typo) must be skipped so a dangling reference still renders as nothing, matching today's behavior, rather than producing an empty box.
  - Thread the options-relationship's own `unique-id` through `OptionsMetadata` so a newly created group has a stable, deterministic id (rather than one derived from array index).

### Alternatives Considered:
A new CALM vendor-extension construct for expressing optional candidate sets was considered and rejected: decisions already reference candidates by unique-id, and `items` already expresses "zero or more, any combination" natively in JSON Schema. No schema change is needed; this is purely a tooling gap.

### Testing Strategy:
- `shared`: new Spectral rule-function unit tests for `ids-are-unique` and `pattern-nodes-must-be-referenced` covering `items`-declared duplicate/unreferenced ids; confirm `nodes-referenced-in-pattern-decision-must-be-in-oneof-or-anyof-block` and the two options-shape rules are unaffected (existing tests should continue to pass unchanged).
- `calm-hub-ui`: new fixtures/tests in `patternTransformer.test.ts` covering (a) a decision referencing only `items`-declared candidates (new group created), (b) a decision referencing a mix of `prefixItems`- and `items`-declared candidates (fold-in), (c) a decision referencing a dangling/typo'd id (must still render nothing, not an empty box). Confirm the existing `handles options metadata on decision groups` test is unaffected.
- Since `shared` is a dependency of `cli` and the VSCode extension, run the full `npm test` from repo root after the `shared` changes, not just `--workspace shared`.

- Update pattern-authoring docs to describe `items`-based optional candidate catalogs as a supported alternative to `prefixItems`-per-slot, with a worked example.
- Release notes: call out the one behavior change with real teeth — duplicate unique-ids inside an `items` catalog, previously silent, now fail `calm validate` with an error. The "must be referenced" extension is a `warn`-level addition and lower priority to call out but should still be mentioned.

### Implementation Checklist:
- [ ] Documentation updated

### Additional Context:
No CALM meta-schema changes are required — `items` is already valid, already-supported JSON Schema. This is entirely a tooling gap between what CALM's schema foundation already permits and what `shared`'s Spectral rules and `calm-hub-ui`'s renderer currently understand.

YoofiTT96:
## Description

Lets a CALM pattern declare "zero or more of these candidates" through a JSON Schema `items.oneOf`/`items.anyOf` open catalog, alongside the existing `prefixItems` positional slots. No meta-schema change is needed — `items` is standard JSON Schema.

Closes #2859.

The reason this touches so much is that three surfaces read pattern decisions independently, and all three were `prefixItems`-only: validation (`shared/src/spectral`), generation (`shared/src/commands/generate`) and visualisation (`calm-hub-ui`). Nothing made them agree, which is where most of the defects below came from. There is now a fixture contract in `test_fixtures/decision-agreement/` that both `shared` and `calm-hub-ui` assert against, so the two cannot drift apart silently again.

### Decisions worth knowing about

**A decision holder must live in `properties.relationships.prefixItems`.** Candidates belong in the catalog; the decision that selects them does not. A holder placed in a catalog instead makes the decision itself optional, which is incoherent, so `calm validate` now rejects it — a new pattern rule, not a generate-time check. `calm generate` never validates, so a user who only generates still gets silence: the decision is simply never offered, same as before. The rule is the fix; enforcement is one command away from where the mistake was made, and that's documented, not accidental.

**`calm generate` never validates**, so it carries its own guard. Choosing a candidate that selection cannot reach now throws instead of quietly producing an architecture missing what you asked for.

**A candidate can be drawn in one box only.** When two decisions name the same candidate, the first keeps it and the second still renders with what is left. Boxing one node twice needs the nesting rework in #2933.

**A decision whose every candidate shares one container loses its box, and its prompt with it.** Container membership beats decision-group membership when a node is both, so a `oneOf`/`anyOf` between two services deployed into the same container renders no box at all — the choice becomes invisible in the diagram, not just unstyled. Deliberate and directly tested, and the flat interim `#2933`'s nesting rework replaces: nesting the box inside the container is the real fix, not something this PR attempts.

**`allOf` for nodes and relationships remains unsupported.** Three parts of the system read it differently, all deliberately. Documented in `shared/AGENTS.md`; convergence is separate work.

**The CLI and the visualiser already disagreed on a decision holder declaring both `oneOf` and `anyOf`, before this PR.** Confirmed against `main`: the CLI offers every choice from both keywords, the visualiser draws only the `oneOf` half. Neither reader was touched here beyond the visualiser's hand-rolled check being swapped for the shared `readChoiceBlock` — same rule, same result. Not this PR's to resolve; documented in `shared/AGENTS.md`'s "Still duplicated" table.

### Behaviour changes

| Change | Before |
|---|---|
| duplicate `unique-id` in an `items` catalog errors | passed silently |
| duplicate `unique-id` in a `oneOf`/`anyOf` slot errors | invisible |
| interface checks see every alternative in a slot | first alternative only; one alternative could borrow another's interfaces |
| duplicate-id errors carry a source position | reported at line 0 |
| a decision naming an unreachable candidate errors | passed |
| a block declaring both `oneOf` and `anyOf` warns | silent, and the `anyOf` candidates were unreachable |
| `calm generate` rejects an unreachable answer | discarded the answer, generated without it |
| `instantiate` emits `[]` for an array with no `prefixItems` | `{}`, invalid where an array is required |
| junit output lists two more rules | — |
| two decisions over one catalog draw two boxes | one box, one prompt lost |
| a decision between containers keeps its box | vanished entirely |
| `calm validate` checks every selected answer of a multi-answer (`anyOf`) decision | only the first recorded answer was applied to the pattern before validating; a second selected candidate went unchecked |

The dual-keyword row above is coverage plus a message fix, not a bug fix: the rule already caught this on a `prefixItems` slot before this PR touched it, and catching it on an `items` catalog too was the intended expansion — only its wording said "items catalog" for both cases, so a `prefixItems` finding read as a false description of where the problem was.

The multi-answer row above is a strictness increase, not just new coverage: an architecture with a bad second answer used to validate clean and will now fail. Worth calling out in the release note alongside the pattern-side changes, since it can surface on an architecture nobody has touched.

Nothing in this repo uses `items` catalogs today, so nothing shipped regresses. Externally authored patterns carrying any of the newly detected faults will start failing `calm validate`. Worth a release note on the `shared`/CLI bump.

### Fixed since first review

Two independent review passes surfaced defects this PR introduced after it was opened; all are fixed on this branch now. (The multi-answer validation fix is the same class of finding, but it earns its own row in the behaviour table above, not a line here — it's a strictness increase worth the extra visibility.)

- The pattern array reader could compose a `nodes`/`relationships` array from two different `allOf` branches, describing an array no single declaration site actually contains.
- `calm diff`, the CALM Hub's visual pattern diff, and the VSCode pattern preview were all blind to `items`-catalog candidates: a catalog addition or removal produced no diff at all, and a catalog-only pattern previewed as empty. All three now see catalog candidates the same way generation and validation already did.
- Leaving one decision fully unanswered while answering another broke `calm validate` with an internal schema-compilation error, because the narrowed pattern ended up with an empty `prefixItems`, which is not legal JSON Schema. "Pick none from this catalog" is this feature's own advertised case, so the decision holder is now dropped instead of narrowed to empty.

### Deferred

Nesting decision boxes and containers stays in #2933, along with the case where two decisions share every candidate. The container-precedence behaviour here is the flat interim that #2933 supersedes.

## Type of Change

- [x] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [x] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
```diff
@@ -621,6 +713,102 @@ function applyPatternLayout(regularNodes: Node[], groupNodes: Node[], edges: Edg
...
+            }
+
+            targetGroup.nodeIds.push(id);
+            nodeToGroupMap.set(id, targetGroup.groupId);
+
+            const node = nodesById.get(id);
+            if (node) node.decisionGroupId = targetGroup.groupId;
+        });
+
+        groupOptionsMap.set(targetGroup.groupId, meta);
```
markscott-ms, line 760: **Two decisions over one catalog silently lose one of them.**

All catalog candidates land in the single group `node-decision-items`. If a pattern declares one catalog and two separate options-relationships each referencing a subset — a natural way to author "pick a cache" + "pick a queue" from one optional catalog — both fold into that same group and this `set` overwrites: the first decision's prompt and choices vanish from the UI, and all candidates render in one box labelled with the second decision.

The `prefixItems` world couldn't hit this, since each slot was its own group. Note this also flips the old `!groupOptionsMap.has(groupId)` "first wins" to "last wins".

Options: allocate a fresh group when the matched target already carries metadata from a different relationship, or at minimum document the one-decision-per-catalog constraint — the current guidance in `pattern-creation.md` only covers the converse (one decision's candidates spanning two declaration sites).

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
```diff
@@ -398,8 +471,28 @@ function createReactFlowNodes(
...
-    // Create decision group parent nodes
+    // Determine each non-container node's parent. A container (deployed-in /
+    // composed-of) takes precedence over a decision group: a node that is both an
+    // optional decision candidate AND a container child renders inside its
+    // container, not in the choice box. Track which decision groups actually keep
+    // at least one child after that precedence is applied.
+    const effectiveParent = new Map<string, string>();
+    const usedDecisionGroupIds = new Set<string>();
+    extractedNodes.forEach((node) => {
+        if (containerNodeIds.has(node.uniqueId)) return;
```
markscott-ms: **A decision group whose candidates are all containers now disappears entirely.**

Container nodes are skipped here, so their `decisionGroupId` never reaches `usedDecisionGroupIds` and the group is dropped by the new guard below. For a `oneOf` between two nodes that are both `deployed-in` containers, the prompt/choices box previously rendered (empty, but carrying the decision text); now nothing at all indicates the two containers are alternatives.

Arguably inside the deferred "decision box inside/around a container" issue, but it's an information regression rather than just empty-box suppression, so worth calling out explicitly on that follow-up.

shared/src/commands/generate/components/options.ts
```diff
@@ -119,12 +119,39 @@ function flattenOneOfAndAnyOf(item: Item, selectionPredicate: (item: SchemaNode)
...
+ *
+ * The selected catalog candidates are appended onto `prefixItems` so that
+ * `instantiate()` - which only understands `prefixItems` - materializes them
+ * the same way it already does for the rest of the array. `items` is then
+ * cleared, since the pattern's array is now fully expressed via `prefixItems`.
+ */
 function flattenCalmItems(pattern: SchemaNode, calmType: 'nodes' | 'relationships', ids: string[]): void {
-    const calmItems = pattern['properties'][calmType]['prefixItems'];
+    const calmProps: SchemaNode = pattern['properties']?.[calmType];
+    if (!calmProps) return;
```
markscott-ms, line 128: This guard only covers half the path: `flattenOptionsRelationships` a few lines below still does `pattern['properties']['relationships']['prefixItems'].map(...)` unguarded, so a pattern with no `relationships` property still throws — just later, and with a less obvious stack. Either guard both or drop this one, since as it stands it reads as protection the caller doesn't actually get.

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.test.ts
```diff
@@ -55,6 +55,61 @@ function makePattern(
     };
 }
 
+// Helper to build a pattern with both prefixItems (mandatory) and an
+// items.oneOf/anyOf open catalog for nodes and/or relationships.
+function makePatternWithItems(
+    prefixNodes: unknown[],
+    itemsCatalogNodes: unknown[] = [],
+    relationships: unknown[] = [],
+    itemsCatalogRelationships: unknown[] = [],
```
markscott-ms, line 64: Both of these parameters are unused — no test passes `itemsCatalogRelationships`, and `catalogType` is never overridden to `'anyOf'`.

That leaves the `rel-decision-items` branch of `extractRelationshipsFromPattern` (and the dashed/decision-coloured edges it produces) with no coverage at all, and the UI-side `anyOf` catalog path untested — the `shared` side does cover `anyOf` via `ids-are-unique.spec.ts`. Either add the tests these params were built for or drop the params.

shared/src/commands/generate/components/options.spec.ts
```diff
@@ -126,6 +126,27 @@ function buildPattern(nodes: object[], relationships: object[]) {
     };
 }
 
+// Builds a pattern with mandatory prefixItems plus an open items.oneOf/anyOf catalog
+function buildPatternWithItemsCatalog(
+    mandatoryNodes: object[],
+    catalogNodes: object[],
+    relationships: object[] = [],
+    catalogRelationships: object[] = []
```
markscott-ms, line 134: `catalogRelationships` is never passed by any test, so the relationships-side `items` catalog in `flattenCalmItems` is uncovered here too. Worth a case that selects a catalog relationship and asserts it's moved into `relationships.prefixItems` with `items` deleted.

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
```diff
@@ -42,6 +42,25 @@ function getPrefixItems(pattern: SchemaObject, key: string): SchemaObject[] {
     return [];
 }
 
+/**
+ * Gets the `items` catalog schema (the `items.oneOf`/`items.anyOf` open-catalog
+ * declaration) for a given top-level key (e.g. 'nodes' or 'relationships')
+ * from a pattern, handling allOf structures.
+ */
+function getItems(pattern: SchemaObject, key: string): SchemaObject | undefined {
```
markscott-ms: Nit: this duplicates `getPrefixItems`'s allOf-walking shape verbatim — a shared `getArrayKeyword(pattern, key, keyword)` would collapse both.

Similarly, the two `if (items) { hasOneOf / hasAnyOf ... }` blocks added to `extractNodesFromPattern` and `extractRelationshipsFromPattern` are near-identical; a small `catalogAlternatives(items)` returning `{ groupType, alternatives } | null` would remove the repeated `Array.isArray` pairs at both sites.

shared/src/commands/generate/components/flatten-allof.ts
```diff
@@ -41,6 +41,11 @@ function deepMergeSchemas(
                 value as unknown[]
             );
         } else {
+            // `items` (the open oneOf/anyOf catalog) falls through here and is not
+            // deep-merged the way `prefixItems` is above - a later allOf branch's
+            // `items` simply replaces an earlier one. Realistic CALM patterns declare
+            // the catalog once per array, so this is not expected to matter in
+            // practice; revisit if patterns start composing `items` across allOf.
```
markscott-ms: Fine as a documented limitation, but the docs in this PR now actively encourage `items` catalogs, which makes "not expected to matter in practice" a bit less safe than it was. A pattern composing two allOf branches that each declare a catalog silently loses one. Consider a debug-level log when `key === 'items'` and both sides are present, so the failure is discoverable from `--verbose` rather than only from reading this comment.

markscott-ms (commented):
## Review

Adds `items.oneOf`/`items.anyOf` open-catalog support across the three surfaces that read pattern decisions independently. The core approach — narrowing a selected catalog into `prefixItems` inside `selectChoices` so `instantiate()` needs no new concept — is the right call and keeps the generation change small. Good comments explaining *why*, sensible decomposition (`extractNodeDecisionGroup`, `extractRelationshipDecisionGroup`, `foldOptionsMetadataIntoDecisionGroups`), and an honest write-up of the behaviour changes.

Most of my findings are gaps rather than defects in what's written; they're left inline. Two things that don't anchor to a diff line:

### `isPatternData` in `calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx:21` wasn't updated

It requires `properties.nodes.prefixItems` to classify JSON as a pattern. A catalog-only pattern — nodes declared solely through `items`, a shape this PR explicitly supports and covers in `patternTransformer.test.ts` ("creates a decision group box for an items catalog with no options relationship") — fails that check and gets routed as an architecture. Hub-sourced patterns are saved by the `calmType === 'Patterns'` fallback on line 80; the file-upload path isn't. One-line fix: accept `items` as well as `prefixItems`.

### `isDefinedInOneOfOrAnyOf` deliberately needing no change is worth a sentence in the PR body

It's the one decision-related pattern rule left untouched, so a reader will wonder. It only fires when an id is present in plain `prefixItems` and absent from any `oneOf`/`anyOf`, so a catalog-declared id passes rather than false-positives. Correct as-is, just non-obvious.

### Verified as fine

- `nodeHasRelationship`'s `$..relationship-type..*@string()` already sweeps catalog relationships and decision `nodes.const` arrays, so the widened `given` in `rules-pattern.ts` behaves as the PR body claims.
- Ordering in `flattenCalmItems` (`[...flattenedPrefixItems, ...selectedCatalogItems]`) is correct against draft 2020-12: selected candidates land at indices >= the original `prefixItems` length, exactly where the original pattern's `items` applies, so the generated architecture still validates against the unmodified pattern.
- The `const`-before-array reordering in `instantiate.ts` is right, and the dedicated test pinning that branch order is a nice touch.

### Test coverage

Solid on the paths that matter, with two new branches uncovered — see the inline notes on the two test helpers. Also worth a test for the multi-decision-over-one-catalog case, whichever way it's resolved, since it pins the intended semantics.

### Risk

Low. The stricter validation only affects a construct nothing in the repo uses, and is already flagged for a release note. The `instantiate` `[]` fallback is the only change touching pre-existing behaviour, and `{}` where an array was required was never valid output — a strict improvement.

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
YoofiTT96: Noted in #2933 as a follow up since a larger issue around handling of decisions spanning containers needs to be addressed

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
YoofiTT96: Noted. Will add to #2933 scope

markscott-ms:
@YoofiTT96 let us know when this is good for re-review

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.test.ts
YoofiTT96: Fixed

shared/src/commands/generate/components/options.ts
YoofiTT96: Fixed

shared/src/commands/generate/components/options.spec.ts
YoofiTT96: Fixed

shared/src/commands/generate/components/flatten-allof.ts
YoofiTT96: Updated PR body

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
YoofiTT96: Consolidated to shared function

markscott-ms:
## Code review findings

Automated review of the items-based decision support changes turned up 6 findings, most-severe first:

1. **`shared/src/commands/generate/components/options.ts:53`** — `getRelationshipsPrefixItems` (used by `extractOptions`, which feeds both interactive prompting and `--option-choices`) never reads `relationships.items`. A decision declared *only* via an items-catalog relationship (per this PR's own docs: "Relationship candidates can use an items catalog in the same way" as nodes) is silently never offered to the user by `calm generate`.

2. **`calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts:808`** — `foldOptionsMetadataIntoDecisionGroups` overwrites `groupOptionsMap` when two independent decisions resolve to the same pre-existing group (e.g. a shared items catalog collapsed to one group, referenced by two separate options-relationships). The second decision's prompt/choices silently clobber the first's.

3. **`shared/src/commands/generate/components/flatten-allof.ts:1005`** — `logDroppedItemsCatalogs` only fires when *both* `allOf` branches declare an items catalog on the same array property. The more common asymmetric collision (one branch has `items`, the other only `prefixItems`) still silently drops the catalog via `deepMergeSchemas`' shallow `properties` spread, with zero diagnostic — even under `--debug`.

4. **`calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts:778`** — when a single decision's choices span two pre-existing decision groups, `matchedGroupIds[0]` picks the surviving group (and its `groupType`) based on incidental array ordering in the pattern author's `choices[].nodes`, not any explicit rule. Lower severity — this only triggers for the "candidates split between prefixItems and items" shape the PR's own new docs call out as a pattern smell to avoid.

5. **`shared/src/spectral/functions/pattern/ids-are-unique.ts:12`** — the `items.oneOf`/`items.anyOf` lookup for nodes, relationships, and interfaces is hand-duplicated across `ids-are-unique.ts`, `node-id-exists.ts`, `interface-id-exists-on-node.ts`, and `rules-pattern.ts`, with no shared helper. A future change to how catalog nodes are located needs the same hand-edit in 4+ places with nothing forcing them to stay in sync.

6. **`shared/src/commands/generate/components/options.ts:141`** — the "oneOf wins over anyOf" catalog-alternative selection rule is reimplemented independently in `options.ts` (inline) and in `patternTransformer.ts`'s own `catalogAlternatives()` helper, rather than sharing one implementation. `shared` (CLI) and `calm-hub-ui` (visualizer) can silently disagree on which alternatives a catalog offers if this precedence rule ever changes in one place but not the other.

---

Taken together, these findings share a pattern: the "candidate via items catalog" concept was added in several independent places (CLI generation, pattern linting, and the Hub UI visualizer) with the core lookup/precedence/grouping logic reimplemented by hand in each, rather than sharing one source of truth. That's what's producing both the gaps (finding 1, and the asymmetric-merge case in finding 3) and the fragility (findings 2, 4-6) above.

Given how central "options"/"decisions" are to CALM pattern generation, it might be worth taking a step back and writing a short design doc that pins down the intended current and future behaviour for options - precedence rules (oneOf vs anyOf), how things merge across `allOf`, how decisions fold into UI groups, and where the canonical implementation of each rule should live - before this area grows further. That would give a shared reference to design future changes against and to catch regressions like these in review, rather than each package evolving its own interpretation independently.

YoofiTT96:
> ## Code review findings
> Automated review of the items-based decision support changes turned up 6 findings, most-severe first:
> 
> 1. **`shared/src/commands/generate/components/options.ts:53`** — `getRelationshipsPrefixItems` (used by `extractOptions`, which feeds both interactive prompting and `--option-choices`) never reads `relationships.items`. A decision declared _only_ via an items-catalog relationship (per this PR's own docs: "Relationship candidates can use an items catalog in the same way" as nodes) is silently never offered to the user by `calm generate`.
> 2. **`calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts:808`** — `foldOptionsMetadataIntoDecisionGroups` overwrites `groupOptionsMap` when two independent decisions resolve to the same pre-existing group (e.g. a shared items catalog collapsed to one group, referenced by two separate options-relationships). The second decision's prompt/choices silently clobber the first's.
> 3. **`shared/src/commands/generate/components/flatten-allof.ts:1005`** — `logDroppedItemsCatalogs` only fires when _both_ `allOf` branches declare an items catalog on the same array property. The more common asymmetric collision (one branch has `items`, the other only `prefixItems`) still silently drops the catalog via `deepMergeSchemas`' shallow `properties` spread, with zero diagnostic — even under `--debug`.
> 4. **`calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts:778`** — when a single decision's choices span two pre-existing decision groups, `matchedGroupIds[0]` picks the surviving group (and its `groupType`) based on incidental array ordering in the pattern author's `choices[].nodes`, not any explicit rule. Lower severity — this only triggers for the "candidates split between prefixItems and items" shape the PR's own new docs call out as a pattern smell to avoid.
> 5. **`shared/src/spectral/functions/pattern/ids-are-unique.ts:12`** — the `items.oneOf`/`items.anyOf` lookup for nodes, relationships, and interfaces is hand-duplicated across `ids-are-unique.ts`, `node-id-exists.ts`, `interface-id-exists-on-node.ts`, and `rules-pattern.ts`, with no shared helper. A future change to how catalog nodes are located needs the same hand-edit in 4+ places with nothing forcing them to stay in sync.
> 6. **`shared/src/commands/generate/components/options.ts:141`** — the "oneOf wins over anyOf" catalog-alternative selection rule is reimplemented independently in `options.ts` (inline) and in `patternTransformer.ts`'s own `catalogAlternatives()` helper, rather than sharing one implementation. `shared` (CLI) and `calm-hub-ui` (visualizer) can silently disagree on which alternatives a catalog offers if this precedence rule ever changes in one place but not the other.
> 
> Taken together, these findings share a pattern: the "candidate via items catalog" concept was added in several independent places (CLI generation, pattern linting, and the Hub UI visualizer) with the core lookup/precedence/grouping logic reimplemented by hand in each, rather than sharing one source of truth. That's what's producing both the gaps (finding 1, and the asymmetric-merge case in finding 3) and the fragility (findings 2, 4-6) above.
> 
> Given how central "options"/"decisions" are to CALM pattern generation, it might be worth taking a step back and writing a short design doc that pins down the intended current and future behaviour for options - precedence rules (oneOf vs anyOf), how things merge across `allOf`, how decisions fold into UI groups, and where the canonical implementation of each rule should live - before this area grows further. That would give a shared reference to design future changes against and to catch regressions like these in review, rather than each package evolving its own interpretation independently.

1, 2, 3, 5, 6 Addressed
4 Deferred to be handled in follow up issue that deals with containment

YoofiTT96:
@markscott-ms PR updated with changes for re-review

markscott-ms:
@YoofiTT96 A few comments:

I'm concerned with the blast radius of this one PR - it combines updates to:

1. The AI skill content
2. The CALM Hub UI
3. The CALM models library
4. The VSCode extension
5. The CLI
6. The shared library used by CLI and hub?

Any chance we can break it out? In particular, the AI skill content and VSCode extension change could be standalone?

YoofiTT96:
> @YoofiTT96 A few comments:
> 
> I'm concerned with the blast radius of this one PR - it combines updates to:
> 
> 1. The AI skill content
> 2. The CALM Hub UI
> 3. The CALM models library
> 4. The VSCode extension
> 5. The CLI
> 6. The shared library used by CLI and hub?
> 
> Any chance we can break it out? In particular, the AI skill content and VSCode extension change could be standalone?

Sure do that. Some of the shared changes were due to consolidation of the different local implementations of functionality across the different surfaces (validation, generation, visualisation) that posed a long term drift risk (part of the review comments raised). However, I think it will easier to split this feature addition by performing the same hand implementation for correctness guarantees across multiple prs and then have one issue to consolidate the implementations where possible. That way we gain confidence for the correctness and blast radius and eventually resolve the risk. Will raise subsequent PRs. Any thoughts on this before then are welcome

PR #2934, opened 2026-08-03
feat(calm-hub): versioned artefact storage redesign — Pattern (#2884)

markscott-ms:
## Description

**This is a direct continuation of [#2923](https://github.com/finos/architecture-as-code/pull/2923)**, which implemented the new Architecture storage. It is intended to be largely a replication of that work for **Pattern**, following the same shape, the same helpers and the same per-type checklist — so if you reviewed #2923, most of this should look familiar and the diff is best read as "the same thing again, for patterns".

Three places it is deliberately *not* a straight replication, each called out in its own section below:

1. **The migration logic is now shared** rather than copied per type — Pattern would have been the second near-identical copy.
2. **Pattern guards `name`/`description` on blank where Architecture overwrites them**, so it needs a different helper operation. Copying Architecture's here would have introduced a known bug into a type that never had it.
3. **The integration suite surfaced a latent defect from #2923**, affecting architectures as well as patterns.

Everything else — the header/version split, the migration step, search, seed data, the size-limit test — follows #2923 step for step.

---

Second type in the [#2884](https://github.com/finos/architecture-as-code/issues/2884) versioned-artefact storage redesign.

> **Scope:** after this merges, `architectures` and `patterns` use the header/version shape while `flows`, `adrs`, `interfaces`, `standards` and `timelines` still use the old one. That mixed state is ADR 0001's incremental rollout working as intended. Controls and Decorators are excluded permanently (ADR 0004). **#2884 stays open** — five types to go.

Pattern moves to one **header** document per pattern in `patterns` plus one **version** document per version in a new `patternVersions` collection, with a version 3 → 4 migration on both backends. `PatternStore` is unchanged, so nothing above the store layer moves.

### The migration logic is now shared

Pattern's split steps would have been ~95% identical to Architecture's, so the fan-out is extracted into `MongoVersionSplitMigration` / `NitriteVersionSplitMigration`, parameterised by collection names, id field, array field and label. Pattern's steps are **74 and 43 lines** against the Architecture originals' ~250 and ~160, and the five remaining types cost about the same each.

This edits the Architecture steps, which the immutable-shipped-step convention normally forbids. Done on the maintainer's explicit call, given they shipped hours earlier, and on the condition that the outcome is identical. **The evidence is that all 20 Architecture step tests pass against completely unmodified test files** — verified by checking `git diff` on the test directory was empty before running them. The one test added afterwards covers the delegating `transitionIndexes()` method that the fan-out path no longer reaches; it asserts nothing new.

The motivation was concrete rather than aesthetic: three separate reviews had already found one defect surviving in one half of a Mongo/Nitrite pair after being fixed in the other. Seven copies of a 250-line migration would have kept generating that.

### Three behaviours preserved, one fixed

Pattern **guards** `name`/`description` on blank where Architecture overwrites them unconditionally — which is why a version write carrying no name wipes an *architecture's* display name (a known bug) but not a pattern's. Porting Pattern onto the existing helper operation would have introduced that bug into a type that never had it, so there is now a second operation, `updatePresentHeaderDetails`, with tests on both backends pinning the difference. Pattern also keeps its `(CreatePatternRequest, namespace)` signature, and Nitrite still validates JSON before checking existence where Mongo reports the missing pattern first.

The one deliberate change: `NitritePatternStore` returned version `"1-0-0"` from create while Mongo returned `"1.0.0"`, so the `Location` header differed by backend for the same call. Both now return `"1.0.0"`, which is what the new shape actually stores.

### `MongoResourceSlice` is deleted

It existed to page into an array field on a shared namespace document and only ever served Architecture and Pattern, so this port left it with no callers — exactly as ADR 0003 predicted it would "the moment Pattern migrates". No test covered it.

### A latent defect from #2923, found here

Integration tests prime collections with an empty `{namespace, <type>: []}` document — necessary under the old shape, because there had to be something to `$push` into. Under the header shape that document *is* a resource, with no id field, so it lists as `"Pattern null"` with zero versions. All ten `MongoPatternIntegration` tests failed on it.

**Eight files did this, three of them for architectures** — shipped undetected in #2923 because no Architecture integration test asserts an empty listing, while `MongoPatternIntegration` does. All eight are fixed here, and `patternVersions` added to the reset so version documents no longer leak between tests.

Also carried across: the streaming fix jpgough-ms raised on #2923 as non-blocking, which had been applied to the Mongo split step but not its Nitrite twin.

### Per-type checklist

The plan's checklist is now six items, all verified for Pattern: search path moved (silent failure otherwise), `init-mongo.js` seed + index list + `LATEST_SCHEMA_VERSION` → 4, `EndToEndResource` index transition, the 413 size-limit test relocated to Flow, integration-test priming stripped, and the null-entries-array search test moved to Flow. Both relocating tests now document their migration chain.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

**Unit:** 2487 tests, 0 failures, JaCoCo 90%-per-class gate met.

**Integration** (`../mvnw -P integration verify`, TestContainers against real MongoDB): 504 tests, 0 failures.

**Seed script** run against a real MongoDB: 4 pattern headers, 4 versions, no old-shape documents, `versionCount` as int32, dot-separated versions, and the index inventory correct on both sides of the rollout — Pattern on `(namespace, patternId)`, the five unmigrated types still on `(namespace)`.

**Coverage caught a real gap:** `MongoSearchStore` dropped to 82% when patterns moved off the array path, because that path lost its only test with an actual match. Rather than add a coverage-only test, `return_results_from_multiple_collections` now searches an architecture, a pattern **and** a flow together — proving both shapes work in one search while the rollout is part-done.

## Checklist


calm-hub/src/integration-test/java/integration/UserAccessGrantsIntegration.java
```diff
@@ -61,9 +61,9 @@ void setup() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("patterns")) {
                 database.createCollection("patterns");
-                database.getCollection("patterns").insertOne(
-                        new Document("namespace", "finos").append("patterns", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/PermittedScopesIntegration.java
```diff
@@ -49,9 +49,9 @@ void setup() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("patterns")) {
                 database.createCollection("patterns");
-                database.getCollection("patterns").insertOne(
-                        new Document("namespace", "finos").append("patterns", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoMappingControllerIntegration.java
```diff
@@ -40,9 +40,9 @@ public void setup() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("patterns")) {
                 database.createCollection("patterns");
-                database.getCollection("patterns").insertOne(
-                        new org.bson.Document("namespace", "finos").append("patterns", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoMcpIntegration.java
```diff
@@ -158,9 +158,9 @@ public void setup() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("architectures")) {
                 database.createCollection("architectures");
-                database.getCollection("architectures").insertOne(
-                        new Document("namespace", "finos").append("architectures", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoAuditLogIntegration.java
```diff
@@ -62,8 +62,9 @@ public void setup() {
 
         if (!database.listCollectionNames().into(new ArrayList<>()).contains("architectures")) {
             database.createCollection("architectures");
-            database.getCollection("architectures").insertOne(
-                    new Document("namespace", "finos").append("architectures", new ArrayList<>()));
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoAuditLogDeniedIntegration.java
```diff
@@ -63,8 +63,9 @@ void setup() {
 
         if (!database.listCollectionNames().into(new ArrayList<>()).contains("patterns")) {
             database.createCollection("patterns");
-            database.getCollection("patterns").insertOne(
-                    new Document("namespace", "finos").append("patterns", new ArrayList<>()));
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoArchitectureIntegration.java
```diff
@@ -42,9 +42,9 @@ public void setupArchitectures() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("architectures")) {
                 database.createCollection("architectures");
-                database.getCollection("architectures").insertOne(
-                        new Document("namespace", "finos").append("architectures", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/ProxyAuthIntegration.java
```diff
@@ -89,9 +89,9 @@ void setup() {
 
             if (!database.listCollectionNames().into(new ArrayList<>()).contains("patterns")) {
                 database.createCollection("patterns");
-                database.getCollection("patterns").insertOne(
-                        new Document("namespace", "finos").append("patterns", new ArrayList<>())
-                );
+// Collection only — no per-namespace document. That priming belonged to the old
```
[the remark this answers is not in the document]
markscott-ms: Fixed in 64e705c.

Correct on all eight, and they share one cause: the old-shape priming was removed with a scripted edit across these files, and the replacement comment's template indented its continuation lines but not the first — so every occurrence landed flush at column zero.

calm-hub/src/integration-test/java/integration/MongoPatternIntegration.java
```diff
@@ -34,8 +34,11 @@ public static void resetPatterns() {
         try (MongoClient mongoClient = MongoClients.create(mongoUri)) {
             MongoDatabase database = mongoClient.getDatabase(mongoDatabase);
 
-            // Drop patterns and resource_mappings to ensure clean state regardless of test ordering
+            // Drop patterns and resource_mappings to ensure clean state regardless of test ordering.
+            // patternVersions goes too — version documents now live outside the header
+            // collection, so dropping only that would leave them behind for the next test.
             database.getCollection("patterns").drop();
+            database.getCollection("patternVersions").drop();
```
jpgough-ms, line 42: Dropping these collections removes their unique indexes with them, and nothing recreates them for the rest of the JVM.

`EndToEndResource.start()` runs `new MongoPatternVersionSplitStep(database).transitionIndexes()` exactly once per container (line 50), creating unique `patterns.(namespace, patternId)` and `patternVersions.(namespace, patternId, version)`. A Mongo `drop()` takes the indexes too. The only other `createIndex`/`transitionIndexes` call in the integration suite is `MongoArchitectureMigrationIntegration`, which covers architectures only.

`MongoVersionDocumentStore.createVersion` now reports duplicates by catching DUPLICATE_KEY from `insertOne`. With no index it never fires, so `createVersion` returns `true` for a version that already exists: POST of an existing pattern version yields 201 and two version documents instead of 409. Every later class that touches patterns — `MongoMappingControllerIntegration`, `PermittedScopesIntegration`, `ProxyAuthIntegration`, `UserAccessGrantsIntegration`, `MongoAuditLogDeniedIntegration` — then runs with no uniqueness constraint, so a regression in duplicate-version rejection passes CI green.

This is newly load-bearing in this PR: the old `MongoPatternStore` detected duplicates with an `$elemMatch` / `$exists:false` conditional update, which needed no index.

Consider recreating the indexes after the drop, mirroring what `MongoArchitectureMigrationIntegration` does for architectures.

calm-hub/src/integration-test/java/integration/performance/MongoDocumentSizeLimitIntegration.java
```diff
@@ -103,22 +109,23 @@ private static Response putVersion(String path, int id, int major, String body)
 
     @Test
     void return_413_when_a_version_write_exceeds_the_document_size_limit() throws Exception {
-        int patternId = createResource("patterns", "patternJson", "size-limit-test-pattern");
-        String requestBody = largeBody("patternJson", "size-limit-test-pattern");
+        int flowId = createResource("flows", "flowJson", "size-limit-test-flow");
```
jpgough-ms, line 138: Repointing this test to `flows` leaves the finos `flows` document at the 16MB ceiling for every test that runs after it — nothing drops `flows`.

`patterns` was safe here because `MongoPatternIntegration` explicitly drops it "to ensure clean state regardless of test ordering". `flows` has no equivalent: `MongoFlowIntegration.setupFlows()` only creates the collection if absent, and `setup()` in this class does no drops.

Anything touching finos flows afterwards breaks:

- `MongoFlowIntegration.end_to_end_get_with_no_flow` asserts `body("values", empty())` and sees the leftover `size-limit-test-flow`
- its subsequent version PUTs get 413 instead of 201
- `MongoConcurrencyIntegration.concurrent_flow_creation_produces_unique_ids_and_no_data_loss` fails, as every concurrent POST hits the ceiling

This passes today only because failsafe's default order walks `integration/` before `integration/performance/`. Reordering, parallelisation, or a future flow test placed under `performance/` turns it into a red build with a 413 that has nothing to do with document size.

Consider dropping `flows` in an `@AfterAll`, or running this test against a dedicated namespace.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
```diff
@@ -5,374 +5,142 @@
...
-
-            // Nitrite has no array-slice projection, so apply the limit/offset window in memory.
-            List<NamespaceResourceSummary> pageResults = page.apply(patternSummaries);
-            LOG.debug("Retrieved {} of {} patterns for namespace '{}'", pageResults.size(), patternSummaries.size(), namespace);
-            return pageResults;
-        } finally {
-            lock.readLock().unlock();
-        }
+        requireNamespace(namespace);
+        return documentStore.listSummariesPaged(namespace, page);
```
jpgough-ms, line 72: `listSummariesPaged` sorts with `Integer.compare(left.getId(), right.getId())` on a nullable id, so a header with no `patternId` NPEs the entire namespace listing.

The inline loop this replaces had an `if (patternId != null)` guard. `NitriteVersionSplitMigration.writeOneResource` does `put(idField, entry.get(idField, Integer.class))`, so an old-shape entry without `patternId` — or a leftover old-shape document the migration did not fan out — produces a header with a null id, and `toSummary` carries it through.

A single such header survives: `List.sort` skips the comparator and the endpoint returns a `"Pattern null"` row. Two or more in one namespace and the unboxing throws, so `GET /api/calm/namespaces/{ns}/patterns` returns 500 and hides every valid pattern in that namespace.

`MongoPatternStore` sorts DB-side and still returns 200 on the same data, so the two backends now disagree. Consider `Comparator.nullsLast(...)` or restoring the null guard.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
```diff
@@ -5,374 +5,142 @@
...
-                throw new PatternVersionNotFoundException();
-            }
+        requirePattern(pattern);
 
-            LOG.debug("Retrieved version '{}' for pattern {} in namespace '{}'",
-                    pattern.getMongoVersion(), pattern.getId(), pattern.getNamespace());
-            return (String) versionObj;
-        } finally {
-            lock.readLock().unlock();
+        String content = documentStore.getVersion(
```
jpgough-ms, line 106: The `instanceof String` guard on stored version content is gone, so non-String `content` now returns 500 instead of 404.

The old `getPatternForVersion` did `Object versionObj = versions.get(...); if (!(versionObj instanceof String)) throw new PatternVersionNotFoundException();`. `testGetPatternForVersion_whenVersionObjIsNotAString_throwsPatternVersionNotFoundException` covered exactly this and is deleted in this PR.

The new path calls `documentStore.getVersion(...)`, which does `versionDocument.get(CONTENT_FIELD, String.class)` — Nitrite's typed getter casts rather than returning null. A `patternVersions` document whose `content` is not a String throws out of the store, so `GET /namespaces/{ns}/patterns/{id}/versions/{v}` 500s where the resource layer is written to produce a 404.

Reachable from a hand-repaired standalone database, or from `NitriteVersionSplitMigration.writeOneResource`, which likewise does `storedVersions.get(key, String.class)` on pre-migration data of an unexpected type.

jpgough-ms (commented):
Automated review of the Pattern versioned-storage redesign. The migration to the header/version shape looks structurally sound; the findings cluster in the integration-test setup, which undermines invariants the new design relies on the database to enforce.

The two worth acting on first are inline: `MongoPatternIntegration` dropping the unique indexes that duplicate-version rejection now depends on, and the 16MB test repointed onto `flows` with no cleanup.

Three further notes land outside the diff hunks, so they are here rather than inline.

---

**Off-diff note — `calm-hub/mongo/init-mongo.js:141`: `isEmptyDatabase` was not extended for patterns.**

The guard gained `db.architectures.countDocuments() === 0` when Architecture migrated. Pattern is now header/version shaped too, but no equivalent `db.patterns.countDocuments() === 0` clause was added, so it tests only: no `schemaVersion` document, zero `namespaces`, zero `architectures`.

Against a database holding old-shape pattern documents but none of those three — a partially hand-restored dev database, or a dump restored collection-by-collection where `patterns` landed first — `isEmptyDatabase` is true. Line 145 then pins `schemaVersion` to `LATEST_SCHEMA_VERSION = 4` and lines 164-165 create the new indexes. `SchemaMigrationRunner.runWithLock` short-circuits, so `MongoPatternVersionSplitStep` never runs, and the Patterns block at line 586 also skips. Those patterns are then permanently unreadable through the API: they surface in listings as `"Pattern null"` with `versionCount: 0`, and `GET .../patterns/{id}` 404s for every real one.

Reachability rests on hand-assembled databases rather than any automated path, so treat this as lower priority than the inline items.

---

**Off-diff note — `calm-hub/src/integration-test/java/integration/performance/MongoDocumentSizeLimitIntegration.java:135`: Pattern lost all document-size coverage.**

The class previously held one assertion per shape, with Pattern owning the 413 half. That half is now Flow, and the surviving positive test `keep_accepting_architecture_versions_well_past_the_old_document_ceiling` exercises `architectures` only.

The rewritten javadoc still claims "Architecture and Pattern have moved to the header/version shape ... The same history that breaks Flow must now be writable, which is the whole point of the redesign." No Pattern counterpart asserts it. Since Pattern now writes via `MongoVersionDocumentStore.createVersion`/`upsertVersion` rather than the old array-push, a defect reintroducing accumulation into one document — or a mis-specified `patternVersions` index — would let a deployment start returning 413 once a pattern's history passed 16MB, with nothing in the suite failing.

---

**Off-diff note — `calm-hub/src/main/java/org/finos/calm/store/mongo/MongoSearchStore.java:76`: stale javadoc now contradicts the code above it.**

Line 61 changes from `searchNamespacedCollection(patternCollection, "patterns", ...)` to `searchHeaderCollection(patternCollection, "patternId", ...)`, but the javadoc below still reads "Architecture reads this way while patterns, flows, standards and interfaces still read the other."

The next sentence documents the silent failure mode — a migrated type left on the old method returns zero results with nothing logged — so this is precisely the note the next migration author will consult to pick a method, and it now names a type that has already migrated. `NitriteSearchStore` has the same code change without the comment, so the two files disagree.

---

Integration tests were not executed for this review; findings are from static analysis of the diff. Given that two of them concern integration-test isolation, `./mvnw -P integration verify` would be worth running before acting on them.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitritePatternStoreShould.java
```diff
@@ -4,935 +4,399 @@
...
-        when(cursor.firstOrNull()).thenReturn(namespaceDoc);
-        when(mockCollection.find(any(Filter.class))).thenReturn(cursor);
-
-        // Act
-        String result = patternStore.getPatternForVersion(pattern);
-
-        // Assert
-        assertThat(result, is(PATTERN_JSON));
-    }
+    // --- getPatternVersions ---
```
jpgough-ms, line 248: The three per-operation namespace-not-found tests that the Architecture reference keeps are absent from both Pattern test classes.

`TestMongoArchitectureStoreShould` and `TestNitriteArchitectureStoreShould` each cover creating a version, getting a version and updating a version in a missing namespace. Both Pattern classes keep only the listing case.

`calm-hub/AGENTS.md` names the Architecture and Pattern store tests as reference implementations for each other. The Nitrite Pattern store also reaches `requireNamespace` on a path Architecture's does not — validation runs between the namespace check and the existence check — so a future reordering that let a version write proceed against a deleted namespace would fail for Architecture and pass silently for Pattern.

calm-hub/src/test/java/org/finos/calm/store/mongo/TestMongoPatternStoreShould.java
```diff
@@ -54,572 +64,362 @@ public class TestMongoPatternStoreShould {
...
+            throw writeError(10334, "object to insert too large");
+        }).when(versionCollection).insertOne(any(Document.class));
 
-        Pattern pattern = new Pattern.PatternBuilder().setNamespace("finos").setId(42).build();
-        List<String> patternVersions = mongoPatternStore.getPatternVersions(pattern);
+        assertThrows(StorageWriteException.class,
+                () -> store.createPatternForNamespace(createRequest(), NAMESPACE));
 
-        assertThat(patternVersions, is(List.of("1.0.0")));
+        verify(headerCollection).deleteOne(any(Bson.class));
```
jpgough-ms, line 243: The `!created` compensation branch has no test here — this file is the only one of the four that omits it.

`MongoPatternStore.createInitialVersion` (MongoPatternStore.java:150-155) has two failure paths: `catch (RuntimeException)` → `deleteHeader` → rethrow, covered by the test above; and `!created` → `deleteHeader` → `StorageWriteException.writeFailed(...)`, untested. `TestMongoArchitectureStoreShould:233` and `TestNitritePatternStoreShould:236` both carry the matching `fail_rather_than_report_success_when_the_initial_version_already_exists`.

Without it, a duplicate-key insert on a freshly allocated patternId can regress to returning 201 with content that was never stored. JaCoCo's 90% per-class line threshold will not catch a single missing branch.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
```diff
@@ -1,278 +1,183 @@
...
-        // so these guards are only reachable by non-REST callers (e.g. direct store usage in tests).
-        if (pattern.getName() != null && !pattern.getName().isBlank()) {
-            setFields.append("patterns.$.name", pattern.getName());
+    /**
+     * Writes the first version of a newly created pattern, removing the header again if that
+     * fails. See {@code MongoArchitectureStore.createInitialVersion} — a header with no
+     * versions cannot be removed through the API, so a failed first version write must not
+     * leave one behind.
+     */
+    private void createInitialVersion(String namespace, int id, Document content) {
```
jpgough-ms: `createInitialVersion` is now duplicated four ways — consider moving it onto the version document stores.

`MongoPatternStore:144-158` is identical to `MongoArchitectureStore:158-172` apart from `ID_FIELD`, and `NitritePatternStore:173-187` is identical to `NitriteArchitectureStore:182-196`. This PR adds copies three and four.

The routine exists to avoid stranding an unremovable header. There is no delete endpoint for these types, so a fix applied to one copy and missed in another leaves headers with `versionCount: 0` visible in listings and search permanently.

It depends on nothing type-specific (`createVersion`, `deleteHeader`, `INITIAL_VERSION`), so `createFirstVersion(namespace, resourceId, content)` on `Mongo`/`NitriteVersionDocumentStore` would own both calls it already makes.

calm-hub/src/integration-test/java/integration/MongoPatternIntegration.java
markscott-ms: Confirmed and fixed in f9d4cfc0 — `transitionIndexes()` now runs after the drop.

Your framing of *why* this only bites now is the part I'd missed: the drop predates this PR and was harmless while duplicates were rejected by the `$elemMatch`/`$exists` conditional update, which needed no index. Moving detection onto `DUPLICATE_KEY` made an existing line of test setup load-bearing without changing it.

I checked whether #2923 shipped the same hole for Architecture — `MongoArchitectureIntegration` has no `@BeforeAll` drop at all, so this is Pattern-only.

calm-hub/src/integration-test/java/integration/performance/MongoDocumentSizeLimitIntegration.java
markscott-ms: Confirmed and fixed in 3526cd4f. Both tests now run against a dedicated `size-limit` namespace rather than an `@AfterAll` drop, since dropping `flows` would take the step-0 index with it and trade one ordering dependency for another.

Worth noting the header/version half had the milder version of the same problem — it leaves ~24MB of version documents in `finos` — so isolating both was the honest fix rather than just the one that breaks things. All namespace-listing assertions use `hasItem`, so the extra namespace is inert.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
markscott-ms: Confirmed and fixed in a432d16a, in `NitriteVersionDocumentStore` rather than the Pattern store so every type gets it.

`Comparator.nullsLast` on the id. The backends disagreeing was the deciding argument — `toSummary` already renders the row, so failing the other resources beside it was never the right answer.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitritePatternStore.java
markscott-ms: Confirmed and fixed in 8d867a1d — guard restored in `NitriteVersionDocumentStore.getVersion`, with a test standing in for the one that was deleted.

Your last line sent me to the migrations, and that half is worse than the read path: a `ClassCastException` in `writeOneResource` aborts the run with the schema lock still held, so one malformed document leaves the whole hub refusing requests. `MongoVersionSplitMigration` has the identical exposure — `org.bson.Document.get(key, Class)` also casts — so both are fixed.

Both now carry the value across unchanged and warn, rather than dropping or coercing it: a migration is the wrong place to discard data, and with the read path fixed the resource stays serviceable until someone repairs the document.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitritePatternStoreShould.java
markscott-ms: Confirmed and fixed in 72a3a2f7 — all four added to the Nitrite class and three to the Mongo one.

The ordering point is the one that makes these worth having rather than symmetric box-ticking: with validation between the namespace check and the existence check, Pattern really would have passed a reordering that Architecture caught.

calm-hub/src/test/java/org/finos/calm/store/mongo/TestMongoPatternStoreShould.java
markscott-ms: Confirmed and fixed in 72a3a2f7.

The branch has since moved into `MongoVersionDocumentStore.createFirstVersion` (see the reply on `MongoPatternStore:144`), but the store test still earns its place — the store builds a real helper over mocked collections, so it pins that Pattern routes creation through the compensating path at all.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoPatternStore.java
markscott-ms: Agreed and done in 7762fd98 — `createFirstVersion(namespace, resourceId, content)` on both helpers, with `INITIAL_VERSION` moved across too.

Four copies is the floor: at full rollout it would have been fourteen. #2935 currently carries ten more that this lets me delete on rebase. ADR 0003 records the operation and the reasoning in 9f49474f.

jpgough-ms (approved):

Issue #2928, opened 2026-08-03
[PoC] Live CALM diagrams in the learn tutorials


gjs-opsflo:
## What this PoC explores

The learn tutorials teach people to build CALM architectures step by step, but the learner never sees what they built — they paste JSON and read "validation succeeded with 4 nodes". Now that live diagrams work in the docs site (#2854, #2900, #2909), this PoC explores embedding them in the learn track: after each step, the learner sees the architecture they just wrote, rendered from the same kind of `.calm.json` file they are editing.

## The experiment

Draft PR #2921 adds live diagrams to **two pages only**, as a taste:

- **Beginner → Build a Complete Architecture**: the finished e-commerce architecture renders at the end of the page, with the platform boundary drawn as a nested container.
- **Intermediate → Model Business Flows**: the order-processing flow the learner modelled animates over the architecture, non-flow elements dimmed.

Both diagrams are generated at build time from committed `.calm.json` files, so a broken tutorial architecture fails the site build — tutorials cannot silently rot.

## How we decide

This follows the run-an-experiment-then-decide style: the draft PR is the experiment, this issue is where we decide.

- **Accepted** → extend to the full learn track, using a single-source pattern (one `.calm.json` per step feeding both the code block via `raw-loader` and the diagram, so fenced JSON and the picture can never drift), and fix the small renderer follow-up found on the way (flow overlay reads `summary`, the flow schema defines `description` — renderer should accept both).
- **Not the right fit** → close this and the draft, no harm done. Especially open to reshaping around the tutorials/learning rework @karlmoll has in progress.

Happy to demo at office hours.

Related: #2921 (the draft PR), #2853 / #2854 / #2900 / #2909 (the docs-embed work this builds on).

jimthompson5802:
@gjs-opsflo   The live diagrams are visually compelling.  I have two questions.

## Beginner: Build Complete Architecture.  

### Will the live diagram replace the ascii version?

<img width="1594" height="568" alt="Image" src="https://github.com/user-attachments/assets/b828ec75-5851-40b6-848e-15a55389d323" />

<img width="1614" height="753" alt="Image" src="https://github.com/user-attachments/assets/f7e6c6dc-3159-4f13-8679-e65fd125eeee" />


## Intermediate: Model Business Flow

### From what I can tell the live diagram depicts `order-processing` and it is showing a bi-directional flow between `order-service` and `payment-service`

[...]

PR #2939, opened 2026-08-05
fix(calm-suite): read flow transition labels from the schema description field

gjs-opsflo:
## Description

The flow overlay in `@calmstudio/diagram` read transition labels from a legacy `summary` field, but the flow meta-schema (`calm/release/1.2/meta/flow.json`) defines `description` on transitions. A schema-correct flow document with no `summary` rendered its sequence-badge tooltips as `undefined`.

Fix: prefer the schema's `description`, keep `summary` as a fallback for older files. Found while building the learn-tutorials PoC (#2928) — its demo files carried both fields as a workaround, which this makes unnecessary.

Side effect: removes two pre-existing typecheck errors in `flowOverlay.ts` (the typed access to the non-schema `summary` field).

## Type of Change

## Affected Components

- [x] CALM Studio (`calm-studio/`) — `packages/web-component`

## Testing

- [x] All existing tests pass

Evidence: new test renders a schema-only flow (no `summary`) and asserts the badge carries the `description` text with no `undefined` in the output; suite 54/54; typecheck error count drops by two.

## Checklist

### DIFF
--- a/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
+++ b/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
@@ -226,6 +226,48 @@ describe('renderFlowOverlay', () => {
     expect(badgeCount).toBe(1);
   });
 
+  it('spreads badges when multiple transitions share a relationship, and all remain visible', () => {
+    const sharedFlow: CalmFlow = {
+      'unique-id': 'shared-flow',
+      name: 'Shared',
+      description: 'Two transitions over one relationship',
+      transitions: [
+        {
+          'relationship-unique-id': 'rel-1',
+          'sequence-number': 1,
+          description: 'Request',
+          direction: 'source-to-destination',
+        },
+        {
+          'relationship-unique-id': 'rel-1',
+          'sequence-number': 2,
+          description: 'Response',
+          direction: 'destination-to-source',
+        },
+      ],
+    };
+    const svg = renderFlowOverlay(sharedFlow, twoEdgeLayouts);
+    const badges = (svg.match(/class="flow-badge/g) ?? []).length;
+    expect(badges).toBe(2);
+    expect(svg).toContain('>1<');
+    expect(svg).toContain('>2<');
+    // Badge centers must not coincide: collect cx values of badge circles
+    const cxs = [...svg.matchAll(/<circle cx="([\d.-]+)" cy="([\d.-]+)" r="10"/g)].map((m) => m[1] + ',' + m[2]);
+    expect(new Set(cxs).size).toBe(cxs.length);
+  });
+
+  it('renders destination-to-source badges hollow (response convention)', () => {
+    const svg = renderFlowOverlay(reverseFlow, twoEdgeLayouts);
+    // reverseFlow's single transition is destination-to-source
+    expect(svg).toContain('class="flow-badge flow-badge-reverse"');
+    expect(svg).toContain('fill="#ffffff"');
+  });
+
+  it('forward badges stay solid', () => {
+    const svg = renderFlowOverlay(sampleFlow, twoEdgeLayouts);
+    expect(svg).not.toContain('flow-badge-reverse');
+  });
+
   it('Test 7: handles flow with 3+ transitions (multi-edge flow)', () => {
     const svg = renderFlowOverlay(multiEdgeFlow, twoEdgeLayouts);
     // Three transitions, rel-3 exists in layout
--- a/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
+++ b/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
@@ -156,6 +156,25 @@ describe('renderFlowOverlay', () => {
     expect(svg).toContain('keyPoints="0;1"');
   });
 
+  it('renders badges from the schema description field when summary is absent', () => {
+    const schemaFlow: CalmFlow = {
+      'unique-id': 'schema-flow',
+      name: 'Schema Flow',
+      description: 'Flow using only schema-defined fields',
+      transitions: [
+        {
+          'relationship-unique-id': 'rel-1',
+          'sequence-number': 1,
+          description: 'Client sends order request',
+          direction: 'source-to-destination',
+        },
+      ],
+    };
+    const svg = renderFlowOverlay(schemaFlow, twoEdgeLayouts);
+    expect(svg).toContain('Client sends order request');
+    expect(svg).not.toContain('undefined');
+  });
+
   it('Test 4: renders sequence badge with correct number for each transition', () => {
     const svg = renderFlowOverlay(sampleFlow, twoEdgeLayouts);
     // Sequence numbers 1 and 2 should appear in badge text elements
--- a/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
+++ b/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
@@ -256,6 +256,27 @@ describe('renderFlowOverlay', () => {
     expect(new Set(cxs).size).toBe(cxs.length);
   });
 
+  it('separates request/response badges even on a short edge (return-lane offset)', () => {
+    const shortLayouts = new Map([
+      ['rel-s', [{ id: 'rel-s', points: [ { x: 100, y: 100 }, { x: 100, y: 140 } ] }]],
+    ]);
+    const rrFlow: CalmFlow = {
+      'unique-id': 'rr',
+      name: 'RR',
+      description: 'request-response on a 40px edge',
+      transitions: [
+        { 'relationship-unique-id': 'rel-s', 'sequence-number': 1, description: 'req', direction: 'source-to-destination' },
+        { 'relationship-unique-id': 'rel-s', 'sequence-number': 2, description: 'res', direction: 'destination-to-source' },
+      ],
+    };
+    const svg = renderFlowOverlay(rrFlow, shortLayouts);
+    const centers = [...svg.matchAll(/<circle cx="([\d.-]+)" cy="([\d.-]+)" r="10"/g)].map((m) => [Number(m[1]), Number(m[2])]);
+    expect(centers.length).toBe(2);
+    const [a, b] = centers as [[number, number], [number, number]];
+    const dist = Math.hypot(a[0] - b[0], a[1] - b[1]);
+    expect(dist).toBeGreaterThanOrEqual(18);
+  });
+
   it('renders destination-to-source badges hollow (response convention)', () => {
     const svg = renderFlowOverlay(reverseFlow, twoEdgeLayouts);
     // reverseFlow's single transition is destination-to-source
--- a/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
+++ b/calm-studio/packages/web-component/src/render/flowOverlay.test.ts
@@ -289,6 +289,91 @@ describe('renderFlowOverlay', () => {
     expect(svg).not.toContain('flow-badge-reverse');
   });
 
+  it('labels badges from legacy summary-only transitions (backward compatibility)', () => {
+    const legacyFlow = {
+      'unique-id': 'legacy-flow',
+      name: 'Legacy',
+      description: 'summary-only transition',
+      transitions: [
+        {
+          'relationship-unique-id': 'rel-1',
+          'sequence-number': 1,
+          summary: 'Legacy label text',
+          direction: 'source-to-destination',
+        },
+      ],
+    } as unknown as CalmFlow;
+    const svg = renderFlowOverlay(legacyFlow, twoEdgeLayouts);
+    expect(svg).toContain('Legacy label text');
+    expect(svg).not.toContain('undefined');
+  });
+
+  it('prefers description over summary when both are present', () => {
+    const bothFlow = {
+      'unique-id': 'both-flow',
+      name: 'Both',
+      description: 'both fields',
+      transitions: [
+        {
+          'relationship-unique-id': 'rel-1',
+          'sequence-number': 1,
+          description: 'Schema label',
+          summary: 'Legacy label',
+          direction: 'source-to-destination',
+        },
+      ],
+    } as unknown as CalmFlow;
+    const svg = renderFlowOverlay(bothFlow, twoEdgeLayouts);
+    expect(svg).toContain('Schema label');
+    expect(svg).not.toContain('Legacy label');
+  });
+
+  it('orders badge spread by sequence-number, not document order', () => {
+    const outOfOrder: CalmFlow = {
+      'unique-id': 'ooo',
+      name: 'OOO',
+      description: 'transitions listed out of sequence',
+      transitions: [
+        { 'relationship-unique-id': 'rel-1', 'sequence-number': 2, description: 'second', direction: 'source-to-destination' },
+        { 'relationship-unique-id': 'rel-1', 'sequence-number': 1, description: 'first', direction: 'source-to-destination' },
+      ],
+    };
+    const svg = renderFlowOverlay(outOfOrder, twoEdgeLayouts);
+    // rel-1's path starts at (10,20) and heads toward (200,80): earlier along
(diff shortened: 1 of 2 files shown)


calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -70,18 +79,41 @@ export function renderFlowOverlay(
...
-    const midX = midPoint.x;
-    const midY = midPoint.y;
+    // One sequence badge per transition. Transitions sharing a relationship
+    // spread along the first edge's path (k-th of n sits at (k+1)/(n+1)) so
+    // request/response pairs never stack invisibly on the midpoint.
+    const relId = transition['relationship-unique-id'];
+    const seen = perRelationshipSeen.get(relId) ?? 0;
+    perRelationshipSeen.set(relId, seen + 1);
+    const count = perRelationshipCount.get(relId) ?? 1;
+    const fraction = (seen + 1) / (count + 1);
```
markscott-ms: The spread index comes from document order rather than `sequence-number`, so a flow that lists its transitions out of sequence gets badges descending along the path. Sorting the per-relationship index by `sequence-number` is a two-line change.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -70,18 +79,41 @@ export function renderFlowOverlay(
...
+    const badgePoint = pointAtFraction(badgeEdge.points, fraction);
+    if (badgePoint === undefined) continue;
+    // Reverse (response) badges shift perpendicular to the path — a "return
+    // lane" beside the edge — so request/response pairs stay separated even
+    // on edges shorter than a badge diameter.
+    const isReverse = direction === 'destination-to-source';
+    const normal = pathNormalAt(badgeEdge.points, fraction);
+    const laneOffset = isReverse && normal !== undefined ? 22 : 0;
```
markscott-ms, line 113: This applies to **every** reverse badge, not just crowded ones. A flow with a single `destination-to-source` transition on its own relationship previously had its badge on the line; it now floats 22px to one side with nothing tying it visually to that edge. On a shared edge you also get both corrections at once — spread along the path *and* the perpendicular shift.

Suggest gating on the crowded case, e.g. only when `count > 1`, or when the along-path separation falls below a badge diameter (`total * (1 / (count + 1)) < 20`).

markscott-ms (commented):
## Overview

`flowOverlay.ts` read `transition.summary` — a legacy CalmStudio field absent from `CalmFlowTransitionSchema` — so schema-valid flows rendered `data-summary="undefined"` and `<title>undefined</title>`. Preferring `transition.description` with a cast-based `summary` fallback is correct: `calm-models/src/types/flow-types.ts:6` confirms `description: string` is the schema field and `summary` doesn't exist, which is also the source of the two typecheck errors this removes.

The diff also carries badge-placement and styling work: transitions sharing a relationship now spread along the path at `(k+1)/(n+1)` rather than stacking at the midpoint, reverse transitions shift into a perpendicular "return lane" and render hollow via a new `flow-badge-reverse` class, and two geometry helpers (`pointAtFraction`, `pathNormalAt`) support that. It's coherent with the bug being fixed — the same request/response demo files that exposed the `undefined` labels are exactly where stacked badges bite. Worth a line in the PR body so the visual change is discoverable from the changelog, but no objection to it riding along.

## Correctness

Inline comments cover the specifics. In summary:

- The lane offset applies to every reverse badge, not just crowded ones — a lone `destination-to-source` transition's badge now floats 22px off its edge.
- The perpendicular direction is arbitrary (always `(-ty, tx)`), so which side a badge lands on depends on layout point ordering; nothing prevents it landing over a node.
- Badge spread follows document order rather than `sequence-number`.
- `pointAtFraction` emits raw floats, so output now contains `cx="140.86000000000001"`-style values where layout integers used to be.
- Pre-existing but adjacent: `${transition['sequence-number']}` is interpolated into text content unescaped. It's typed `number`, but these are parsed JSON documents, so a string value would inject markup — cheap to route through `escapeAttr` while you're in this code.

## Code quality

- `pointAtFraction` and `pathNormalAt` duplicate the segment-length walk verbatim, and the copies diverge on degenerate input (`pointAtFraction` returns `first` when `total === 0`; `pathNormalAt` returns `undefined`). One function returning `{ point, normal }` removes both the duplication and the divergence.
- The attribute is still `data-summary` while the value now comes from `description`. I found no reader in `calm-studio/` — only the `:global(.flow-badge)` cursor rule at `CalmDiagram.svelte:234` — so either rename to `data-description` or drop it, since `<title>` already supplies the tooltip.
- Hard-coded `#ffffff` for the hollow badge: the component supports `:host([theme="dark"])` with `--calm-bg: #1e1e1e` (`CalmDiagram.svelte:228`), so on dark the "hollow" reading inverts to a white disc. The file hard-codes `#3b82f6` throughout already, so this is existing debt rather than new — follow-up material.
- A short note on the `(transition as { summary?: string })` cast saying when the fallback can be retired would help future cleanup.

## Test coverage

- The core regression test is well-targeted: schema-only flow, asserts the description text appears and `undefined` doesn't.
- **The `summary` fallback is untested.** Nothing asserts a legacy `summary`-only transition still produces a label — existing fixtures use `summary` but only assert on `animateMotion`, `keyPoints`, and badge counts, never label text. Backward compatibility is the stated reason for keeping the cast, so it should have a test, plus one asserting `description` wins when both are present.
- The geometry tests assert only inequality (centres differ, distance >= 18) and would pass under badly wrong placement. Pinning an expected fraction point on a known polyline would lock the behaviour down.
- Minor: the new test counts `/class="flow-badge/g` (unterminated) while the existing test near line 226 counts `class="flow-badge"` (terminated), which now silently excludes reverse badges — it passes only because that fixture is forward-only.
- `expect(svg).toContain('fill="#ffffff"')` couples to a literal colour; asserting `flow-badge-reverse` is enough.
- Fixtures still put `summary` on `CalmFlow`-typed literals, which remain excess-property type errors. The claimed two-error drop is real for the production file; the test file's errors could be cleared in the same pass.

## Verdict

Approve-with-comments. The label fix is correct and the badge work is a reasonable companion. Worth addressing before merge: gate the lane offset to the crowded case, and add a test for the `summary` fallback. Float precision, helper duplication, and the `data-summary` naming are fine as follow-ups.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -70,18 +79,41 @@ export function renderFlowOverlay(
...
+    const midX = badgePoint.x + (normal?.x ?? 0) * laneOffset;
+    const midY = badgePoint.y + (normal?.y ?? 0) * laneOffset;
```
markscott-ms: `pointAtFraction` returns raw floats, so the rendered SVG now carries values like `cx="140.86000000000001"` where it previously emitted layout integers. Rounding to 2dp before interpolation keeps the output smaller and less brittle for anything downstream that string-matches it.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -70,18 +79,41 @@ export function renderFlowOverlay(
...
+    const badgeClass = isReverse ? 'flow-badge flow-badge-reverse' : 'flow-badge';
+    const circleFill = isReverse ? '#ffffff' : '#3b82f6';
+    const circleExtra = isReverse ? ' stroke="#3b82f6" stroke-width="2"' : '';
+    const numberFill = isReverse ? '#3b82f6' : 'white';
     parts.push(
-      `<g class="flow-badge" data-summary="${escapeAttr(transition.summary)}">`,
-      `  <circle cx="${midX}" cy="${midY}" r="10" fill="#3b82f6"/>`,
-      `  <text x="${midX}" y="${midY}" fill="white" font-size="9" font-weight="bold" text-anchor="middle" dominant-baseline="central">${transition['sequence-number']}</text>`,
-      `  <title>${escapeAttr(transition.summary)}</title>`,
+      `<g class="${badgeClass}" data-summary="${escapeAttr(transitionLabel)}">`,
```
markscott-ms: The attribute is still named `data-summary` while its value now comes from `description`. I couldn't find a reader anywhere in `calm-studio/` — the only consumer of the badge is the `:global(.flow-badge)` cursor rule in `CalmDiagram.svelte:234` — so this could be renamed to `data-description`, or dropped entirely since the sibling `<title>` already provides the tooltip.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -158,6 +190,70 @@ function getReferencedNodeIdsWithFlatFallback(rel: CalmRelationship): string[] {
...
+  let target = Math.min(Math.max(fraction, 0), 1) * total;
+  for (let i = 0; i < lens.length; i++) {
+    const len = lens[i] ?? 0;
+    if (target <= len) {
+      const a = points[i];
+      const b = points[i + 1];
+      if (a === undefined || b === undefined || len === 0) return undefined;
+      const tx = (b.x - a.x) / len;
+      const ty = (b.y - a.y) / len;
+      return { x: -ty, y: tx };
```
markscott-ms, line 254: The normal is always `(-ty, tx)`, so which side of the edge a reverse badge lands on depends purely on the layout engine's point ordering — nothing stops it landing on top of a node, another edge, or off-canvas. A comment acknowledging that is probably enough for now; biasing away from the diagram centroid would be the fuller fix.

Separately, this function re-walks the segment lengths exactly as `pointAtFraction` does, and the two copies handle degenerate input differently (`pointAtFraction` returns `first` when `total === 0`, this returns `undefined`). One function returning `{ point, normal }` would remove the duplication and the divergence.

calm-studio/packages/web-component/src/render/flowOverlay.test.ts
```diff
@@ -156,6 +156,25 @@ describe('renderFlowOverlay', () => {
...
+          'sequence-number': 1,
+          description: 'Client sends order request',
+          direction: 'source-to-destination',
+        },
+      ],
+    };
+    const svg = renderFlowOverlay(schemaFlow, twoEdgeLayouts);
+    expect(svg).toContain('Client sends order request');
+    expect(svg).not.toContain('undefined');
+  });
```
markscott-ms, line 176: Good regression test for the schema path. The legacy branch is still uncovered though: no test asserts that a `summary`-only transition produces a label (the existing fixtures use `summary` but only assert on `animateMotion`, `keyPoints`, and badge counts, never the label text). Since backward compatibility is the reason for keeping the cast, worth adding that case plus one asserting `description` wins when both fields are present.

gjs-opsflo:
Thanks @markscott-ms — all six addressed in `65c211d6`:

- **Sequence ordering**: spread rank now sorts by `sequence-number` within each relationship (test: out-of-order transition arrays place badge 1 before badge 2 along the path).
- **Over-applied lane offset**: gated on the crowded case — `count > 1` AND along-path spacing under a badge diameter. A lone reverse badge stays on its line (test added).
- **Float coordinates**: rounded to 2dp at all emission points (test asserts no 3+-decimal runs).
- **`data-summary`**: renamed `data-description` to match its content; confirmed no readers beyond the class-based cursor rule.
- **Normal side / duplicated walk**: both helpers now share one `walkPolyline` with consistent degenerate handling, and the unbiased-side limitation is acknowledged in a comment, with centroid-bias noted as the fuller fix if it shows up in practice.
- **Label test gaps**: added legacy `summary`-only coverage and a description-wins-when-both-present test.

Suite 64/64, typecheck unchanged.

gjs-opsflo:
@markscott-ms gentle nudge — all six review points were addressed in 65c211d6 (sequence ordering, crowding gate, rounding, data-description, unified path walk, label tests). Ready for another look whenever you have time. No rush, and thanks again for the detailed first pass.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -70,18 +90,46 @@ export function renderFlowOverlay(
...
+    const circleExtra = isReverse ? ' stroke="#3b82f6" stroke-width="2"' : '';
+    const numberFill = isReverse ? '#3b82f6' : 'white';
     parts.push(
-      `<g class="flow-badge" data-summary="${escapeAttr(transition.summary)}">`,
-      `  <circle cx="${midX}" cy="${midY}" r="10" fill="#3b82f6"/>`,
-      `  <text x="${midX}" y="${midY}" fill="white" font-size="9" font-weight="bold" text-anchor="middle" dominant-baseline="central">${transition['sequence-number']}</text>`,
-      `  <title>${escapeAttr(transition.summary)}</title>`,
+      `<g class="${badgeClass}" data-description="${escapeAttr(transitionLabel)}">`,
+      `  <circle cx="${midX}" cy="${midY}" r="10" fill="${circleFill}"${circleExtra}/>`,
+      `  <text x="${midX}" y="${midY}" fill="${numberFill}" font-size="9" font-weight="bold" text-anchor="middle" dominant-baseline="central">${transition['sequence-number']}</text>`,
```
rocketstack-matt, line 131: `transition['sequence-number']` goes into the badge text unescaped, while every other dynamic value here (and in `nodeRenderer.ts`) is routed through an escape helper. The type says `number`, but this is parsed JSON reaching `{@html svgContent}` in `CalmDiagram.svelte:166` — a document with a string in that field injects markup. Raised in the original review body as a while-you're-here item; still open.
```suggestion
      `  <text x="${midX}" y="${midY}" fill="${numberFill}" font-size="9" font-weight="bold" text-anchor="middle" dominant-baseline="central">${escapeAttr(String(transition['sequence-number']))}</text>`,
```

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -158,6 +206,62 @@ function getReferencedNodeIdsWithFlatFallback(rel: CalmRelationship): string[] {
...
+    target -= seg.len;
+  }
+  return undefined;
+}
+
+function round2(n: number): number {
+  return Math.round(n * 100) / 100;
+}
+
 function escapeAttr(str: string): string {
```
rocketstack-matt, line 265: Pre-existing, not introduced here: this duplicates `nodeRenderer.ts`'s `escapeXml` but doesn't escape `'`. Since this file is already being touched, worth exporting `escapeXml` from `nodeRenderer.ts` and importing it here instead of maintaining two slightly different escaping helpers in the same package. Non-blocking.

rocketstack-matt (commented):
The six points from the earlier review are addressed in 65c211d6 (verified against the diff + tests: sequence-number ordering, crowding-gated lane offset, 2dp rounding, `data-description` rename, shared `walkPolyline`, and both label-fallback tests). Three smaller items below.

calm-studio/packages/web-component/src/render/flowOverlay.ts
```diff
@@ -158,6 +206,62 @@ function getReferencedNodeIdsWithFlatFallback(rel: CalmRelationship): string[] {
...
+/** Unit normal of the segment containing the fraction point; undefined for degenerate walks. */
+function normalOnWalk(walk: PolylineWalk, fraction: number): { x: number; y: number } | undefined {
+  if (walk.total === 0) return undefined;
+  let target = Math.min(Math.max(fraction, 0), 1) * walk.total;
+  for (const seg of walk.segments) {
+    if (target <= seg.len) {
+      if (seg.len === 0) return undefined;
+      const tx = (seg.b.x - seg.a.x) / seg.len;
+      const ty = (seg.b.y - seg.a.y) / seg.len;
+      return { x: -ty, y: tx };
```
rocketstack-matt, line 254: Following up on the duplication point from the earlier review: `walkPolyline` now shares the segment/length precompute, but `pointOnWalk` and `normalOnWalk` still each re-walk the segment list to find the target fraction. A single function returning `{ point, normal }` (as originally suggested) would still collapse that. Non-blocking.

rocketstack-matt:
@gjs-opsflo nudge

rocketstack-matt:
Closing due to inactivity, please reopen when ready to proceed.

PR #2942, opened 2026-08-05
Layout Persistence in CALM Hub

jpgough-ms:
## Description
- Adds shared, server-persisted default layouts for architecture diagrams: non-versioned `GET`/`PUT` at `.../namespaces/{namespace}/architectures/{architectureId}/layout`, one flat document per `(namespace, architectureId)` — see ADR 0005 for why this isn't the header/version shape.
- Rebased onto `upstream/main` now that #2935 is merged: `MongoLayoutIndexStep.fromVersion()` moved `4` → `9` (upstream now occupies 2–8) and `init-mongo.js`'s `LATEST_SCHEMA_VERSION` moved to `10`.
- Closes every point from @markscott-ms's review:
  - **Orphan layouts (issue 1)**: `LayoutResource.saveLayout` now 404s if the target architecture doesn't exist (`ArchitectureStore.architectureExists`), so a layout can no longer be saved against an id nothing in the UI can reach. Checked on `PUT` only — a `GET` for an unknown architecture already 404s via the existing `layoutNotFoundResponse`.
  - **Unreachable `DELETE` (issue 2)**: removed entirely (YAGNI) rather than wired up — nothing called it, `LayoutStore.deleteLayout`, both backends' implementations, and `LayoutNotFoundException` are gone. **Consequence worth flagging explicitly**: there is now no API path to remove a saved layout at all, only to overwrite it. Not a new restriction — there's no architecture `DELETE` either — but raising it so it isn't rediscovered.
  - **Desktop save errors invisible (issue 3)**: replaced the dead wrapper `title` and the mobile-only in-menu text with one inline `role="alert"` banner (matching `Drawer`'s existing `dropError` pattern) rendered in the main pane on both desktop and mobile.
  - **Slug-resolution key divergence (issue 4)**: `viewportKeyOverride` is now `string | null | undefined` — `null` means "this is an architecture, but resolution settled with no match," which suppresses `Drawer`'s slug fallback instead of triggering it. Stays `undefined` while still resolving, so the "Loading saved layout…" gate is unaffected.
  - **`onPositionsChange` effect/dependency loop (issue 5)**: `ArchitectureGraph` now holds the callback in a ref and calls through a stable wrapper, so a future caller passing an inline arrow can't loop forever. No `eslint-disable`.
  - **Smaller notes**: `layout-service.ts` failures now carry a user-readable message (403/413/404/generic) instead of the raw log-prefix string; `useDefaultLayout`'s return value is memoised (`DiagramSection`'s save/reset callbacks now depend on the stable `save`/`reset` functions, not the whole object); dropped `layoutEpoch` from `ArchitectureGraph`'s `ReactFlow` `key` — it was forcing a remount on every save/reset that snapped the viewport back to whatever was saved at mount time (a real bug, confirmed and now covered by a regression test) — the parse effect already re-applies positions on its own.
  - **`this.ax = axiosInstance ?? apiClient` (pushback)**: kept the existing `if/else` form. It's the majority convention in this codebase (`control-service`, `counts-service`, `interface-service`, `search-service`) and is the literal template in `calm-hub-ui/AGENTS.md`; `user-access-service`'s `??` is the outlier. Happy to converge the other way as a separate sweep if preferred.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

## Checklist


calm-hub/mongo/init-mongo.js
```diff
@@ -134,7 +134,7 @@ logSection("Schema baseline");
 // Raise LATEST_SCHEMA_VERSION whenever a migration step is added, and seed that step's
 // target shape below. Document shape must match MongoSchemaVersionStore: _id
 // "schemaVersion", int version, in the calm collection.
-const LATEST_SCHEMA_VERSION = 4;
+const LATEST_SCHEMA_VERSION = 5;
```
markscott-ms: this will conflict with #2935. Strongly recommend 2935 is merged first as it a bug fix.

markscott-ms: if this is updated from `5` to `10` it will reduce conflict

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoLayoutIndexStep.java
```diff
@@ -0,0 +1,111 @@
...
+    @ConfigProperty(name = "calm.database.mode", defaultValue = "mongo")
+    String databaseMode;
+
+    public MongoLayoutIndexStep(MongoDatabase database) {
+        this.database = database;
+    }
+
+    @Override
+    public int fromVersion() {
+        return 4;
```
markscott-ms: make this 9 to avoid conflict with #2935 , additional changes will be necessary in the javadoc in this file

markscott-ms (changes requested):

markscott-ms (commented):
## Overview

Adds shared, server-persisted default layouts for architectures:

- **Backend**: `LayoutResource` (GET/PUT/DELETE at `.../architectures/{id}/layout`), `LayoutStore` + Mongo/Nitrite implementations + producer, `LayoutNotFoundException`, `MongoLayoutIndexStep`, audit wiring, and namespace-deletion guarding.
- **Frontend**: `LayoutService`, `useDefaultLayout`, a `CalmLayout` wire model, `canWriteNamespace` on `useCurrentUserAccess`, and Save/Reset controls in `DiagramSection` threaded down through `Drawer` → `ReactFlowVisualizer` → `ArchitectureGraph`.
- **ADR 0005** documents the decision to *not* put the layout shape under `calm/`.

The design is well-reasoned throughout, the precedence rule (local scratch > server default > auto-layout) is right, and the commentary explaining *why* is unusually good. Test coverage is genuinely strong — the slug/numeric viewport-key equivalence test, the post-reflow persistence test, and the `BsonMaximumSizeExceededException` path are all tests that catch real regressions rather than restate the code.

---

## Issues

### 1. Layouts can be saved for architectures that don't exist

`LayoutResource.saveLayout` validates the namespace and the `for` path, but never checks that `architectureId` resolves to a real architecture. `MongoLayoutIntegration` / `NitriteLayoutIntegration` demonstrate this directly: they seed no architecture and `PUT .../architectures/5/layout` returns 204.

Consequence, given `NamespaceContentService.hasContent` now consults `getArchitectureIdsWithLayoutForNamespace`: any authenticated write-scoped user can PUT a layout for an arbitrary id and permanently block deletion of that namespace with a document nothing in the UI can reach or remove. Worth either an existence check in the resource, or excluding layouts from `hasContent` (the comment there already concedes the check is defensive rather than reachable).

### 2. `deleteDefaultLayout` is unreachable from the UI

The DELETE endpoint, the service method, and its tests all exist, but nothing calls it — "Reset to default layout" only clears the localStorage scratch layer. So once a bad shared default is saved, there is no way to remove it, only to overwrite it. Either wire up a "Clear saved default" action, or note in the PR description that DELETE is API-only for now.

### 3. Save errors are invisible on desktop

In `DiagramSection.tsx`, `saveError` is rendered only in the mobile view-options menu. On desktop it is surfaced via `title={defaultLayoutState.saveError ?? undefined}` on the wrapping `<div>` — but both child buttons carry their own `title` ("Save as default layout" / "Reset to default layout"), which wins on hover. In practice a desktop save failure produces no feedback at all: the button un-disables and nothing else happens.

### 4. Failed slug resolution reintroduces the key divergence the override exists to prevent

When `fetchMappings` finds no match (or rejects), `viewportKeyOverride` stays `undefined` and `Drawer` falls back to `${data.name}/${data.id}` — i.e. the **slug**. Drags are then persisted under `ns/my-arch` while the same architecture visited numerically uses `ns/42`. That is precisely the split `viewportKeyOverride` was introduced to close, just in the error path. `useDefaultLayout.test.tsx` covers the resolution failure for `canSave` / `defaultLayout`, but not for the resulting scratch key.

### 5. `onPositionsChange` is both called by and a dependency of the same effect

`ArchitectureGraph.tsx` calls `onPositionsChange?.(...)` inside the parse effect and lists it in the dep array. It is safe today only because `DiagramSection` wraps it in a `useCallback` keyed on `viewportKey`. Any future call site passing an inline arrow gets an infinite render loop with no obvious cause. Worth either holding the callback in a `useRef`, or a prominent warning on the prop's JSDoc in `visualizer-contracts.ts`.

---

## Smaller notes

- `layout-service.ts`: non-404 GET failures reject with a generic `Error`; `useDefaultLayout` catches and settles `defaultLayout` to `null`, so a 500 or 403 is indistinguishable from "no default saved" and the user silently gets the auto-layout. Acceptable degradation, but the console-only error is the sole signal.
- `ArchitectureGraph`'s ReactFlow `key` now folds in `layoutEpoch`, so a successful save remounts the pane. Worth eyeballing that the persisted viewport store actually restores the user's pan/zoom rather than re-fitting after Save.
- `LayoutService` constructor: `this.ax = axiosInstance ?? apiClient` reads better than the if/else.
- `handleSaveLayout` / `handleResetLayout` are `useCallback`s depending on `[defaultLayoutState]`, which is a fresh object every render — the memo does nothing. Harmless (neither is passed to a memoized child), but misleading.
- The PR description is the untouched template: no description, no boxes ticked. Repo convention (`AGENTS.md`) asks for it populated with accurate status.

## Things done well

- Mapping both the server-side (`MongoWriteException` / `BSONObjectTooLarge`) and client-side (`BsonMaximumSizeExceededException`) 16MB paths to `capacityExceeded`, with javadoc explaining why the second exists only for the flat shape.
- `MongoLayoutIndexStep`'s tolerant `namespace_1` drop, and the reasoning for why a new step is needed rather than editing step 0 — exactly right, and matches the "shipped steps are immutable" rule.
- ADR 0005 is a good catch on the CODEOWNERS bypass, and the `for`-means-three-things observation is a strong argument.
- The `never()` assertion added to `TestMongoIndexInitializationStepShould` for `layouts` locks in the deliberate exclusion.

calm-hub/mongo/init-mongo.js
jpgough-ms: Rebased onto `upstream/main` now that #2935 is merged. `MongoLayoutIndexStep.fromVersion()` is now `9` (upstream occupies 2–8) and `LATEST_SCHEMA_VERSION` in `init-mongo.js` is now `10`, with the layouts index and javadoc updated to match.

jpgough-ms: Rebased onto `upstream/main` now that #2935 is merged. `MongoLayoutIndexStep.fromVersion()` is now `9` (upstream occupies 2–8) and `LATEST_SCHEMA_VERSION` in `init-mongo.js` is now `10`, with the layouts index and javadoc updated to match.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoLayoutIndexStep.java
jpgough-ms: Rebased onto `upstream/main` now that #2935 is merged. `MongoLayoutIndexStep.fromVersion()` is now `9` (upstream occupies 2–8) and `LATEST_SCHEMA_VERSION` in `init-mongo.js` is now `10`, with the layouts index and javadoc updated to match.

jpgough-ms:
Thanks for the thorough review — rebased onto `upstream/main` (#2935 merged) and addressed every point:

- **Issue 1 (orphan layouts)**: `LayoutResource.saveLayout` now 404s via a new `ArchitectureStore.architectureExists` check when the target architecture doesn't exist, before the store write. Checked on `PUT` only — `GET` for an unknown architecture already 404s via `layoutNotFoundResponse`. Both integration suites moved off the shared `finos` namespace into a dedicated `layout` namespace with an API-allocated architecture id (mirroring `MongoDocumentSizeLimitIntegration`'s pattern), since `EndToEndResource`/`NitriteEndToEndResource` aren't `restrictToAnnotatedClass` and share a database with `MongoArchitectureIntegration`/`NitriteArchitectureIntegration`, which assert `finos` has no architectures.
- **Issue 2 (unreachable DELETE)**: removed entirely instead of wiring it up (YAGNI) — endpoint, `LayoutStore.deleteLayout`, both backends, `LayoutNotFoundException`, and the frontend `deleteDefaultLayout` are gone. Flagged in the PR description: there's now no API path to remove a saved layout at all, only overwrite it — not a new restriction, since there's no architecture `DELETE` either.
- **Issue 3 (desktop errors invisible)**: replaced the wrapper `title` and the mobile-only in-menu text with one inline `role="alert"` banner in the main pane on both layouts, matching `Drawer`'s existing `dropError` pattern.
- **Issue 4 (slug key divergence)**: `viewportKeyOverride` is now `string | null | undefined` — `null` means "settled, no match," which suppresses `Drawer`'s slug fallback. Stays `undefined` while still resolving so the loading gate isn't affected.
- **Issue 5 (onPositionsChange effect/dep loop)**: `ArchitectureGraph` now holds it in a ref and calls through a stable wrapper — no `eslint-disable`.
- **Smaller notes**: `layout-service.ts` now returns status-specific messages (403/413/404/generic) instead of the log prefix, since that string is now what the new alert banner shows; `useDefaultLayout`'s result is memoised; dropped `layoutEpoch` from `ArchitectureGraph`'s `ReactFlow` `key` — turned out to be a real bug (it forced a remount on every save/reset, snapping the viewport back to whatever was saved at mount time), now covered by a regression test.
- **`this.ax = axiosInstance ?? apiClient`**: kept the `if/else` form — it's the majority convention in this codebase (`control-service`, `counts-service`, `interface-service`, `search-service`) and matches `calm-hub-ui/AGENTS.md`'s documented template. Happy to converge the other way as a separate sweep if you'd rather.

Full details in the updated PR description. All unit + integration tests pass locally (`../mvnw clean verify` and `../mvnw -P integration verify`, both green with JaCoCo's 90%/class gate met), and `npm test`/`npm run lint`/`npm run build` are clean on the frontend.

calm-hub-ui/src/service/layout-service.ts
```diff
@@ -0,0 +1,73 @@
...
+ * logging. `useDefaultLayout` assigns this straight to `saveError`, which
+ * DiagramSection now renders inline (see the layout-save-error banner) — so
+ * unlike the generic `errorMessage` prefix logged alongside it, this one has
+ * to mean something to someone who isn't reading the console.
+ */
+function layoutFailureMessage(action: 'save' | 'load', error: unknown): string {
+    const status = axios.isAxiosError(error) ? error.response?.status : undefined;
+    switch (status) {
+        case 403:
+            return `Couldn't ${action} the default layout — you don't have write access to this namespace.`;
```
markscott-ms: The switch is shared between save and load, so a failed load can say "you don't have write access to this namespace" or "it's too large to store". Neither is true on a read. Only the 500 and network cases are exercised by tests for load, so this isn't caught. Suggest restricting 403/413 to the save path:

```
case 403:
    return action === 'save'
        ? `Couldn't save the default layout — you don't have write access to this namespace.`
        : `Couldn't load the default layout — you don't have access to this namespace.`;
```

calm-hub/src/main/java/org/finos/calm/resources/LayoutResource.java
```diff
@@ -0,0 +1,157 @@
...
+     * so a save whose {@code for} silently names a different architecture than the URL can be
+     * rejected before it reaches the store. A layout with no {@code for} at all is accepted;
+     * the field's contract is defined entirely by this resource's own behaviour — compared
+     * against {@link #canonicalArchitecturePath} and rejected with a 400 on mismatch — not by
+     * an external schema document.
+     *
+     * @throws JsonParseException if the body is not valid JSON
+     */
+    private String parseForTarget(String layoutJson) {
+        Document parsed = Document.parse(layoutJson);
```
markscott-ms, line 160: parseForTarget calls `Document.parse(layoutJson)` with no null guard. An empty body should surface as `JsonParseException` → 400, but a null body would NPE into a 500. `NitriteLayoutStore.validateLayoutJson` explicitly guards null; the resource doesn't.

calm-hub/src/main/java/org/finos/calm/resources/LayoutResource.java
```diff
@@ -0,0 +1,157 @@
...
+            return CalmResourceErrorResponses.invalidNamespaceResponse(namespace);
+        }
+    }
+
+    /**
+     * Save (create or overwrite) the default layout for an architecture.
+     *
+     * @param namespace      the namespace the architecture belongs to
+     * @param architectureId the id of the architecture
+     * @param layoutJson     the layout as a raw JSON body, conforming to the draft layout schema
```
markscott-ms: still says "conforming to the draft layout schema" — the schema is exactly what ADR 0005 removed. Same doc block's @Operation description is fine.

markscott-ms (changes requested):

calm-hub-ui/src/service/layout-service.ts
jpgough-ms: Fixed — restricted the 403/413 wording to the save path; `getDefaultLayout`'s 403 now says "you don't have access to this namespace" instead of the write-access phrasing, and added a test for it (`rejects with a load-specific message on 403, not the save-side "write access" wording").

calm-hub/src/main/java/org/finos/calm/resources/LayoutResource.java
jpgough-ms: Fixed, and it's worse than a null-body NPE: an absent body actually binds to `layoutJson = ""` here (confirmed empirically), and `Document.parse("")` throws `BsonInvalidOperationException` — a type the existing `JsonParseException` catch doesn't see — so that was ALSO a 500, separately from the null case. `parseForTarget` now guards `null || isBlank()` up front, matching `MappingControllerResource`'s existing convention. Added two tests: an absent body and an explicit empty-string body, both now 400.

calm-hub/src/main/java/org/finos/calm/resources/LayoutResource.java
jpgough-ms: Fixed — reworded to match LayoutStore's own javadoc ("a CALM Hub-internal shape, not a validated CALM community schema") with a pointer to ADR 0005.

markscott-ms (approved):

PR #2958, opened 2026-08-08
fix(calm-hub): denormalize ADR title onto the header to fix N+1 search cost

### LINKED ISSUE #2946: calm-hub: ADR search costs 1 + 2N round trips per namespace
Split out of review on #2935 so it doesn't hold that PR up. Called out as a known trade in the PR description.

## Problem

`searchAdrCollection` resolves the latest revision for every ADR header it scans, and each resolution is two more database calls:

```
per header:  getLatestVersionContent -> listVersions + getVersion
total:       1 + 2N round trips for N ADRs in the namespace
```

The cost is uncapped until `MAX_RESULTS_PER_TYPE` trips, so a broad query over a namespace with many ADRs pays it in full. Every other resource type reads its summary straight off the header in one query — ADR is the only type that needs a version read at all, because its title lives in the revision rather than the header.

Affected on both backends:
- `calm-hub/src/main/java/org/finos/calm/store/mongo/MongoSearchStore.java` (`searchAdrCollection`)
- `calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteSearchStore.java` — worse here, since each call also acquires the read lock

For contrast, `CountsService`'s equivalent path has a 5s TTL cache. The search path has none.

## Why it isn't a one-line fix

The obvious remedy is a `latestRevision` (or denormalised title) pointer on the ADR header, so search reads one document per ADR. [ADR 0001](https://github.com/finos/architecture-as-code/blob/main/calm-hub/decisions/0001-versioned-artefact-storage.md) explicitly rejects a latest-version pointer on the header: it is a second source of truth for something the version collection already knows, and it has to be updated in the same breath as every version write, which is exactly the kind of two-write invariant the header/version split was meant to remove.

So the options are a real decision, not a tidy-up:

1. **Denormalise the ADR title onto the header** — narrower than a full `latestRevision` pointer, but still a second write on every revision, and still capable of going stale.
2. **Cache**, as `CountsService` does — no new invariant, bounded staleness, but does nothing for a cold cache.
3. **Batch the version reads** — one query for the latest revision of all N headers instead of 2N individual reads. Keeps a single source of truth; the work is in expressing "latest per group" on both backends.
4. **Revisit ADR 0001's rejection** for this case specifically, and record the outcome either way.

Option 3 looks the most promising, since it removes the round trips without introducing anything that can go stale.

## Acceptance

- ADR search no longer scales its round trips linearly with ADR count, on both Mongo and Nitrite
- Whichever option is taken, the reasoning is recorded against ADR 0001 (amended or superseded)
- Coverage for the chosen path on both backends

## Context

- Raised by @jpgough-ms in review of #2935
- Parent redesign: #2884

markscott-ms:
## Description
Postponed from #2935.

`searchAdrCollection` resolved every ADR header's latest revision to read its title, costing 2 extra DB round trips per header (`1 + 2N` total for `N` ADRs scanned).

Every other versioned type (Architecture, Pattern, Flow, Standard, Interface, Timeline) already denormalizes its display field onto the header at write time via `updateHeaderDetails`/`updatePresentHeaderDetails`. ADR was the one type that never did - its header was created with `createHeader(namespace, id, null, null)` and never updated again.

This PR makes ADR consistent with its six siblings instead of adding a new batched read path exclusively for it:

- `MongoAdrStore`/`NitriteAdrStore` now write the title onto the header on every create/update (first version falls back to `"Untitled ADR"` if none is given; a later revision that omits a title leaves the header's existing title standing, via `updatePresentHeaderDetails`'s existing no-op-on-blank behaviour).
- New one-time migration steps (`MongoAdrTitleBackfillStep` / `NitriteAdrTitleBackfillStep`, schema version 10 → 11) backfill every existing ADR header that has no title yet.
- `searchAdrCollection` on both backends now reads `header.name` directly — one query, structurally identical to `searchHeaderCollection`, no version-collection reads or Nitrite lock contention at all.
- `calm-hub/mongo/init-mongo.js` seed data and `LATEST_SCHEMA_VERSION` updated to match.
- New [ADR 0006](calm-hub/decisions/0006-denormalize-adr-title-onto-header.md) records the reasoning, including why this doesn't reopen [ADR 0001](calm-hub/decisions/0001-versioned-artefact-storage.md)'s rejection of a `latestVersion` pointer — it follows the precedent 0001 already set for `versionCount` (a real, read-heavy consumer justifies a narrow denormalized field), not a pointer into the version collection.

Closes #2946

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Commit Message Format ✅
Commit follows `fix(calm-hub): ...` conventional commit format.

## Testing
- [x] All existing tests pass

Ran `../mvnw clean verify -Ddependency-check.skip=true` — 2616 tests, 0 failures, JaCoCo 90%-per-class coverage gate passes on all touched classes. Not run: `-P integration verify` (requires Docker, not available in this environment) — worth a maintainer/CI check given the migration step touches real Mongo/Nitrite behavior.

## Checklist
- [x] I have updated documentation if necessary (new ADR 0006 + decisions/README.md index)

### DIFF
--- a/calm-hub/decisions/0006-denormalize-adr-title-onto-header.md
+++ b/calm-hub/decisions/0006-denormalize-adr-title-onto-header.md
@@ -1,0 +1,95 @@
+# ADR 0006: Denormalize ADR title onto the header
+
+**Status**: Implemented.
+
+## Context
+
+[#2946](https://github.com/finos/architecture-as-code/issues/2946) (split out of
+review on [#2935](https://github.com/finos/architecture-as-code/pull/2935), parent
+redesign [#2884](https://github.com/finos/architecture-as-code/issues/2884)):
+`MongoSearchStore`/`NitriteSearchStore`'s `searchAdrCollection` resolves the latest
+revision's title for every ADR header it scans — `1 + 2N` round trips for `N` headers
+scanned, uncapped until 50 *matches* are found (the cap doesn't bound headers scanned).
+On Nitrite it's worse: 2 lock acquisitions and 2 unindexed full-collection scans per
+header, since Nitrite creates no indexes at all.
+
+ADR is the only versioned type this happens to, because it's the only one whose header
+carries no denormalized display field. Every other type — Architecture, Pattern, Flow,
+Standard, Interface, Timeline — writes its `name`/`description` onto the header via
+`{Mongo,Nitrite}VersionDocumentStore#updateHeaderDetails`/`updatePresentHeaderDetails`
+on every version write (e.g. `MongoArchitectureStore.java:156`,
+`MongoPatternStore.java:146`), so their search path reads the header alone, in one
+query. ADR's header is created with `createHeader(namespace, id, null, null)` and never
+updated again — its title has only ever lived in the revision content.
+
+[ADR 0001](0001-versioned-artefact-storage.md) rejected a `latestVersion` pointer on the
+header: "a cached pointer would be one more place for staleness bugs with no current
+consumer." That same ADR, in the same section, later **reversed** an equivalent
+rejection for `versionCount` once a real consumer existed — `calm-hub-ui`'s
+`ItemCard.tsx` reads it on every namespace listing page view — reasoning that reads
+vastly outnumber writes for that field. ADR title now has exactly that shape of
+consumer: search, at scale, versus a title write only on ADR creation or a subsequent
+revision. This ADR follows 0001's own precedent rather than reopening its `latestVersion`
+rejection: it denormalizes a resolved *value* (the title string), not a pointer into the
+version collection, and only for ADR — the one type actually missing it.
+
+Four options were on the table (batch the version reads; cache with a TTL like
+`CountsService`'s; revisit 0001's rejection generically; or this one). Batching was the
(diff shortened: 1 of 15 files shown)


markscott-ms:
Addressed in c1aed17e: both search stores now fall back to the placeholder title on a blank `name` as well as a null one, and `MongoAdrTitleBackfillStep` filters untitled headers in memory (matching `NitriteAdrTitleBackfillStep`) rather than a Mongo query that couldn't express "blank or missing" without leaving whitespace-only titles stuck forever. Added coverage for the blank case in both search-store tests and the Mongo backfill step test. Full `../mvnw clean verify` still green (2619 tests, JaCoCo gate passes).

rocketstack-matt (approved):

PR #2959, opened 2026-08-08
refactor(calm-hub): consolidate requireNamespace into NamespaceStore default method

### LINKED ISSUE #2947: calm-hub: requireNamespace is reimplemented in 14 store classes
Split out of review on #2935. Pre-dates that PR — it exists across all seven resource types on both backends, including Architecture and Pattern, which were migrated earlier.

## Problem

The same namespace guard is written out privately in **14** store classes (the review estimated ~10):

**Mongo** — `MongoAdrStore`, `MongoArchitectureStore`, `MongoFlowStore`, `MongoInterfaceStore`, `MongoPatternStore`, `MongoStandardStore`, `MongoTimelineStore`

**Nitrite** — `NitriteAdrStore`, `NitriteArchitectureStore`, `NitriteFlowStore`, `NitriteInterfaceStore`, `NitritePatternStore`, `NitriteStandardStore`, `NitriteTimelineStore`

```java
private void requireNamespace(String namespace) throws NamespaceNotFoundException {
    if (!namespaceStore.namespaceExists(namespace)) {
        throw new NamespaceNotFoundException();
    }
}
```

A change to the contract — caching the existence check, adding a metric, changing what gets thrown — needs 14 edits in lockstep, and nothing fails if one is missed.

## Not quite identical

Worth knowing before extracting: the two backends' copies differ. The Nitrite versions log a warning before throwing; the Mongo versions throw silently. Unifying them is therefore a small behaviour decision (keep the logging for both, or drop it) rather than a pure move, and it will shift log output for the Mongo backend either way.

## Why it wasn't done in #2935

That PR is already 17 commits across both backends and seven types. Touching all 14 stores to unify a guard it didn't introduce would have widened the diff well past what the redesign needs, and made the storage-shape changes harder to review on their own.

## Acceptance

- One shared helper (a base class or a small utility taking the `NamespaceStore`) used by all 14 stores
- The Mongo/Nitrite logging difference resolved deliberately, and the choice noted
- Existing namespace-not-found coverage still passes on both backends

## Context

- Raised by @jpgough-ms in review of #2935
- Parent redesign: #2884

markscott-ms:
## Description
Consolidates the `requireNamespace` guard that was duplicated privately across 16 store classes (7 Mongo + 7 Nitrite resource stores, plus the Mongo/Nitrite layout stores) into a single shared implementation: a **default method on `org.finos.calm.store.NamespaceStore`**, `requireNamespace(String)`. Call sites now read `namespaceStore.requireNamespace(x)` — a change to the contract needs one edit instead of 16.

As a result, whilst the PR is 34 files, a large number of them are a formulaic and similar change.

The two backends' copies weren't quite identical: Nitrite logged a WARN before throwing `NamespaceNotFoundException`, Mongo (mostly) threw silently. This PR deliberately unifies the behavior to **warn-then-throw for all backends** — the warning is now logged from `NamespaceStore`'s own logger rather than the individual store class.

`MongoLayoutStore`/`NitriteLayoutStore` weren't in the issue's literal list of 14 stores but have the identical duplicated method, so they're included here too for full consistency (no duplicate left behind).

Closes #2947

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing
- [x] All existing tests pass

New `TestNamespaceStoreShould` covers the default method directly. Store-level test mocks needed a small adjustment: Mockito stubs an interface default method to a no-op by default, so each store test's `@BeforeEach` now opts the mocked `NamespaceStore` into running the real `requireNamespace` body via `doCallRealMethod()`, which then defers to the already-mocked `namespaceExists()` — no change to what each test actually asserts. Ran the full calm-hub build (`../mvnw clean verify -Ddependency-check.skip=true`, which includes the 90%-per-class JaCoCo gate): 2593 tests, 0 failures, BUILD SUCCESS.

## Checklist


calm-hub/src/main/java/org/finos/calm/store/NamespaceStore.java
```diff
@@ -3,14 +3,34 @@
...
+     *
+     * @param namespace the namespace to validate
+     * @throws NamespaceNotFoundException if no namespace with the given name exists
+     */
+    default void requireNamespace(String namespace) throws NamespaceNotFoundException {
+        if (!namespaceExists(namespace)) {
+            NAMESPACE_STORE_LOG.warn("Namespace '{}' not found", namespace);
+            throw new NamespaceNotFoundException();
+        }
+    }
```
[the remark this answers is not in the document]
markscott-ms: I reject this change - to do so would introduce a logger lookup on every time the log message is invoked - which should not be often in normal operation but adds processing to abusive usage.

Defining the property as a field yes is part of the public interface, but I judge this acceptable as the interface is used only in this application - it is not a library - so there is no real hindrance it causes.

rocketstack-matt (commented):
Clean, mechanical refactor — meets #2947's acceptance criteria (one shared guard, Mongo/Nitrite logging deliberately unified, coverage preserved). No defects found in a critical pass or an independent Copilot CLI pass.

One completeness note (non-blocking, no line to anchor since the file isn't touched by this diff): `MongoDecoratorStore.validateNamespace` / `NitriteDecoratorStore.validateNamespace` are the same pattern this PR consolidates — check `namespaceStore.namespaceExists`, warn-log, throw `NamespaceNotFoundException`, already logging-unified between backends, 5 call sites each. Not in #2947's literal list, but the PR already went beyond that list for the Layout stores "for full consistency (no duplicate left behind)" — same rationale applies here. Worth folding in now or calling out as deliberate follow-up scope. (`MongoResourceMappingStore` repeats the identical inline check 6x with no extraction — same theme, lower priority since it was never even extracted to a helper.)

markscott-ms:
THanks @rocketstack-matt I think I started 2959 before the decorators had been merged, so missed it. Will fix.

markscott-ms:
@rocketstack-matt Folded in — `MongoDecoratorStore`/`NitriteDecoratorStore`.`validateNamespace` and the inline checks in `MongoResourceMappingStore`/`NitriteResourceMappingStore` now all go through `namespaceStore.requireNamespace(...)` too. Same rationale as the Layout stores: no duplicate left behind. Full `../mvnw clean verify -Ddependency-check.skip=true` passes (2593 tests, 0 failures, JaCoCo gate green).

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteDecoratorStoreShould.java
```diff
@@ -50,7 +52,8 @@ class TestNitriteDecoratorStoreShould {
     private NitriteDecoratorStore decoratorStore;
 
     @BeforeEach
-    void setUp() {
+    void setUp() throws NamespaceNotFoundException {
+        lenient().doCallRealMethod().when(namespaceStore).requireNamespace(anyString());
```
jpgough-ms: This is the only one of 20 refactored test classes that wraps the `requireNamespace` stub in `lenient()`. The other 19 rely on every test method happening to touch `namespaceStore` — add a namespace-independent test method to any of them (e.g. a constructor or logging test) and Mockito's strict-stubbing check throws `UnnecessaryStubbingException`. Either wrap all 20 stubs in `lenient()` for consistency, or drop it here if this class doesn't need it.

jpgough-ms (commented):
Automated review pass. One off-diff note plus a few lower-priority observations below.

**Off-diff note** — `MongoUserAccessStore` and `NitriteUserAccessStore` weren't migrated: both still contain the exact `if (!namespaceStore.namespaceExists(x)) throw NamespaceNotFoundException` shape this PR eliminated everywhere else. Drop-in replacement: `namespaceStore.requireNamespace(namespace)`, guarded by the existing `GLOBAL_ACCESS` special case as with the other 6 call sites. See `MongoUserAccessStore.java:137` and `NitriteUserAccessStore.java:141`.

**Other findings** (not blocking, noted for awareness):
- `NamespaceStore.java:143` — the require-or-throw pattern is only generalized for namespaces; Mongo/Nitrite pairs for Flow, Architecture, Pattern, Standard, Interface, Timeline, and Adr still duplicate near-identical `requireX`/`requireXExists` helpers.
- `NamespaceStore.java:13` — `NAMESPACE_STORE_LOG` is declared as a public logger constant on the interface, breaking the codebase's usual private-static-final-`Logger`-per-class convention.
- `MongoDecoratorStore.java:269` — the shared default method logs a generic "Namespace not found" message under one `NamespaceStore` logger category, losing the per-class/operation context the deleted helpers used to log (e.g. "not found when retrieving decorators").
- `pom.xml:97` — this diff bundles the `NamespaceStore` refactor with an unrelated `server-only` Maven profile, a `flows.md` docs addition, and a `package-lock.json` sync.
- `AGENTS.md:23` — the `server-only` profile's UI-copy/gitignore/`mvn clean` explanation is duplicated near-verbatim across `pom.xml`, `README.md`, and `AGENTS.md`.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteDecoratorStoreShould.java
markscott-ms: Fixed — removed `should_log_initialization_message()`. It never actually verified the log message (per its own comment), and its only assertion (`db.getCollection("decorators")` was called) duplicated what every other test in the class already exercises via `@BeforeEach`. It was also the sole reason `lenient()` was needed here, since it never touched `namespaceStore`. With it gone, the stub is back to a plain `doCallRealMethod()`, matching the other 19 refactored test classes.

markscott-ms:
> **Off-diff note** — `MongoUserAccessStore` and `NitriteUserAccessStore` weren't migrated: both still contain the exact `if (!namespaceStore.namespaceExists(x)) throw NamespaceNotFoundException` shape this PR eliminated everywhere else. Drop-in replacement: `namespaceStore.requireNamespace(namespace)`, guarded by the existing `GLOBAL_ACCESS` special case as with the other 6 call sites. See `MongoUserAccessStore.java:137` and `NitriteUserAccessStore.java:141`.
> 
> **Other findings** (not blocking, noted for awareness):
> 
> * `NamespaceStore.java:143` — the require-or-throw pattern is only generalized for namespaces; Mongo/Nitrite pairs for Flow, Architecture, Pattern, Standard, Interface, Timeline, and Adr still duplicate near-identical `requireX`/`requireXExists` helpers.

Will address in separate PR.

> * `NamespaceStore.java:13` — `NAMESPACE_STORE_LOG` is declared as a public logger constant on the interface, breaking the codebase's usual private-static-final-`Logger`-per-class convention.

Java does not support `private` fields on interfaces. Which either means you have a public field, or a private static method that instantiates a logger on every call. Given that this interface is not exposed beyond CALMHub, i.e. this isn't a library, I've considered a public field is fine because we don't need to worry about API breakage as much.

> * `MongoDecoratorStore.java:269` — the shared default method logs a generic "Namespace not found" message under one `NamespaceStore` logger category, losing the per-class/operation context the deleted helpers used to log (e.g. "not found when retrieving decorators").

Accepted loss of specificity in the log message.

> * `pom.xml:97` — this diff bundles the `NamespaceStore` refactor with an unrelated `server-only` Maven profile, a `flows.md` docs addition, and a `package-lock.json` sync.
> * `AGENTS.md:23` — the `server-only` profile's UI-copy/gitignore/`mvn clean` explanation is duplicated near-verbatim across `pom.xml`, `README.md`, and `AGENTS.md`.

Reviewer had own files in active branch. Not applicable to this PR.

calm-hub/src/test/java/org/finos/calm/store/nitrite/TestNitriteDecoratorStoreShould.java
markscott-ms: `lenient()` was used for a pointless test. Test and `lenient()` removed

jpgough-ms (approved):

PR #2962, opened 2026-08-09
feat(calm-hub): migrate Control storage to the header/version shape

markscott-ms:
## Description
Closes the last gap in #2884: `Control` was the one type still on the old one-document-per-domain, double-nested shape (`domain -> control -> configurations[]`), risking MongoDB's 16MB document limit exactly as the seven already-migrated types did before ADR 0001.

Composes `MongoVersionDocumentStore`/`NitriteVersionDocumentStore` twice per store — once for a control's requirement (`namespace=domain`), once for its configurations (`namespace=domain::controlId`, a synthetic composite since `configurationId` is already a global counter) — rather than reusing the shared `MongoVersionSplitMigration`/`NitriteVersionSplitMigration` fan-out, which only handles one nesting level under a field literally named `namespace`. See [ADR 0007](calm-hub/decisions/0007-control-storage-header-version-split.md) for the full design.

Also fixes `MongoSearchStore`/`NitriteSearchStore`'s control search, which read the old nested shape directly and would otherwise have silently stopped finding any control once the collection held headers instead.

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Commit Message Format ✅
Commits follow Conventional Commits (`feat(calm-hub): ...`, `refactor(calm-hub): ...`).

## Testing
- [x] All existing tests pass

`../mvnw clean verify -Ddependency-check.skip=true` (unit suite + 90%-per-class JaCoCo gate) and `../mvnw -P integration verify` (TestContainers, incl. rewritten `MongoControlIntegration`/`NitriteControlIntegration`) both pass.

## Checklist
- [x] I have updated documentation if necessary (ADR 0007, ADR 0004 status note, `decisions/README.md`)


calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        LOG.info("Control version split complete: {} domain document(s) fanned out into "
+                        + "{} control header(s) and {} configuration header(s)",
+                oldDocumentIds.size(), migratedControls, migratedConfigurations);
+    }
+
+    private void writeControlHeaderAndVersions(MongoCollection<Document> headers, MongoCollection<Document> versions,
+                                               String domain, Document control) {
+        Integer controlId = control.getInteger(CONTROL_ID_FIELD);
```
jpgough-ms: **Correctness:** Uncaught `ClassCastException` on a non-numeric `controlId`.

`control.getInteger(CONTROL_ID_FIELD)` throws if the stored value isn't a Mongo Integer (e.g. legacy/hand-edited data), with no surrounding try/catch. `SchemaMigrationRunner` catches the resulting `RuntimeException`, logs it, and leaves the schema version unadvanced and the migration lock held — `SchemaMigrationInProgressFilter` then 503s all requests. Every restart retries the same document and throws again, permanently blocking the app until an admin fixes the data by hand.

calm-hub/src/main/java/org/finos/calm/migration/steps/NitriteControlSplitMigration.java
```diff
@@ -0,0 +1,198 @@
...
+                        + "{} control header(s) and {} configuration header(s)",
+                oldDocumentIds.size(), migratedControls, migratedConfigurations);
+    }
+
+    private void writeControlHeaderAndVersions(String domain, Document control) {
+        Integer controlId = control.get(CONTROL_ID_FIELD, Integer.class);
+        Document storedVersions = control.get(REQUIREMENT_FIELD, Document.class);
+        Map<String, String> keysByCanonicalVersion = collapseToCanonicalVersions(storedVersions, domain, controlId);
+
+        Filter headerFilter = Filter.and(where(NAMESPACE_FIELD).eq(domain), where(CONTROL_ID_FIELD).eq(controlId));
```
jpgough-ms: **Correctness:** Header/version writes use non-atomic `remove()` + `insert()` instead of a single upsert.

Repeated for control versions, config header, and config versions — two separate Nitrite calls with nothing wrapping them. A process kill between `remove` and `insert` permanently deletes that document. `update(filter, doc, UpdateOptions.updateOptions(true))` gives a true single-call upsert and is already used elsewhere (`NitriteVersionDocumentStore.upsertVersion`). Same pattern also in `NitriteVersionSplitMigration.java:130-131,141-143`.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
```diff
@@ -0,0 +1,280 @@
...
+            Document oldDocument = oldDomainDocs.find(Filters.eq("_id", oldDocumentId)).first();
+            if (oldDocument == null) {
+                continue;
+            }
+            String domain = oldDocument.getString(DOMAIN_FIELD);
+            for (Document control : oldDocument.getList(CONTROLS_FIELD, Document.class, List.of())) {
+                writeControlHeaderAndVersions(controlHeaders, controlVersions, domain, control);
+                migratedControls++;
+
+                String configNamespace = domain + "::" + control.getInteger(CONTROL_ID_FIELD);
```
jpgough-ms: **Correctness:** Missing `controlId` silently builds the namespace `"<domain>::null"` instead of failing.

`domain + "::" + control.getInteger(CONTROL_ID_FIELD)` returns `null` (not a throw) when `controlId` is absent, so concatenation yields the literal string `"<domain>::null"`. `MongoControlStore.configurationNamespace(domain, int controlId)` takes a primitive `int` and can never reproduce that string, so any configs migrated under it become permanently unreachable via any API. Requires legacy/corrupted data missing `controlId` — plausible rather than confirmed, since normal application writes can't produce it.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
```diff
@@ -0,0 +1,280 @@
...
+        createNewIndexes();
+    }
+
+    /**
+     * Replaces the old one-document-per-domain constraint with the four the new shape needs,
+     * in one call. Only safe against a database with no old-shape documents left — see the
+     * class javadoc's "Why the new unique indexes have to wait" section. {@link #migrate()}
+     * does not call this; it does the same two things with the fan-out in between.
+     */
+    public void transitionIndexes() {
```
jpgough-ms, line 100: **Correctness:** `transitionIndexes()`'s "no old-shape documents left" precondition is enforced only in a javadoc comment, not in code.

It calls `dropOldDomainIndex` then `createNewIndexes()` directly with no runtime check. If a future caller (new test harness, ops script, or a change to `EndToEndResource`'s step ordering) invokes it against a database with unmigrated domain documents still present, the `(namespace, controlId)` unique index collides on `(null, null)` and throws — leaving the migration lock held and the app permanently 503'ing. Correctly ordered today in the only real caller, but this class has already had two prior fix commits for step-ordering bugs, so a code-level guard is worth adding rather than relying on the comment.

jpgough-ms (commented):
Automated review pass (medium effort) over the Control storage migration changes. 7 findings below: 4 correctness, 3 duplication/simplification.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
```diff
@@ -0,0 +1,280 @@
...
+        }
+        return content;
+    }
+
+    /**
+     * Maps each canonical version to the stored key it came from, keeping the first when
+     * several collapse onto one. See {@link MongoVersionSplitMigration#collapseToCanonicalVersions}
+     * for why the old shape can hold several keys meaning one version.
+     */
+    private Map<String, String> collapseToCanonicalVersions(Document storedVersions, String scopeLabel, Integer id) {
```
jpgough-ms, line 304: **Simplification:** `collapseToCanonicalVersions` is reimplemented verbatim in three places.

The canonicalise/`putIfAbsent`/warn-on-collision algorithm already exists as `MongoVersionSplitMigration.collapseToCanonicalVersions` (lines 268-279); this class (263-279) and `NitriteControlSplitMigration.java` (181-197) copy it verbatim instead of sharing a helper. A future collision-handling fix has to land in three places — the class's own javadoc already warns this exact failure mode (a fix applied to one copy, missed in another) has happened before here.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoControlStore.java
```diff
@@ -112,289 +79,147 @@ public List<ControlDetail> getControlsForDomain(String domain) throws DomainNotF
...
-        Document versionDoc = (Document) requirement.get(latestKey);
-        return versionDoc != null ? versionDoc.getString("title") : null;
+    /**
+     * Configurations are scoped under a synthetic namespace combining the domain and owning
+     * control id, so a control's configurations can be listed without a second nesting level
+     * — {@code configurationId} is already globally unique (see class javadoc), so this
+     * exists purely to scope listing, not to establish uniqueness. {@code ":"} is safe: the
+     * domain name pattern ({@code DomainStore}'s validation) disallows it.
+     */
+    private String configurationNamespace(String domain, int controlId) {
```
jpgough-ms, line 218: **Simplification:** The `domain + "::" + controlId` namespace-scoping convention is inlined identically in four places.

`MongoControlStore.configurationNamespace` (217-219), `NitriteControlStore.configurationNamespace` (211-213, whose own javadoc calls out that it's "identical value on both backends"), `MongoControlSplitMigration.java:162`, and `NitriteControlSplitMigration.java:85` all hand-build this string separately. A change to the separator or scoping scheme has to be applied in all four and kept in sync only by convention — a mismatch between the migration's copy and the store's copy would silently corrupt data rather than fail a test.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
```diff
@@ -0,0 +1,280 @@
...
+                    versionDocument, upsert);
+        }
+    }
+
+    /**
+     * Reads one version's stored content, preserving whatever is there rather than casting.
+     * See {@link MongoVersionSplitMigration#contentOf} for why: a migration is the wrong place
+     * to abort over one malformed document.
+     */
+    private Object contentOf(Document storedVersions, String key, String scopeLabel, Integer id) {
```
jpgough-ms, line 288: **Simplification:** `contentOf` (preserve non-Document/non-String content with a WARN log) is duplicated from the sibling version migrations rather than reused.

`MongoControlSplitMigration.contentOf` (247-256) duplicates `MongoVersionSplitMigration.contentOf` (240-250) near-verbatim; `NitriteControlSplitMigration.contentOf` (166-175) duplicates the Nitrite version-migration equivalent. A wording/behavior change to the malformed-content warning needs coordinated edits across files. Flagged as plausible rather than confirmed: the Mongo and Nitrite variants check different types (`Document` vs `String`), so a single cross-backend helper isn't trivial — only a per-backend one is.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
markscott-ms: Fixed in 61e75102: both migration classes now validate `controlId`/`configurationId` via a `requireId` helper before use, throwing a clear `IllegalStateException` naming the exact domain/control rather than letting a raw `ClassCastException` (wrong type) propagate uncaught. Added regression tests for missing and non-numeric ids.

calm-hub/src/main/java/org/finos/calm/migration/steps/NitriteControlSplitMigration.java
markscott-ms: Fixed in 61e75102: header/version writes now use a single atomic `update(filter, doc, UpdateOptions.updateOptions(true))` call instead of separate `remove()`+`insert()`, matching the pattern already used in `NitriteVersionDocumentStore`. Added a regression test asserting `insert()`/`remove()` are never called.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
markscott-ms: Fixed in 61e75102, same `requireId` validation as the sibling comment above — a missing/non-numeric `controlId` now throws before a configuration namespace is ever built, so `"<domain>::null"` can no longer be produced.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
markscott-ms: Fixed in 61e75102: added `assertNoOldShapeDocumentsRemain`, called from `createNewIndexes()`, which throws `IllegalStateException` if an old-shape domain document is still present. Added a direct test (`refuse_to_build_the_new_indexes_while_an_old_shape_document_still_exists`) that calls `transitionIndexes()` against a database with old-shape data and asserts it throws.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoControlStore.java
markscott-ms: Fixed in 61e75102: extracted into `ControlConfigurationNamespace.of(domain, controlId)` (new file, with its own unit test), now the single source of truth used by `MongoControlStore`, `NitriteControlStore`, `MongoControlSplitMigration`, and `NitriteControlSplitMigration`.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
markscott-ms: Acknowledged, left as-is deliberately. This duplicates MongoVersionSplitMigration's method by design — see ADR 0007's "The migration fan-out is new code, not a reuse of MongoVersionSplitMigration" section: that shared class is used by the seven already-migrated types.

Generally, I don't like touching migration steps that may have already been executed, and consider them immutable. Risk of introducing different behaviour.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoControlSplitMigration.java
markscott-ms: Same call as the `collapseToCanonicalVersions` comment above — left as-is deliberately, same ADR 0007 rationale.

jpgough (approved):

jpgough-ms (approved):

PR #2969, opened 2026-08-10
feat(calm-hub): add layout persistence support for patterns

jpgough-ms:
## Description

- Add a pattern layout store/resource/migration, structurally mirroring the existing per-namespace architecture layout stack
- Wire save/reset default layout actions through to patterns end-to-end (DiagramSection -> Drawer -> PatternVisualizer -> PatternGraph)
- Key scratch-position and viewport storage by namespace+calmType+id, since architecture and pattern ids are drawn from independent counters and could otherwise collide; migrate pre-existing localStorage entries forward on first read
- Return resource-specific wording (architecture vs pattern) from the shared layout-mismatch error response

### Review round 2 — addressed all feedback from @markscott-ms

- Fixed a real correctness bug: `Drawer.tsx` forwarded `defaultLayout`/`layoutEpoch` unconditionally, so a dropped file inherited the previously-loaded resource's saved layout; both now collapse alongside `viewportKey` when a file is dropped
- Added `PatternGraph.test.tsx` (mirroring `ArchitectureGraph.test.tsx`'s precedence/gating coverage) and `Drawer.test.tsx` regression tests for the above — verified they fail against the pre-fix code
- Extracted `buildViewportKey` as the single source of truth for the `namespace/calmType/id` key shape, replacing hand-duplicated templates in `Drawer.tsx` and `useDefaultLayout.ts`
- Parameterized `CalmResourceErrorResponses`'s not-found responses on `resourceType`, replacing four near-identical architecture/pattern methods (message text unchanged)
- Pushed the target-existence check into `LayoutStore`/`PatternLayoutStore#upsertLayout`, removing a duplicate namespace-existence round trip on every layout save; removed the now-dead `architectureExists`/`patternExists`
- Resolved `PatternLayoutResource`/`PatternLayoutStore`'s circular javadoc cross-reference
- Rebased onto `main` now that #2971 has merged, and renumbered `MongoPatternLayoutIndexStep` to `fromVersion() == 12` / `LATEST_SCHEMA_VERSION` `13`, since #2971's `MongoResourceMappingIndexStep` took `11` first

## Type of Change

## Affected Components

- [x] CALM Hub (`calm-hub/`)

## Testing

- [x] All existing tests pass

- 2690 unit tests pass (`../mvnw test`)
- 526 integration tests pass (`../mvnw -P integration verify`, TestContainers Mongo), JaCoCo coverage gate met
- `npm test --workspace calm-hub-ui`: 115 files / 1395 tests pass
- `npm run lint` (all workspaces): 0 errors

## Checklist

### DIFF
--- a/calm-hub-ui/src/hub/hooks/useDefaultLayout.ts
+++ b/calm-hub-ui/src/hub/hooks/useDefaultLayout.ts
@@ -1,31 +1,39 @@
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { CalmService } from '../../service/calm-service.js';
-import { LayoutService } from '../../service/layout-service.js';
+import { LayoutService, LayoutResourceType } from '../../service/layout-service.js';
 import { isSlug } from '../../model/calm.js';
 import { CalmLayout, pinsToStoredPositions, storedPositionsToPins } from '../../model/layout.js';
 import { clearStoredNodePositions, StoredNodePosition } from '../../visualizer/services/node-position-service.js';
 
 export interface UseDefaultLayoutResult {
     /**
-     * Namespace + resolved numeric architecture id — the single key both the
+     * Namespace + calmType + resolved numeric id — the single key both the
      * localStorage scratch layer and the server layout are addressed by, so a
      * diagram reached via a slug route and via its numeric route share one
-     * scratch entry and one server call. Three states:
+     * scratch entry and one server call. The calmType component exists
+     * because architecture ids and pattern ids are allocated from independent
+     * counters — an Architecture 7 and a Pattern 7 can coexist in the same
+     * namespace, and without a type qualifier they would collide on one
+     * scratch-position entry and one viewport entry. `Drawer`'s own fallback
+     * key computation must stay in this same `namespace/calmType/id` shape
+     * for the same reason. Three states:
      *  - a string: the resolved key.
-     *  - `undefined`: not applicable (a pattern or dropped file), or a slug that
-     *    is still resolving — callers with a loading state to show key off this.
-     *  - `null`: this *is* an architecture, but resolution finished without a
-     *    numeric id (an unresolvable slug, or the mapping fetch failed). Never
-     *    `undefined` here — a caller that fell back to some other key (e.g. the
-     *    raw slug) in this state would reintroduce the exact split this key
-     *    exists to prevent. See `Drawer`'s `viewportKeyOverride` handling.
+     *  - `undefined`: not applicable (a dropped file, or a calmType this hook
+     *    doesn't support), or a slug that is still resolving — callers with a
+     *    loading state to show key off this.
+     *  - `null`: this *is* a supported type, but resolution finished without
+     *    a numeric id (an unresolvable slug, or the mapping fetch failed).
+     *    Never `undefined` here — a caller that fell back to some other key
+     *    (e.g. the raw slug) in this state would reintroduce the exact split
+     *    this key exists to prevent. See `Drawer`'s `viewportKeyOverride`
+     *    handling.
      */
     viewportKey: string | null | undefined;
     /** undefined = loading, null = none stored, else the resolved default layout. */
     defaultLayout: StoredNodePosition[] | null | undefined;
     /** Bumped after a successful save or a reset, to force a clean re-apply. */
     layoutEpoch: number;
-    /** False while a slug id hasn't resolved to a numeric architecture id — Save stays disabled. */
+    /** False while a slug id hasn't resolved to a numeric id — Save stays disabled. */
     canSave: boolean;
     saving: boolean;
     saveError: string | null;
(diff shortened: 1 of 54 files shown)


calm-hub/mongo/init-mongo.js
```diff
@@ -141,7 +141,7 @@ logSection("Schema baseline");
 // Raise LATEST_SCHEMA_VERSION whenever a migration step is added, and seed that step's
 // target shape below. Document shape must match MongoSchemaVersionStore: _id
 // "schemaVersion", int version, in the calm collection.
-const LATEST_SCHEMA_VERSION = 11;
+const LATEST_SCHEMA_VERSION = 12;
```
markscott-ms: this is going to clash with PR 2971.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoPatternLayoutIndexStep.java
```diff
@@ -0,0 +1,87 @@
...
+    @ConfigProperty(name = "calm.database.mode", defaultValue = "mongo")
+    String databaseMode;
+
+    public MongoPatternLayoutIndexStep(MongoDatabase database) {
+        this.database = database;
+    }
+
+    @Override
+    public int fromVersion() {
+        return 11;
```
markscott-ms: Clash with PR 2971. Be careful merging - whichever is first will mean the other needs updating.

calm-hub/src/main/java/org/finos/calm/resources/PatternLayoutResource.java
```diff
@@ -0,0 +1,146 @@
...
+import org.finos.calm.store.PatternStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.finos.calm.resources.ResourceValidationConstants.NAMESPACE_MESSAGE;
+import static org.finos.calm.resources.ResourceValidationConstants.NAMESPACE_REGEX;
+
+/**
+ * Resource for managing the shared, default layout of a pattern in a given namespace. A
+ * structural twin of {@link LayoutResource} — see {@link PatternLayoutStore}'s class javadoc
```
markscott-ms: Hey, you're sending me in loops... PatternLayoutResource tells me to see PatternLayoutStore as to why there isn't just one layouts collection... and that class tells me to read here!

Can you adjust the comments to be less infinite loopy?

markscott-ms (changes requested):

markscott-ms (commented):
Automated code review findings - I filtered a few out, below and inline are what I kept, but a couple of them may be nits-too-far. Use judgement.

**test-coverage** (general, not tied to a single line): `PatternGraph.tsx`/`PatternVisualizer.tsx` have no direct test coverage for the new default-layout logic (`awaitingDefaultLayout` gating, position precedence, `onPositionsChange` reporting), unlike `ArchitectureGraph.tsx`'s equivalent logic which is directly unit tested. `DiagramSection.test.tsx` mocks `Drawer` entirely and `Drawer.test.tsx` mocks `PatternVisualizer` entirely, so this new logic path is never exercised under test — a regression in the awaiting/precedence logic (e.g. the dropped-file stale-layout issue flagged separately on `Drawer.tsx`) would not be caught by the existing suite.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
```diff
@@ -270,6 +274,12 @@ export function Drawer({
                                 onEdgeClick={handlePatternEdgeClick}
                                 onBackgroundClick={closeSidebar}
                                 viewportKey={viewportKey}
+                                defaultLayout={defaultLayout}
```
markscott-ms: **correctness**: `defaultLayout`/`layoutEpoch` are forwarded to `PatternVisualizer`/`ReactFlowVisualizer` unconditionally, but only `onPositionsChange` is guarded against a dropped file, so a dropped file inherits the previously-loaded resource's saved layout.

Scenario: a user has Architecture/Pattern X open (with a saved default layout) and drag-and-drops a locally edited copy of the same file (the documented use case per the nearby comment). `fileInstance` becomes truthy so `viewportKey` collapses to `undefined` and `awaitingDefaultLayout` is `false`, but the `defaultLayout` prop is untouched — the graph's effect computes `effectivePositions = localPositions ?? defaultLayout ?? null` and applies the stale, unrelated saved-layout positions (matched by node id) onto the freshly dropped file's parsed nodes instead of using a fresh auto-layout.

calm-hub/src/main/java/org/finos/calm/resources/CalmResourceErrorResponses.java
```diff
@@ -54,16 +54,39 @@ public static Response layoutNotFoundResponse(String namespace, int architecture
                 .build();
     }
 
+    /**
+     * Returns a 404 response for a pattern layout write against a pattern id that does not
+     * exist. Mirrors {@link #architectureNotFoundResponse(String, int)}.
+     */
+    public static Response patternNotFoundResponse(String namespace, int patternId) {
```
markscott-ms: **reuse**: `patternNotFoundResponse`/`patternLayoutNotFoundResponse` duplicate `architectureNotFoundResponse`/`layoutNotFoundResponse` with only the noun swapped, inconsistent with this same file's own `resourceType`-parameterization pattern used nearby for `invalidLayoutTargetResponse`.

A future change to the not-found message wording or sanitization logic must be made in four near-identical methods instead of one parameterized method; a fix applied to only the architecture pair silently leaves the pattern pair (or a future third resource type) with stale wording.

calm-hub/src/main/java/org/finos/calm/resources/PatternLayoutResource.java
```diff
@@ -0,0 +1,146 @@
...
+    ) {
+        try {
+            String forPath = LayoutRequestParsing.parseForTarget(layoutJson);
+            String expectedPath = LayoutRequestParsing.canonicalPath(namespace, "patterns", patternId);
+            if (forPath != null && !forPath.equals(expectedPath)) {
+                return CalmResourceErrorResponses.invalidLayoutTargetResponse(forPath, expectedPath, "pattern");
+            }
+
+            // Checked on the write path only, not on get — mirrors LayoutResource#saveLayout.
+            if (!patternStore.patternExists(namespace, patternId)) {
```
markscott-ms: **efficiency**: `saveLayout` calls `patternStore.patternExists()` and then `patternLayoutStore.upsertLayout()`, each independently re-checking namespace existence — two DB round trips for the same fact on every save (mirrors a pre-existing inefficiency in `LayoutResource`, newly duplicated here rather than fixed).

calm-hub-ui/src/hub/hooks/useDefaultLayout.ts
```diff
@@ -1,121 +1,136 @@
...
+    // still resolving; once resolution for a supported type has settled
+    // without a match, this is null — not undefined — so Drawer's fallback is
+    // suppressed rather than silently keying off the raw slug (see
+    // viewportKeyOverride's doc).
     const viewportKey =
-        architectureId !== undefined
-            ? `${namespace}/${architectureId}`
-            : isArchitecture && !resolvingId
+        resolvedId !== undefined
+            ? `${namespace}/${calmType}/${resolvedId}`
```
markscott-ms: **reuse**: The namespace/calmType/id viewport-key template is hand-built independently here and in `Drawer.tsx`'s `computedViewportKey`, kept in sync only by a comment rather than a shared function.

A future change to the key shape (e.g. adding a version component) edited in one file but not the other would silently reintroduce the architecture/pattern id collision this PR set out to fix.

calm-hub/mongo/init-mongo.js
jpgough-ms: Resolved — #2971 merged first (as agreed), so this branch's `MongoPatternLayoutIndexStep` renumbers to `fromVersion() == 12` and `LATEST_SCHEMA_VERSION` bumps to `13`. Rebased onto the merged main and pushed.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoPatternLayoutIndexStep.java
jpgough-ms: Resolved — #2971 merged first, so this step renumbers to `fromVersion() == 12` / schema `13` (see the javadoc, updated to record the coordination). Rebased and pushed.

calm-hub/src/main/java/org/finos/calm/resources/PatternLayoutResource.java
jpgough-ms: Fixed — the loop is broken: `PatternLayoutResource` keeps only its own reason (the `AuditRequestFilter` path-param mapping) and points *down* to `PatternLayoutStore` for the storage rationale; `PatternLayoutStore`'s interface javadoc now states that rationale directly (moved up from `MongoPatternLayoutStore`) instead of pointing back up to the resource.

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
jpgough-ms: Good catch, fixed. `Drawer.tsx` now derives `effectiveDefaultLayout`/`effectiveLayoutEpoch` alongside `viewportKey` — both collapse to `undefined` whenever `fileInstance` is set, for both the pattern and architecture branches. Added regression tests in `Drawer.test.tsx` (verified they fail against the pre-fix code) plus a new `PatternGraph.test.tsx` mirroring `ArchitectureGraph.test.tsx`'s precedence/gating coverage, since that logic had no direct test before.

calm-hub/src/main/java/org/finos/calm/resources/CalmResourceErrorResponses.java
jpgough-ms: Fixed — replaced the four hard-coded-noun methods with two `resourceType`-parameterized ones (`resourceNotFoundResponse`, `resourceLayoutNotFoundResponse`), matching `invalidLayoutTargetResponse`'s existing convention. Message text is byte-for-byte unchanged (capitalized via a small helper), so no assertions elsewhere needed to move.

calm-hub/src/main/java/org/finos/calm/resources/PatternLayoutResource.java
jpgough-ms: Fixed for both architecture and pattern. `LayoutStore`/`PatternLayoutStore#upsertLayout` now do the existence check themselves (querying the header collection directly) instead of the resource calling `ArchitectureStore#architectureExists`/`PatternStore#patternExists` first — one `requireNamespace` per save instead of two. `architectureExists`/`patternExists` had no other callers, so removed them from both interfaces and all four impls along with their now-dead tests.

calm-hub-ui/src/hub/hooks/useDefaultLayout.ts
jpgough-ms: Fixed — extracted `buildViewportKey(namespace, calmType, id)` in `node-position-service.tsx` as the single source of truth for the key shape, used by both `useDefaultLayout.ts` and `Drawer.tsx`'s fallback computation. `legacyStorageKeyFor`'s segment-count check now references the same `VIEWPORT_KEY_SEGMENTS` constant.

markscott-ms (approved):

PR #2971, opened 2026-08-10
fix(calm-hub): scope resource mapping identity by resource type

### LINKED ISSUE #2970: Calm Hub Key Collisions/Issues
### Steps to Reproduce:

1. Create a pattern using `calm hub push pattern repo.pattern.json -t major`
2. Create an architecture using `calm hub push architecture repo.archtiecture.json -t major`
3. The `repo.pattern.json` and `repo.architecture.json`. Because both use the repo path segment and end up incrementing and swapping the type in Calm Hub

You can see these results by running
```
calm hub list patterns 
calm hub list architectures
```

### Expected Result:

Resource type should be included in the identity of a customId. 

### Actual Result:

Patterns and architectures of the same name collide, even if they are of different types

### Environment:
Main

### Additional Context:
...add any other context about the problem here. If applicable, add screenshots to help explain...

jpgough-ms:
## Description

Fixes #2970. `calm hub push pattern repo.pattern.json` and `calm hub push architecture repo.architecture.json` collided in CALM Hub — both derive the customId `repo` from their `$id`, but `resource_mappings` was keyed by `(namespace, customId)` only, with `resourceType` stored on every document but ignored on every lookup.

- Second push's version-probe resolved the mapping by customId alone, returning the **pattern's** version list
- CLI computed a semver bump off that wrong version and POSTed; server looked up the mapping untyped again and wrote the **architecture document into the `patterns` collection** under the pattern's numeric id
- `hub list patterns` showed `repo`; `hub list architectures` did not, despite the push reporting `201 Created`
- Reads (`GET .../versions/{v}`) had the same untyped-lookup bug independent of any write — a wrong-type document could be served back with a rewritten `$id`

Fix: make `ResourceType` part of the mapping's identity everywhere it's looked up by customId.

- `ResourceMappingStore.getMapping` / `updateMappingNumericId` / `deleteMapping` now require a `ResourceType` parameter — no untyped overload, since a 3-field unique index makes an untyped lookup ambiguous
- Both Mongo and Nitrite stores updated; Nitrite's app-level duplicate check now scopes on `(namespace, resourceType, customId)`
- New `MongoResourceMappingIndexStep` migration (schema v11→12) widens the Mongo unique index from `(namespace, customId)` to `(namespace, resourceType, customId)` for existing deployments — editing `MongoIndexInitializationStep` alone only fixes fresh databases, since it runs once at schema version 0 (see `MongoLayoutIndexStep` for the same precedent)
- Service/resource layers thread the type through; the two `resource_mappings` conflict messages now name the resource type, so a mismatched CLI subcommand/file fails with a legible 409 instead of silently corrupting data
- `mongo/init-mongo.js` dev-seed index and `LATEST_SCHEMA_VERSION` updated to match

**Scope**: server-only (`calm-hub/`), no CLI changes. Traced the CLI's existing 404→`[]` handling (`calm-hub-client.ts:365-368`) against the fixed server — the version-probe now correctly 404s for a genuinely new resource, the CLI forces `1.0.0`, and the push proceeds correctly. The full repro (including the "increments and swaps type" symptom) resolves server-side.

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

- 2631 unit tests pass (`../mvnw test`), including new coverage in `TestMongoResourceMappingStoreShould`, `TestNitriteResourceMappingStoreShould`, and a new `TestMongoResourceMappingIndexStepShould`
- 526 integration tests pass (`../mvnw -P integration verify`, TestContainers Mongo), including new cross-type regression tests in `MongoMappingControllerIntegration` and `NitriteMappingControllerIntegration` that push the same customId as a pattern and an architecture, then assert each type's read returns the correct document and each type's list contains exactly one entry
- JaCoCo coverage gate met
- Ran `/code-review` locally (twice, fixing two findings in between: a missed `LATEST_SCHEMA_VERSION` bump per `calm-hub/AGENTS.md`, and two integration assertions that used `hasItem` instead of an exact-count check)

## Checklist

Known limitation, not blocking: if a deployment already hit the write-corruption failure mode described above, the stray wrong-type document isn't cleaned up by this migration — it becomes unreachable once the correct-type mapping becomes creatable. No backfill needed since `resourceType` is present on every existing `resource_mappings` document.

### DIFF
--- a/calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
+++ b/calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
@@ -173,10 +173,14 @@ public class MongoIndexInitializationStep implements SchemaMigrationStep {
                 .createIndex(new Document("domain", 1), uniqueIndex);
         LOG.info("Ensured unique index on controls.domain");
 
-        // Resource mappings — unique (namespace, customId) and reverse lookup index
+        // Resource mappings — unique (namespace, resourceType, customId) and reverse lookup
+        // index. resourceType is part of the key so the same customId can be reused across
+        // different resource types (e.g. a pattern and an architecture can both be "repo") —
+        // see MongoResourceMappingIndexStep for the migration that carries existing deployments
+        // to this shape.
         database.getCollection("resource_mappings")
-                .createIndex(new Document("namespace", 1).append("customId", 1), uniqueIndex);
-        LOG.info("Ensured unique index on resource_mappings.(namespace, customId)");
+                .createIndex(new Document("namespace", 1).append("resourceType", 1).append("customId", 1), uniqueIndex);
+        LOG.info("Ensured unique index on resource_mappings.(namespace, resourceType, customId)");
 
         database.getCollection("resource_mappings")
                 .createIndex(new Document("namespace", 1).append("resourceType", 1).append("numericId", 1));
--- a/calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
+++ b/calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
@@ -173,14 +173,10 @@ public class MongoIndexInitializationStep implements SchemaMigrationStep {
                 .createIndex(new Document("domain", 1), uniqueIndex);
         LOG.info("Ensured unique index on controls.domain");
 
-        // Resource mappings — unique (namespace, resourceType, customId) and reverse lookup
-        // index. resourceType is part of the key so the same customId can be reused across
-        // different resource types (e.g. a pattern and an architecture can both be "repo") —
-        // see MongoResourceMappingIndexStep for the migration that carries existing deployments
-        // to this shape.
+        // Resource mappings — unique (namespace, customId) and reverse lookup index
         database.getCollection("resource_mappings")
-                .createIndex(new Document("namespace", 1).append("resourceType", 1).append("customId", 1), uniqueIndex);
-        LOG.info("Ensured unique index on resource_mappings.(namespace, resourceType, customId)");
+                .createIndex(new Document("namespace", 1).append("customId", 1), uniqueIndex);
+        LOG.info("Ensured unique index on resource_mappings.(namespace, customId)");
 
         database.getCollection("resource_mappings")
                 .createIndex(new Document("namespace", 1).append("resourceType", 1).append("numericId", 1));
--- a/calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
+++ b/calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
@@ -84,7 +84,7 @@ public class MappingControllerService {
         try {
             ResourceMapping existing;
             try {
-                existing = mappingStore.getMapping(namespace, name);
+                existing = mappingStore.getMapping(namespace, resourceType, name);
             } catch (MappingNotFoundException ignored) {
                 existing = null;
             }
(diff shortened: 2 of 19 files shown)


markscott-ms:
Requirements make sense ... now checking the fix...

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
```diff
@@ -173,10 +173,14 @@ private void createUniqueIndexes() {
                 .createIndex(new Document("domain", 1), uniqueIndex);
         LOG.info("Ensured unique index on controls.domain");
 
-        // Resource mappings — unique (namespace, customId) and reverse lookup index
+        // Resource mappings — unique (namespace, resourceType, customId) and reverse lookup
+        // index. resourceType is part of the key so the same customId can be reused across
+        // different resource types (e.g. a pattern and an architecture can both be "repo") —
```
markscott-ms: This file should not be changed - and your comment in MongoResourceMappingIndexStep even says why.

We should never modify the behaviour of any already merged migration/step class because they must work in sequence, on anything from an empty CalmHub to a completely up to date one.

markscott-ms (changes requested):

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
```diff
@@ -462,7 +462,8 @@ private Response createNewResource(String namespace, ResourceType resourceType,
             return CalmResourceErrorResponses.invalidNamespaceResponse(namespace);
         } catch (DuplicateMappingException e) {
             return Response.status(Response.Status.CONFLICT)
-                    .entity("Resource name already exists in namespace: " + STRICT_SANITIZATION_POLICY.sanitize(name)).build();
+                    .entity(resourceType.name().toLowerCase() + " '" + STRICT_SANITIZATION_POLICY.sanitize(name)
+                            + "' already exists in namespace '" + STRICT_SANITIZATION_POLICY.sanitize(namespace) + "'").build();
```
markscott-ms: **Cleanup:** these two new 409 conflict messages build their entity string inline (`resourceType.name().toLowerCase() + ... + STRICT_SANITIZATION_POLICY.sanitize(...)`) instead of going through `CalmResourceErrorResponses`, the helper this file already uses for its other error responses (`invalidNamespaceResponse`/`invalidDomainResponse`, used a few lines above). Worth centralizing there too so the sanitize-and-concatenate pattern isn't duplicated for the next person to copy (same issue applies to the `addNewVersion` conflict message a few lines down).

willosborne (commented):
Looks good, other than Mark's suggested changes.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoIndexInitializationStep.java
jpgough-ms: Reverted `MongoIndexInitializationStep` back to its pre-PR state in a9b9bfd2 — the widened `(namespace, resourceType, customId)` index is now established solely by `MongoResourceMappingIndexStep`, which already tolerantly drops the narrower index before creating the replacement. Verified with `../mvnw -P integration verify` (all 526 integration tests pass, including the cross-type regression tests), so a fresh database and an existing deployment still converge on the same index shape.

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
jpgough-ms: Centralized both messages into `CalmResourceErrorResponses` (new `resourceAlreadyExistsResponse` / `versionAlreadyExistsResponse` helpers, matching the existing `namespaceNotEmptyResponse`/`domainNotEmptyResponse` style) in ac41f32f. Response bodies are unchanged — added exact-body assertions in `TestCalmErrorResponsesShould` to pin them.

markscott-ms (approved):

PR #2973, opened 2026-08-10
fix(calm-hub): migrate UserAccessStore namespace checks to requireNamespace

markscott-ms:
## Description
`MongoUserAccessStore` and `NitriteUserAccessStore` still contained the `if (!namespaceStore.namespaceExists(x)) throw NamespaceNotFoundException` shape that PR #2959 eliminated everywhere else, replacing it with `NamespaceStore#requireNamespace`. This was flagged in a review comment on #2959 as missed. This PR applies the same drop-in replacement to the remaining 4 call sites in each store, guarded by the existing `GLOBAL_ACCESS` special case where it was already present.

## Type of Change
- [ ] 📚 Documentation update

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

Ran `../mvnw clean verify -Ddependency-check.skip=true` — 2620 tests, 0 failures, 0 errors, JaCoCo coverage passed.

## Checklist

### DIFF
--- a/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoUserAccessStore.java
+++ b/calm-hub/src/main/java/org/finos/calm/store/mongo/MongoUserAccessStore.java
@@ -44,8 +44,8 @@ public class MongoUserAccessStore implements UserAccessStore {
             throws NamespaceNotFoundException {
 
         log.info("User-access details: {}", userAccess);
-        if (!GLOBAL_ACCESS.equals(userAccess.getNamespace()) && !namespaceStore.namespaceExists(userAccess.getNamespace())) {
-            throw new NamespaceNotFoundException();
+        if (!GLOBAL_ACCESS.equals(userAccess.getNamespace())) {
+            namespaceStore.requireNamespace(userAccess.getNamespace());
         }
 
         return createUserAccess(userAccess, "namespace", userAccess.getNamespace());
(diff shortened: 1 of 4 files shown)


markscott-ms:
Pre-empting a review comment about `lenient()`...

The Nitrite test uses `@ExtendWith(MockitoExtension.class)` with Mockito's strict stubs mode (the default for that extension), which fails the test if a stub set up in `@BeforeEach` is never invoked by a given test method. `UserAccessStore` has domain-scoped methods (`createUserAccessForDomain`, `getUserAccessForDomain`, `deleteUserAccessForDomain`, etc.) that never touch `namespaceStore.requireNamespace(...)` at all (think grant/revoke for global admin permission) — so for those test methods the blanket `doCallRealMethod().when(namespaceStore).requireNamespace(...)` stub in `setup()` is dead code, and strict-stubs flags it as `UnnecessaryStubbingException`. `lenient()` opts that one stub out of the unused-stub check.

The Mongo test class is `@QuarkusTest` with `@InjectMock`, which doesn't use Mockito's strict-stubs JUnit extension — unused stubs there are simply never enforced, so the same "not every test calls it" situation never throws.

jpgough-ms:
Pre-existing bug, not introduced by this PR: `getUserAccessForNamespaceAndId` skips the `GLOBAL_ACCESS` guard its sibling methods (`create`/`get`/`delete`) apply.

- `MongoUserAccessStore.java:137` and `NitriteUserAccessStore.java:141` call `namespaceStore.requireNamespace(namespace)` unconditionally.
- Repro: `POST .../namespaces/GLOBAL/user-access` succeeds (create special-cases `GLOBAL_ACCESS`). `GET .../namespaces/GLOBAL/user-access/{id}` on that same record then 404s, since `GLOBAL` is never a persisted `NamespaceInfo`.
- Confirmed against `main` — the identical unconditional check existed before this refactor, just under a different name.

Not blocking here since it's carried forward rather than caused by this change. Worth opening a follow-up issue to add the same guard to both stores.

jpgough-ms (approved):

Issue #2953, opened 2026-08-07
Layout Persistence Parity Between CALM Hub and VS Code Plugin


byrash:
## Feature Proposal

**Background**

  PR [#2942 ](https://github.com/finos/architecture-as-code/pull/2942)("Layout Persistence in CALM Hub") was merged to main on 2026-08-06. It introduced:

  - GET/PUT endpoints at .../namespaces/{namespace}/architectures/{architectureId}/layout
  - Flat storage model: one MongoDB document per (namespace, architectureId) in a layouts collection (schema version 9→10)
  - Frontend useDefaultLayout hook with slug→numeric resolution, save/reset actions, and a three-tier precedence (local scratch > server default > auto-layout)
  - LayoutService, MongoLayoutStore, NitriteLayoutStore, LayoutStoreProducer, MongoLayoutIndexStep
  - Architecture-existence guard on PUT, audit trail integration, namespace-deletion guard
  - ADR 0005 documenting why the layout shape is Hub-internal, not a CALM schema

  This feature request builds directly on that foundation to close the remaining parity gap with the VS Code plugin's layout handling.

  ---
  **Problem Statement**

  CALM Hub UI and the VS Code plugin both render architecture diagrams using the same ReactFlow components, but their layout persistence formats diverge — the Hub stores only node positions (x, y) while the VS Code
  plugin stores full geometry (x, y, w, h). This means:

  - A layout saved in CALM Hub cannot reproduce the exact rendering the VS Code plugin produces (and vice versa)
  - Without width/height, ReactFlow auto-sizes nodes based on content, causing layout drift across clients, browser widths, and font rendering differences
  - Container (group) nodes cannot be faithfully restored without explicit dimensions — reflowContainersToFitChildren guesses boundaries rather than applying saved ones

  ---
  1. Enhanced Layout Persistence (include width & height)

  Current Model (Hub — as shipped in PR #2942)

  {
      "for": "/api/calm/namespaces/finos/architectures/5",
      "pins": [
          { "unique-id": "node-a", "position": { "x": 93, "y": 175 } },
          { "unique-id": "node-b", "position": { "x": 72, "y": 339 } }
      ]
  }

  - Array of objects, scanned by unique-id to find a node → O(n)
  - No dimensions — position-only, rendering is non-deterministic
  - Stored as a parsed BSON Document inside the layouts collection
  - Conversion layer: pinsToStoredPositions / storedPositionsToPins in calm-hub-ui/src/model/layout.ts

[...]

PR #2987, opened 2026-08-13
Feat/issue 2953 layout parity between calm hub and code plugin

byrash:
## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [ ] All existing tests pass

## Checklist


byrash:
Addresses https://github.com/finos/architecture-as-code/issues/2953 

This supports both in Document & Layout in Calm Hub.

byrash:
@markscott-ms  @jpgough-ms  Please let us know what you think. Thank You.

calm-hub-ui/src/visualizer/components/reactflow/ArchitectureGraph.tsx
```diff
@@ -173,11 +173,13 @@ export function ArchitectureGraph({
...
-        // never silently discard a user's in-progress work — which in turn wins
-        // over the parsed auto-layout when neither is present.
+        // Precedence: a provided default layout (from document _layout or
+        // server-saved) wins over stale localStorage scratch, so the document's
+        // authoritative positions are always applied on page load. Scratch only
+        // applies when there is no default — i.e. a fresh architecture with
+        // neither a document _layout nor a server-saved layout.
         const localPositions = viewportKey ? loadStoredNodePositions(viewportKey) : null;
-        const effectivePositions = localPositions ?? defaultLayout ?? null;
+        const effectivePositions = defaultLayout ?? localPositions ?? null;
```
rocketstack-matt: This flips precedence so the default layout now overwrites an unsaved local drag whenever this effect re-runs (data reload, source toggle, any `layoutEpoch` bump) — silently discarding in-progress work. That's what the comment above used to explicitly guard against, and it also contradicts issue #2953's own spec ("Local scratch ... always wins").

`setLayoutSource` already clears scratch before switching sources, so the toggle feature doesn't need this flip to work — reverting the order fixes the regression with no loss of functionality:
```suggestion
        const effectivePositions = localPositions ?? defaultLayout ?? null;
```
Same issue in `PatternGraph.tsx:131`.

calm-hub-ui/src/visualizer/services/node-position-service.tsx
```diff
@@ -123,22 +131,42 @@ function parseStoredPositions(data: string): StoredNodePosition[] | null {
...
+        }
+        return updated;
     });
 
-    return reflowContainersToFitChildren(merged);
+    // Only skip reflow when every container's dimensions came from the stored
+    // positions (not from the node's own pre-existing width/height). This ensures
+    // position-only restores (legacy data, localStorage scratch) still reflow.
+    const containerIds = new Set(
+        merged.filter((n) => n.type === 'group').map((n) => n.id)
```
rocketstack-matt: This only recognizes `type === 'group'` as a container, but `PatternGraph.tsx` also treats `'decisionGroup'` nodes as containers (`GROUP_NODE_TYPES = ['group', 'decisionGroup']`, and `patternTransformer.ts` wires children to a decision group via `parentId` the same way). For pattern diagrams, a `decisionGroup` container missing explicit w/h in the stored positions won't be caught by `allContainersDimensioned`, so the needed reflow gets skipped and that container renders the wrong size.

Consider deriving `containerIds` the same way `reflowContainersToFitChildren` does — any node id referenced by another node's `parentId` — rather than checking a hardcoded type string, so this doesn't need to track every container type by name.

calm-hub/src/main/resources/application.properties
```diff
@@ -11,6 +11,9 @@ calm.standalone.database-name=calmSchemas
 calm.standalone.username=admin
 calm.standalone.password=admin
 calm.standalone.init-script-path=mongo/init-mongo.js
+# Seed a demo namespace/architecture/layout on first startup when the database is empty.
+# Set to false to start standalone mode with a completely empty database.
+calm.standalone.seed-demo-data=true
```
rocketstack-matt: Defaulting this to `true` means any real standalone/Nitrite Hub deployment with an empty database gets an unsolicited "demo" namespace with fabricated architecture data on first boot, not just this repo's own local dev flow — unrelated to the layout-parity issue this PR addresses. The integration test resource already has to explicitly set this to `false`, which suggests the production default should probably be `false` too (opt-in), or this seeder should be split into its own PR.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoLayoutFormatMigrationStep.java
```diff
@@ -0,0 +1,108 @@
...
+ * { "for": "...", "nodes": { "node-a": { "x": 0, "y": 0 } } }
+ * }</pre>
+ *
+ * <p>Idempotent: documents that already have a {@code nodes} map (or no {@code pins} field) are
+ * skipped. Positions without width/height are migrated as-is — the frontend handles missing
+ * dimensions gracefully by falling back to reflow.</p>
+ */
+@LookupIfProperty(name = "calm.database.mode", stringValue = "mongo", lookupIfMissing = true)
+@ApplicationScoped
+public class MongoLayoutFormatMigrationStep implements SchemaMigrationStep {
```
rocketstack-matt, line 36: Every other schema-affecting migration in this package ships a Mongo step plus a mutually-exclusive Nitrite twin at the same `fromVersion()` (e.g. `MongoControlVersionSplitStep`/`NitriteControlVersionSplitStep`, `MongoAdrTitleBackfillStep`/`NitriteAdrTitleBackfillStep`). This PR only adds the Mongo side — existing standalone/Nitrite layouts stored in the legacy `pins` format never get migrated to `nodes`. The frontend's `apiResponseToStoredPositions` fallback masks this (no user-visible break), but it's inconsistent with the established pattern. Worth adding a `NitriteLayoutFormatMigrationStep` twin, or a comment noting why one isn't needed.

rocketstack-matt (changes requested):
Two blocking issues below (unsaved-edit data loss, missed container type in the reflow-skip check), plus two non-blocking design/scope notes.

Separately: issue #2953 is still open with an unresolved objection from @markscott-ms about persisting layout inside the architecture document at all, and no explicit design sign-off — worth settling there before this merges, independent of the code-level points below.

calm-hub/src/main/resources/application.properties
byrash: Our assumption was standalone or nitrite DB is more for Local development purposes. For Production use cases backend storage will be either Mongo DB or GitHub, so we ensured 

MongoDB is NOT affected. Based on below stuff.

  The seeder is guarded by two gates:
  1. @LookupIfProperty(name = "calm.database.mode", stringValue = "standalone") (line 22) — CDI won't even instantiate this bean unless the mode is standalone
  2. if (!"standalone".equals(databaseMode)) return (line 44) — belt-and-suspenders runtime check

  So MongoDB deployments (calm.database.mode=mongo, which is the default) will never run this seeder regardless of the seed-demo-data property value.
  
  @rocketstack-matt  let me know if nitrite DB or standalone mode is also for production needs, we can flip the property as suggested. Thank You

byrash:
> Two blocking issues below (unsaved-edit data loss, missed container type in the reflow-skip check), plus two non-blocking design/scope notes.
> 
> Separately: issue #2953 is still open with an unresolved objection from @markscott-ms about persisting layout inside the architecture document at all, and no explicit design sign-off — worth settling there before this merges, independent of the code-level points below.

Hello @rocketstack-matt , Appreciate your feedback. All the mentioned issues has now been resolved and change should be in PR soon.

Regarding layout, Our priority is to have uniform rendering of CALM Diagrams in all tools in CALM suite being it VS Code or Calm Hub or Calm docify rendered markdown files.  As you are aware we have choose Git as the backend store which acts as the single source of truth for all the tools in CALM suite and works consistently. Hence having layout as part of calm file itself was a design decision chosen so that it renders in VS Code plugin, reading layout form calm file itself. We are happy for it to be external as long as its works consistent across all the tools. Until we have a solution that works meeting those needs, we have added this as a tactical solution to be able to support the ask. 

@markscott-ms  Can we add it as an item for our next WG item to discuss and brainstorm further on this matter, if required. 

Many Thanks! 

FYI: @jimthompson5802

calm-hub/src/main/resources/application.properties
rocketstack-matt: Standalone is also the backend for the read-only image we publish to hub.calm.finos.org. `Dockerfile.readonly-static` bakes in `QUARKUS_PROFILE=standalone`, and `calm-hub/ec2/docker-compose.yml` runs `latest-read-only-native` on the box behind that domain. 

Something we could consider once we do have a GitHub backed version is whether we want to use that to run the hosted instance which would enable us to manage a separate repository for all the stored artifacts - what do you think about that @jpgough-ms?

byrash: made required changes, Thank You

rocketstack-matt:
@byrash just flagging the failing build - happy to take another look once resolved.

byrash:
> @byrash just flagging the failing build - happy to take another look once resolved.

Build is all good now. Thanks!

calm-hub-ui/src/visualizer/components/reactflow/ArchitectureGraph.tsx
rocketstack-matt: Confirmed fixed — precedence reverted to `localPositions ?? defaultLayout ?? null` in both ArchitectureGraph.tsx:182 and PatternGraph.tsx:131, with new tests covering local-scratch-wins-over-default and the epoch-bump-after-clear sequence.

calm-hub-ui/src/visualizer/services/node-position-service.tsx
rocketstack-matt: Confirmed fixed — containerIds is now derived generically from any node's parentId rather than a hardcoded type check, so decisionGroup containers are covered the same as group.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoLayoutFormatMigrationStep.java
rocketstack-matt: Confirmed fixed — NitriteLayoutFormatMigrationStep added as the mutually-exclusive twin, with its own test suite.

rocketstack-matt (approved):
Both blocking items and both non-blocking items from my prior review are resolved and tested.

byrash:
Hello @markscott-ms can we please merge this PR ? Many Thanks!

markscott-ms (approved):

Issue #2929, opened 2026-08-03
Accessibility: WCAG 2.2 AA audit findings for CALM Hub UI


eddie-knight:
## Bug Report

WCAG 2.2 AA audit of CALM Hub UI. Confirmed failures at **Level A and AA**. No a11y linting or CI gate currently exists.

### What was run

- `eslint-plugin-jsx-a11y` 6.10.2 (`flatConfigs.recommended`) over `calm-hub-ui/src/**/*.tsx`, tests excluded.
- Playwright 1.62 (Chromium, 1440×900) + `axe-core` 4.12.1 against the dev server, tags `wcag2a, wcag2aa, wcag21a, wcag21aa, wcag22aa`, routes `#/`, `#/namespace/finos`, `#/domain/security`, `#/admin/entitlements`. API stubbed at the network layer — no backend needed.
- Both themes (`data-theme=light|dark`).

### Findings

**1. Contrast below AA — both themes (1.4.3, AA)** — same three elements fail in each theme; AA requires 4.5:1.

| Theme | Ratio | Colours | Element |
|---|---|---|---|
| light | **2.40:1** | `#9aa6b8` on `#fcfcfd` | `.text-[10px].uppercase.tracking-[0.1em]` section label |
| light | **2.46:1** | `#9aa6b8` on `#ffffff` | explore-rail heading (`NAMESPACES` / `CONTROL DOMAINS`) |
| light | **4.23:1** | `#64748b` on `#eef2f7` | count badge on namespace links |
| dark | **3.07:1** | `#64748b` on `#1e293b` | same section label |
| dark | **3.75:1** | `#64748b` on `#0f172a` | same explore-rail heading |
| dark | **3.38:1** | `#8a94a6` on `#334155` | same count badge |

**2. Link distinguished by colour alone — dark theme only (1.4.1, A)** — `link-in-text-block` on `.hover:underline`. Not reproducible in light theme.

**3. Keyboard-unreachable interactive elements (2.1.1, A)** — 17 static hits (9× `no-static-element-interactions`, 8× `click-events-have-key-events`):

| File | Line |
|---|---|
| `hub/components/value-table/ValueTable.tsx` | 24 |
| `visualizer/components/reactflow/CustomNode.tsx` | 159, 324 |
| `visualizer/components/reactflow/controls-panel/ControlCard.tsx` | 9, 94 |
| `visualizer/components/reactflow/controls-panel/ControlRequirementItem.tsx` | 146 |
[...]

PR #2989, opened 2026-08-14
fix(calm-hub-ui): meet WCAG AA contrast for redesign text tokens

opensource-joe:
Addresses the contrast findings in #2929 (finding 1, and the dark-theme half of finding 2's surface set). The keyboard, combobox and `autoFocus` findings are untouched, to keep this reviewable.

## The failures

Two redesign text tokens sit below the 4.5:1 that WCAG 2.2 SC 1.4.3 requires for normal text, in both themes. The text they colour is 10px (rail section labels) and 11px (count badges), so the 18.66px bold / 24px large-text exemption does not apply.

| Theme | Token | Value | Worst surface | Ratio |
|---|---|---|---|---|
| light | `--calm-redesign-faint-alt` | `#9AA6B8` | `--calm-redesign-surface` `#F8FAFC` | **2.36:1** |
| light | `--calm-redesign-muted-alt` | `#64748B` | `--calm-redesign-badge-bg` `#EEF2F7` | **4.23:1** |
| dark | `--calm-redesign-faint-alt` | `#64748B` | `--calm-redesign-surface` `#1E293B` | **3.07:1** |
| dark | `--calm-redesign-muted-alt` | `#8A94A6` | `--calm-redesign-badge-bg` `#334155` | **3.39:1** |

I reproduced all six ratios in the issue exactly. Two further pairings turned up that the audit did not list, both from the same two tokens: `faint-alt` on `--calm-redesign-surface` (`#F8FAFC`) at 2.36:1, and `muted-alt` on `--calm-redesign-badge-bg-faint` (`#F4F6F9`) at 4.40:1. Nine assertions fail in total against the current palette.

## The fix

Each token moves to the next step that clears AA against every surface it is actually used on, keeping the same hue:

| | before | after | worst ratio after |
|---|---|---|---|
| light `faint-alt` | `#9AA6B8` | `#64748B` | 4.55:1 |
| light `muted-alt` | `#64748B` | `#5A6678` | 5.18:1 |
| dark `faint-alt` | `#64748B` | `#8A94A6` | 4.78:1 |
| dark `muted-alt` | `#8A94A6` | `#A4ACBA` | 4.53:1 |

Three of the four reuse a value already in the palette. Only the dark badge text needed a new one, and `#A4ACBA` is the minimum that clears 4.5:1 on `#334155` rather than a jump to slate-300 `#CBD5E1`, which would have brightened all dark-mode meta text noticeably.

## What this costs, which is your call not mine

The five-step text scale does not survive AA in light mode, and you should decide what to do about that rather than inherit my answer.

Above white, 4.5:1 permits nothing lighter than roughly slate-500. `faint-alt` therefore lands on `#64748B`, which is what `muted-alt` used to be, and `muted-alt` lands on `#5A6678`, which is what `muted` already is. So `muted` and `muted-alt` now hold the same value in light mode, and the intended `muted > mutedAlt > faintAlt` gradation is two steps rather than three.

That is inherent to the palette rather than to these hexes: the scale has more faint steps than AA allows for small text on a near-white background. Options as I see them, and I am happy to redo this as whichever you prefer:

1. Take this as-is and accept the collapse, then re-space the scale separately.
2. Re-space now, so the remaining steps are visually distinct at AA.
3. Keep the faint tier light and stop using it for text, giving those labels a different role.

One related note: `--calm-redesign-faint` fails on both themes too (3.06:1 light, 3.07:1 dark), but `colors.redesign.faint` has no consumers anywhere in `src`, so there is no live failure and I left it alone rather than widen the diff. It is a trap for whoever adopts it first. Say the word and I will include it.

## The test

The issue notes there is no a11y lint or CI gate. `contrast.test.ts` is a narrow first piece of that: it parses the two theme blocks out of `theme.css`, resolves each token pair that occurs in the UI, and asserts the WCAG ratio. No browser or axe run needed, so it costs nothing in CI.

The pairing is the point. `muted-alt` reads fine on the page background and still failed at 4.23:1 inside a badge, which is exactly the case a per-token review misses.

Two implementation notes for review. It reads `theme.css` off disk rather than via `import './theme.css?raw'`, because Vitest stubs the CSS transform by default and the raw import resolves to an empty string, which would have made every assertion pass against nothing. And it asserts both tokens are defined before comparing, so a rename cannot turn the test green by making both sides `undefined`.

## Verification

Run in a clean `node:26-bookworm` container:

- `npm run test --workspace calm-hub-ui`: 116 files, 1419 tests, all passing (1395 before, plus the 24 new assertions).
- `npm run lint --workspace calm-hub-ui` (eslint + stylelint): passing.
- The new test is not vacuous. Reverting `theme.css` to the current values turns it red with **9** failures, each naming the pair and the measured ratio, and every one of the issue's six numbers appears among them.

One thing worth flagging for anyone running the suite locally: `DiagramSection.test.tsx` fails to resolve `@finos/calm-models/diff` unless `npm run build --workspace calm-models` has been run first. That is pre-existing on `main` and unrelated to this change. I confirmed it by stashing my changes and reproducing it on a clean tree.

What I have not done is view these colours in a running browser. The ratios are computed from the token values, which is exactly what axe measures, but the judgement about whether the new light-mode greys still look right is yours.

### DIFF
--- a/calm-hub-ui/src/theme/contrast.test.ts
+++ b/calm-hub-ui/src/theme/contrast.test.ts
@@ -1,0 +1,122 @@
+/**
+ * Colour-contrast guard for the redesign text tokens.
+ *
+ * The redesign palette is consumed as CSS custom properties, so a contrast
+ * failure is a property of a token pair rather than of any one component. That
+ * makes it checkable without a browser: read the two theme blocks out of
+ * `theme.css`, resolve each pair, and compute the WCAG ratio.
+ *
+ * Every pair below is one that actually occurs in the UI. The threshold is 4.5:1
+ * (WCAG 2.2 SC 1.4.3, AA) because all of this text is small: the rail section
+ * labels are 10px and the count badges 11px, nowhere near the 18.66px bold /
+ * 24px large-text exemption.
+ *
+ * The point is the pairing. A token that reads comfortably on the page
+ * background can still fail on a badge, which is exactly how two of these were
+ * missed.
+ */
+import { existsSync, readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Reads `theme.css` off disk.
+ *
+ * Not `import './theme.css?raw'`: Vitest stubs the CSS transform by default, so
+ * a raw import resolves to an empty string and every assertion below would pass
+ * against nothing. Both candidate roots are tried because the suite runs from
+ * the workspace directory but can be invoked from the repository root.
+ */
+function readThemeCss(): string {
+    const candidates = [join(process.cwd(), 'src/theme/theme.css'), join(process.cwd(), 'calm-hub-ui/src/theme/theme.css')];
+    const found = candidates.find((candidate) => existsSync(candidate));
+    if (!found) {
+        throw new Error(`Could not locate theme.css. Looked in: ${candidates.join(', ')}`);
+    }
+    return readFileSync(found, 'utf8');
+}
+
+const THEME_CSS = readThemeCss();
+
+/** WCAG 2.2 SC 1.4.3 minimum for text below the large-text threshold. */
+const AA_NORMAL_TEXT = 4.5;
+
+/**
+ * The dark block's selector. Matched by pattern rather than as a literal because
+ * this file is read through Vite's `?raw`, which returns the CSS after PostCSS
+ * has normalised it, and PostCSS is free to change the attribute quoting.
+ */
+const DARK_SELECTOR = /:root\[data-theme=['"]?dark['"]?\]/;
+
+/** Token pairs that occur in the UI, as `[text, surface]`. */
+const PAIRS: readonly [string, string][] = [
+    // Rail section labels: `NAMESPACES`, `CONTROL DOMAINS`.
+    ['calm-redesign-faint-alt', 'calm-redesign-surface-alt'],
+    ['calm-redesign-faint-alt', 'calm-redesign-surface'],
+    ['calm-redesign-faint-alt', 'calm-bg-base'],
+    // Count badges on the namespace links and the segmented type tabs.
+    ['calm-redesign-muted-alt', 'calm-redesign-badge-bg'],
+    ['calm-redesign-muted-alt', 'calm-redesign-badge-bg-faint'],
+    ['calm-redesign-muted-alt', 'calm-bg-base'],
+    // Body copy and headings on the page and the diagram stage.
+    ['calm-redesign-muted', 'calm-bg-base'],
+    ['calm-redesign-body', 'calm-bg-base'],
+    ['calm-redesign-body', 'calm-redesign-surface'],
+    ['calm-redesign-body-alt', 'calm-bg-base'],
+    ['calm-redesign-ink', 'calm-bg-base'],
+    ['calm-redesign-ink', 'calm-redesign-canvas'],
+];
+
+/**
+ * Pulls the six-digit hex custom properties out of one theme block.
+ *
+ * The light block is everything before the dark selector, which keeps this
+ * honest about cascade order: a token redeclared later in the same block wins,
+ * and so does the last match here.
+ */
+function tokensFor(theme: 'light' | 'dark'): Record<string, string> {
+    const darkMatch = DARK_SELECTOR.exec(THEME_CSS);
+    if (!darkMatch) {
+        throw new Error(`theme.css no longer contains a ${DARK_SELECTOR} block`);
+    }
+    const darkStart = darkMatch.index;
+
+    const block = theme === 'dark' ? THEME_CSS.slice(darkStart) : THEME_CSS.slice(0, darkStart);
+    const tokens: Record<string, string> = {};
+    for (const match of block.matchAll(/--([\w-]+):\s*(#[0-9A-Fa-f]{6})\s*;/g)) {
+        tokens[match[1]] = match[2];
+    }
+    return tokens;
+}
+
+/** Relative luminance, per WCAG 2.x. */
+function luminance(hex: string): number {
+    const channels = [1, 3, 5].map((offset) => {
+        const value = parseInt(hex.slice(offset, offset + 2), 16) / 255;
+        return value <= 0.04045 ? value / 12.92 : ((value + 0.055) / 1.055) ** 2.4;
+    });
+    return 0.2126 * channels[0] + 0.7152 * channels[1] + 0.0722 * channels[2];
+}
+
+function contrastRatio(foreground: string, background: string): number {
+    const [lighter, darker] = [luminance(foreground), luminance(background)].sort((a, b) => b - a);
+    return (lighter + 0.05) / (darker + 0.05);
+}
+
+describe.each(['light', 'dark'] as const)('%s theme contrast', (theme) => {
+    const tokens = tokensFor(theme);
+
+    it.each(PAIRS)('%s on %s meets WCAG AA for normal text', (text, surface) => {
+        const foreground = tokens[text];
+        const background = tokens[surface];
+
+        // A renamed or deleted token would otherwise make this pass vacuously.
+        expect(foreground, `--${text} is not defined in the ${theme} theme`).toBeDefined();
+        expect(background, `--${surface} is not defined in the ${theme} theme`).toBeDefined();
+
+        const ratio = contrastRatio(foreground, background);
+        expect(ratio, `--${text} (${foreground}) on --${surface} (${background}) is ${ratio.toFixed(2)}:1`).toBeGreaterThanOrEqual(
+            AA_NORMAL_TEXT
+        );
+    });
+});
--- a/calm-hub-ui/src/theme/contrast.test.ts
+++ b/calm-hub-ui/src/theme/contrast.test.ts
@@ -1,3 +1,10 @@
+// @vitest-environment node
+//
+// Runs under node, not jsdom, for the same reason as the sibling
+// `monaco-theme.test.ts`: this suite reads `theme.css` off disk through
+// `import.meta.url`, and under jsdom that is not a `file:` URL, so
+// `fileURLToPath` throws before a single pair is checked. Nothing here touches
+// the DOM.
 /**
  * Colour-contrast guard for the redesign text tokens.
  *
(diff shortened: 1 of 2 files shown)


opensource-joe:
@markscott-ms Would you be able to approve the workflow runs on this PR when you get a chance?

EasyCLA cleared this morning, but as a first-time contributor here my workflows need a maintainer to release them, so the 11 build and test jobs are all sitting at `action_required` and none of them have actually run yet. Worth flagging because the PR reads as green at a glance: the two checks that do report, EasyCLA and `label`, both pass, so nothing looks blocked unless you check the runs directly.

Nothing is needed from me for this, and I am not asking for a review yet, just for CI to be allowed to run so any failures are mine to fix before I take up reviewer time.

For context on what is waiting, this addresses the contrast half of #2929. It fixes four redesign text tokens in `calm-hub-ui` that fell below WCAG 4.5:1, and adds `src/theme/contrast.test.ts`, which parses `theme.css` and asserts the ratios for the token pairs the UI actually uses, so the palette cannot regress silently. Locally 1419 tests pass, and reverting the palette turns that guard red with 9 failures, so it is not a vacuous test.

Happy to rebase if the branch has gone stale in the meantime.

opensource-joe:
Following up on my own ask above, as a status update rather than a second request: thank you @markscott-ms, the approval came through and CI has now run.

All 12 workflows executed and all 14 checks pass, including the full build, test and lint matrix, plus EasyCLA. The branch is mergeable with no conflicts.

So the thing I said I wanted before taking up reviewer time now exists. This is `REVIEW_REQUIRED` with no reviewer assigned, and it is a small change whenever someone has a moment: a contrast fix to four redesign text tokens in `calm-hub-ui/src/theme/theme.css`, plus `contrast.test.ts`, a guard test that computes the WCAG contrast ratios so a future token edit that drops below AA fails the build rather than shipping quietly.

calm-hub-ui/src/theme/theme.css
```diff
@@ -82,9 +82,9 @@
     --calm-redesign-body: #41506A;
     --calm-redesign-body-alt: #475569;
     --calm-redesign-muted: #5A6678;
-    --calm-redesign-muted-alt: #64748B;
+    --calm-redesign-muted-alt: #5A6678; /* was #64748B, 4.23:1 on the badge bg */
```
YoofiTT96: Any reason why the previous values remain as comments in the new diff?

YoofiTT96 (commented):
Besides the comment, only wondering if there is any visible change in the default case to be aware of? If so can a screenshot be added

opensource-joe:
Both fixed, thanks.

The commented-out values are gone. They were carrying the evidence for each new hex, but that evidence belongs in contrast.test.ts, which asserts the ratios and fails if anyone regresses them. The CSS diff is now a plain four-line value swap.

On the visible change: yes, and it is not behind a flag. Both tokens are live. faint-alt colours the rail section labels, the intro browse section titles and the search-results meta line. muted-alt colours the count badges, the namespace and domain breadcrumbs, and the item-card meta. Light mode gets darker, dark mode gets lighter.

Below is the explore rail at /namespace/finos in each theme, before and after, shown at 2x. The catalogue data is mocked, so theme.css is the only difference between the two halves of each image.

compare-light

compare-dark

One thing still needs your call, and it is in the PR description rather than here: light-mode muted-alt now equals muted, because above a near-white surface AA leaves no room for a distinct step. I am happy to re-space the scale instead if you would rather not collapse it.

markscott-ms:
@opensource-joe could you re-space the light scale?

The images were very helpful and are a great visual improvement.

opensource-joe:
Re-spaced, thanks.

The collapse came from moving `muted-alt` onto `#5A6678`, which `muted` already
held. Rather than nudge one token, I spread all three across the range AA
leaves, measuring each on the darkest surface it is actually used on:

| token | before | after | worst ratio |
|---|---|---|---|
| `muted` | `#5A6678` | `#515F75` | 6.47:1 on `#FFFFFF` |
| `muted-alt` | `#5A6678` | `#5A697F` | 4.97:1 on `#EEF2F7` |
| `faint-alt` | `#64748B` | unchanged | 4.55:1 on `#F8FAFC` |

Steps go from 0.0 and 5.5 in CIE L* to 4.0 and 4.4. `muted` darkens slightly,
which is the cost: it colours body copy and headings, so it is the one change
that shows up outside the rail and the badges.

Light text scale compared across main, this PR as it stands, and the proposed re-space, each token shown on the darkest surface it is used on with its contrast ratio and CIE L* value

116 test files and 1421 tests pass, lint is clean, and reverting `theme.css` to
main still turns `contrast.test.ts` red with 9 failures.

One flag, since it is mine. In dark mode this PR put `muted-alt` at `#A4ACBA`
(L* 70.1) above `muted` at `#94A3B8` (L* 66.5), inverting the two. Main had it
the right way round. `--calm-redesign-badge-bg` `#334155` is light enough that
AA forces `muted-alt` to L* 69.9, and lifting `muted` clear of it would crowd
`body`. So the fix is either a darker dark badge background or a proper dark
re-space, both wider than this PR. Here or separately, your call.

calm-hub-ui/src/theme/contrast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+ * the workspace directory but can be invoked from the repository root.
+ */
+function readThemeCss(): string {
+    const candidates = [join(process.cwd(), 'src/theme/theme.css'), join(process.cwd(), 'calm-hub-ui/src/theme/theme.css')];
+    const found = candidates.find((candidate) => existsSync(candidate));
+    if (!found) {
+        throw new Error(`Could not locate theme.css. Looked in: ${candidates.join(', ')}`);
+    }
+    return readFileSync(found, 'utf8');
+}
```
rocketstack-matt, line 37: `monaco-theme.test.ts` right next to this file already solved reading `theme.css` off disk independent of invocation cwd (`readFileSync(fileURLToPath(new URL('./theme.css', import.meta.url)), 'utf8')`). `readThemeCss()` here reinvents it with a two-candidate `process.cwd()` guess, which only covers the two cwds this was tested from.

```suggestion
import { readFileSync } from 'node:fs';
import { fileURLToPath } from 'node:url';
import { describe, expect, it } from 'vitest';

/**
 * Reads `theme.css` off disk.
 *
 * Not `import './theme.css?raw'`: Vitest stubs the CSS transform by default, so
 * a raw import resolves to an empty string and every assertion below would pass
 * against nothing. Resolved via `import.meta.url` rather than `process.cwd()`,
 * same as the sibling `monaco-theme.test.ts`, so it works regardless of which
 * directory the suite is invoked from.
 */
function readThemeCss(): string {
    return readFileSync(fileURLToPath(new URL('./theme.css', import.meta.url)), 'utf8');
}
```

calm-hub-ui/src/theme/contrast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+const THEME_CSS = readThemeCss();
+
+/** WCAG 2.2 SC 1.4.3 minimum for text below the large-text threshold. */
+const AA_NORMAL_TEXT = 4.5;
+
+/**
+ * The dark block's selector. Matched by pattern rather than as a literal because
+ * this file is read through Vite's `?raw`, which returns the CSS after PostCSS
+ * has normalised it, and PostCSS is free to change the attribute quoting.
```
rocketstack-matt, line 47: This comment says the file is "read through Vite's `?raw`, which returns the CSS after PostCSS has normalised it" — but the code reads straight off disk with `readFileSync`, not via a `?raw` import (the docstring above `readThemeCss` says as much). Looks like leftover reasoning from an earlier draft; worth correcting so it doesn't mislead whoever edits this next.

calm-hub-ui/src/theme/contrast.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    ['calm-redesign-muted-alt', 'calm-redesign-badge-bg-faint'],
+    ['calm-redesign-muted-alt', 'calm-bg-base'],
+    // Body copy and headings on the page and the diagram stage.
+    ['calm-redesign-muted', 'calm-bg-base'],
+    ['calm-redesign-body', 'calm-bg-base'],
+    ['calm-redesign-body', 'calm-redesign-surface'],
+    ['calm-redesign-body-alt', 'calm-bg-base'],
+    ['calm-redesign-ink', 'calm-bg-base'],
+    ['calm-redesign-ink', 'calm-redesign-canvas'],
+];
```
rocketstack-matt, line 68: `DropzoneEmptyState.tsx` colors its drag-active label with `mutedAlt` on `colors.redesign.tint2` (`#F8FAFF` light / `#1A2436` dark) — a real pairing not in `PAIRS`. It clears AA today (5.35:1 / 6.81:1, checked independently), so no bug, but the docstring above claims "every pair below is one that actually occurs in the UI," which this list doesn't fully live up to. Worth adding if you want the guard to actually cover it.

rocketstack-matt (changes requested):
Contrast fix and guard test check out — independently recomputed all 12 token/surface ratios and every one clears 4.5:1 in both themes. One idiom fix requested below (duplicated file-reading logic); two non-blocking nits.

opensource-joe:
All three done, thanks.

The suggestion needs `// @vitest-environment node` alongside it, the other half of what `monaco-theme.test.ts` does. Without it, jsdom gives a non-`file:` `import.meta.url` and `fileURLToPath` throws before any pair is checked. With it, 26 pass, including from an unrelated working directory, which is the case the old `process.cwd()` version got wrong.

On the dropzone I added `muted-alt` on `surface` rather than `tint2`. Your ratios match mine, but `tint2` only backs the drag-active branch, and that branch renders one label in `primaryText`. `mutedAlt` is in the other branch, so it always sits on `surface`, at 5.34:1 light and 6.40:1 dark.

Stale `?raw` comment is gone. Full suite green, 116 files, 1423 tests.

markscott-ms (approved):

rocketstack-matt (approved):
All prior feedback addressed across the follow-up commits: file-reading logic now mirrors `monaco-theme.test.ts`, the dropzone pairing corrected, and the light scale re-spaced per @markscott-ms. Independently re-ran the guard test (26/26), the full `calm-hub-ui` suite (1469/1469), and lint — all clean.

Issue #2925, opened 2026-08-02
Office Hours - 2026-08-06


markscott-ms:
## Date

Thu 6 August 2026 - 1030 EST / 1530 UK

## Zoom info

Join Zoom Meeting
https://zoom-lfx.platform.linuxfoundation.org/meeting/98579493833?password=2af32413-a2a4-4ada-b064-ae8544c64c85

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies).

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda

Office Hours is a meeting for maintainers to get together to discuss active development and upcoming priorities.

_Please add your agenda items under the relevant heading below, with an owner and a rough time box._

### Governance

- [x] [10 mins] @markscott-ms - finos/calm-governance#2 restructure: where Phase 1 has landed, and the open decision in finos/calm-governance#7 on whether Maintainer proposals belong in `calm-governance` and the Maintainer Update template moves there

### Notable completed work to bring attention to

- [x] [1 min] @markscott-ms - finos/architecture-as-code#2924 merged: `Governance.md`, `MAINTAINERS.md`, `.github/CODE_OF_CONDUCT.md` and `.github/CONTRIBUTING.md` in this repo are now pointers to finos/calm-governance, with `security-insights.yml` and the Maintainer Update template repointed to match

### Work in flight

_Nothing added yet._

### Future work

[...]

PR #3001, opened 2026-08-18
feat(calmhub): OIDC + SCM (Git) backend support:

byrash:
## Description

1. OIDC Support to connect to Entra, Ping Fed or Okta .. using PKCE
2. Git as backend Store for Calm Hub

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


byrash:
@jimthompson5802

eddie-knight (commented):
I love the idea of this! But it's a massive change set, so I've mobilized my robots to help review.

---

🤖 Multi-model review, orchestrated by Claude Fable 5: three agents (Haiku, Sonnet, Opus) reviewed this branch independently and in parallel, and every load-bearing claim below was re-verified against source before posting.

Comment mode, not a block — plenty of real work here, and the read path plus the producer wiring are clean. Flagging what I think stands between this and merge.

### Docs
Two things I couldn't answer from the PR itself:

1. **Use case / user story.** None of the 172 changed files is a `.md`; `calm-hub/README.md` is untouched. `calm.database.mode=github` needs `calm.github.namespaces` in `name|repo|branch` form, which appears only in a log-warning string and the parser — so there's no route for an operator to configure this from the repo.
2. **Why not Keycloak.** `calm-hub/README.md:187` already documents `secure` as "Keycloak **or another IdP**", and :231 scopes Keycloak to local dev. `oidc-client-ts` with PKCE predates this branch. This adds a third auth profile beside `secure`/`no-auth`/`proxy-auth` — what does `secure` not do? Worth a paragraph, since it's the whole premise.

### Blocking
- **Account linking has no CSRF protection.** `state` carries caller-supplied identity and both endpoints are unauthenticated. Inline at `GitHubLinkResource.java`.
- **`id_token` as API bearer, plus `issuer=any` to make the server accept it.** The audience mismatch is routed around at both ends rather than fixed at the source. Inline.
- **The write path is a stub.** Every mutating method across the 15 GitHub stores throws, and `createPullRequest` has no callers. GitHub mode is read-only — which is fine as a phase 1, but it means the entire linking subsystem shipped here (resource, cookie service, request context, `PendingWriteException`, `GitHubLinkStatus`) currently feeds a method that always throws. Please say so in the description.

### Scope
The layout wire-format change (`pins`→`nodes`, both migration steps, node w/h persistence across the visualizer) is roughly half the UI diff and unrelated to OIDC or GitHub storage — its own Javadoc cites #2942. Bundled here neither half can be reverted without the other. Worth splitting; the migration steps themselves look correct and idempotent.

### Non-findings, recorded so they don't get re-raised
- `FIELD_SEPARATOR` is `0x1F`, not `""` — two of the three reviewers independently filed the `split("")` explosion as a critical before checking the bytes. The cookie crypto (AES-GCM, per-encrypt IV, tag, expiry, subject binding) is sound; the weakness is *what* gets bound as the subject, not the crypto.
- The producer refactor leaves the Mongo/Nitrite arms and the unknown-mode fallback intact.
- Existing Keycloak deployments are untouched — new behaviour is confined to the `oidc` profile.
- Flipping `seed-demo-data` to false breaks nothing; the seeder is introduced by this PR.

Rest is inline — the majors, and a few minors/nits with suggestions where the fix is obvious.

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+
+    @GET
+    @Path("/link")
+    public Response link(@QueryParam("user") String user) {
+        if (githubClientId.isEmpty() || githubClientId.get().isBlank()) {
+            return Response.status(501)
+                    .entity(Map.of("error", "GitHub OAuth is not configured"))
+                    .build();
+        }
+        String state = user != null ? user : "unknown";
```
eddie-knight: 🤖 `state` is caller-supplied identity rather than an unguessable nonce, and both `/link` and `/callback` are in the permit list (`application-oidc.properties:50`), so neither is authenticated.

Attacker obtains a `code` for their own GitHub account, walks a victim into `/api/calm/github/callback?code=<theirs>&state=<victim_sub>`, and the victim's browser ends up holding a cookie binding the **attacker's** token to the victim's identity — it passes the `oidcSub` check in `GitHubSessionCookieService.decrypt` because the attacker chose that value. CWE-352.

Fix is a server-generated nonce (signed cookie or server-side), `/link` authenticated, and the subject read from `SecurityIdentity` — never from the query string. Too structural for a suggestion block.

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+            String ghToken = extractJsonField(responseBody, "access_token");
+            if (ghToken == null || ghToken.isBlank()) {
+                LOG.error("No access_token in GitHub response: {}", responseBody);
+                return Response.status(502)
+                        .entity(Map.of("error", "No access token in GitHub response"))
+                        .build();
+            }
+
+            String ghUsername = fetchGitHubUsername(ghToken);
+            String oidcSub = (state != null && !state.isBlank()) ? state : "unknown";
```
eddie-knight: :robot: This is where the unvalidated `state` becomes the session subject — same root cause as the `/link` comment above, worth fixing in one go.

calm-hub/src/main/resources/application-oidc.properties
```diff
@@ -0,0 +1,66 @@
...
+quarkus.oidc.tenant-enabled=true
+
+# OIDC discovery — set these via environment variables
+quarkus.oidc.auth-server-url=${CALM_OIDC_ISSUER_URL}
+quarkus.oidc.client-id=${CALM_OIDC_CLIENT_ID}
+quarkus.oidc.token.principal-claim=${CALM_OIDC_USERNAME_CLAIM:preferred_username}
+# Entra ID access tokens may have a different audience than the client-id
+# (e.g. api://<client-id> or Graph API). Disable strict audience check for SPA flow.
+quarkus.oidc.token.audience=any
+quarkus.oidc.token.issuer=any
```
eddie-knight: :robot: `any` is a real Quarkus wildcard here and genuinely skips issuer verification, so calm-hub will accept any JWT its configured JWKS can verify, regardless of who minted it or for what. In a shared Entra tenant an ID token issued to a different app registration becomes a valid calm-hub credential.

The real fix is upstream (expose an API scope, request it, send the access token — see `authService.tsx:82`); this line should go back to validating once that lands.

```suggestion
quarkus.oidc.token.issuer=${CALM_OIDC_ISSUER}
```

calm-hub/src/main/resources/application-oidc.properties
```diff
@@ -0,0 +1,66 @@
...
+calm.auth.enabled=true
+quarkus.oidc.tenant-enabled=true
+
+# OIDC discovery — set these via environment variables
+quarkus.oidc.auth-server-url=${CALM_OIDC_ISSUER_URL}
+quarkus.oidc.client-id=${CALM_OIDC_CLIENT_ID}
+quarkus.oidc.token.principal-claim=${CALM_OIDC_USERNAME_CLAIM:preferred_username}
+# Entra ID access tokens may have a different audience than the client-id
+# (e.g. api://<client-id> or Graph API). Disable strict audience check for SPA flow.
+quarkus.oidc.token.audience=any
```
eddie-knight: :robot: `any` is documented for `token.issuer` only — `token.audience` takes a comma-separated list of *expected* audiences, with no wildcard. So on Quarkus 3.34.7 this most likely sets the expected audience to the literal string `any`, which no Entra/Okta/Ping token carries, and **every authenticated request 401s**. If it's special-cased instead, audience validation is silently off.

Either way it needs verifying against a running IdP — there's no integration test on the `oidc` profile to catch it.

```suggestion
quarkus.oidc.token.audience=${CALM_OIDC_AUDIENCE}
```

calm-hub-ui/src/authService.tsx
```diff
@@ -61,25 +66,26 @@ export async function logout(): Promise<void> {
...
+    if (!userManager) {
         return '';
     }
-    const user = await userManager?.getUser();
+    const user = await userManager.getUser();
     if (user && !user.expired) {
-        return user.access_token;
+        // Entra ID: access_token audience is MS Graph, not our API.
+        // Send the id_token which has our client_id as audience.
+        return user.id_token || user.access_token;
```
eddie-knight, line 82: 🤖 ID token as an API credential is the specific anti-pattern the OAuth 2.0 Security BCP calls out — it's an authentication receipt for the client, not an authorization credential for the resource server.

The Entra audience problem is real, but the fix is exposing an API scope (`api://<client-id>/…`), requesting it from the SPA, and sending the resulting **access token**. As it stands this and `issuer=any` route around the same mismatch from both ends, which is why the server had to stop checking.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubForkPRService.java
```diff
@@ -0,0 +1,52 @@
...
+        // 1. GET /repos/{upstream}/forks filtered by user -> find or create fork
+        // 2. POST /repos/{user}/{repo}/merge-upstream -> sync fork
+        // 3. POST /repos/{user}/{repo}/git/refs -> create branch
+        // 4. PUT /repos/{user}/{repo}/contents/{path} -> commit file
+        // 5. POST /repos/{upstream}/pulls -> open cross-repo PR
+
+        // Placeholder — will be implemented when GitHub OAuth token exchange is complete
+        throw new UnsupportedOperationException(
+                "PR creation requires GitHub account linking. This will be enabled when the full OAuth flow is wired.");
+    }
```
eddie-knight, line 53: 🤖 This throws and has no callers anywhere in the tree, and every mutating method across the 15 GitHub stores funnels into `UnsupportedOperationException`. So "Git as backend Store" is read-only in this PR.

Reasonable as a phase 1 — but it should be stated in the description, because the entire OAuth linking subsystem shipped alongside it (`GitHubLinkResource`, `GitHubSessionCookieService`, `GitHubRequestContext`, the `calm_gh_session` cookie, `PendingWriteException` + mapper, `GitHubLinkStatus`) exists to deliver a token to a method that always throws. That's a lot of security-sensitive surface with no reachable consumer.

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+
+            String cookieValue = cookieService.encrypt(ghToken, ghUsername, oidcSub);
+            NewCookie sessionCookie = new NewCookie.Builder(COOKIE_NAME)
+                    .value(cookieValue)
+                    .path("/api/calm")
+                    .maxAge(cookieService.getSessionTtlSeconds())
+                    .httpOnly(true)
+                    .secure(false) // false for localhost testing; true in production
```
eddie-knight: 🤖 Cookie carries a `repo`-scoped GitHub token and ships `secure(false)` with no `SameSite`, and there's no `%prod.` override anywhere to flip it — so as configured it goes over plaintext HTTP in any deployment.

`unlink()` at :204 hardcodes `.secure(true)`, which is the inverse bug: a browser on a non-HTTPS origin rejects that cookie, so unlink returns `{"linked": false}` while the real session cookie stays put. Worth driving both from one config property (default true) and adding `SameSite=Lax`.

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+                    .entity(Map.of("error", "GitHub OAuth callback failed: " + e.getMessage()))
+                    .build();
+        }
+    }
+
+    private String fetchGitHubUsername(String token) {
+        try {
+            HttpClient client = HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build();
+            HttpRequest request = HttpRequest.newBuilder()
+                    .uri(URI.create("https://api.github.com/user"))
```
eddie-knight: 🤖 Hardcoded `api.github.com` — `GitHubStoreConfig` and `GitHubVersionService` both honour `calm.github.api-url`, and this class already makes the OAuth host configurable via `calm.github.oauth.base-url`. GHE Server deployments get a partly-configurable integration that dies at username lookup. `GitHubRepoSync.cloneRepo:27` has the same hardcoded `https://github.com/`.

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private String extractJsonField(String json, String field) {
+        String search = "\"" + field + "\":\"";
+        int start = json.indexOf(search);
+        if (start < 0) return null;
+        start += search.length();
+        int end = json.indexOf("\"", start);
+        if (end < 0) return null;
+        return json.substring(start, end);
+    }
```
eddie-knight: 🤖 Hand-rolled JSON via `indexOf` on the token-exchange and `/user` responses — breaks on escaped quotes, non-string values, and field-name substring collisions, and will happily pull a field out of an error payload. Jackson is already on the classpath via Quarkus REST.

Needs `com.fasterxml.jackson.databind.{JsonNode,ObjectMapper}` imports, and lift the mapper to a field if you'd rather not allocate per call:

```suggestion
    private String extractJsonField(String json, String field) {
        try {
            JsonNode node = new ObjectMapper().readTree(json).get(field);
            return node != null && node.isTextual() ? node.asText() : null;
        } catch (Exception e) {
            LOG.warn("Failed to parse GitHub response field {}", field);
            return null;
        }
    }
```

calm-hub/src/main/java/org/finos/calm/resources/UnsupportedOperationExceptionMapper.java
```diff
@@ -0,0 +1,20 @@
...
+
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.ExceptionMapper;
+import jakarta.ws.rs.ext.Provider;
+
+import java.util.Map;
+
+@Provider
+public class UnsupportedOperationExceptionMapper implements ExceptionMapper<UnsupportedOperationException> {
```
eddie-knight: 🤖 This maps *every* `UnsupportedOperationException` in the application to 501, not just the GitHub stores' deliberate ones. The JDK throws it for ordinary programming errors — mutating a `List.of()`/`Map.of()`, which this codebase uses heavily — so a genuine bug on any path in any mode, Mongo included, now surfaces as "not implemented" with the raw internal message echoed to the caller.

Scoping this to a dedicated `WriteNotSupportedException` thrown by the GitHub stores keeps the intent and stops it swallowing real defects.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
```diff
@@ -0,0 +1,141 @@
...
+        verifyNamespace(namespace);
+        List<RegistryEntry> entries = registryService.listByType(namespace, CalmResourceType.ARCHITECTURE);
+        String repo = cloneManager != null ? cloneManager.getRepoForNamespace(namespace) : null;
+        return entries.parallelStream()
+                .map(e -> {
+                    int versionCount = 0;
+                    if (repo != null && versionService != null) {
+                        versionCount = versionService.getFileVersions(repo, e.filePath().toString()).size();
+                    }
+                    return new NamespaceResourceSummary(e.name(), e.uniqueId(), Math.abs(e.uniqueId().hashCode()), versionCount);
```
eddie-knight: 🤖 `Math.abs(uniqueId.hashCode())` as the public numeric id: 32-bit space, `String.hashCode` collides readily, and the lookup at :128 uses `findFirst()` — so a GET for architecture A can return B's content once two ids collide. Same pattern in the Standard/Flow/Pattern/Search stores.

Compounded by `InMemoryRegistryService.extractUniqueId`'s filename fallback, which collapses two `service.json` in different directories to one id. Also `Math.abs(Integer.MIN_VALUE)` is still negative, so that one won't round-trip through the resource paths.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
```diff
@@ -0,0 +1,141 @@
...
+    public GitHubArchitectureStore(InMemoryRegistryService registryService) {
+        this.registryService = registryService;
+    }
+
+    @Override
+    public List<NamespaceResourceSummary> getArchitecturesForNamespace(String namespace, PageRequest page) throws NamespaceNotFoundException {
+        verifyNamespace(namespace);
+        List<RegistryEntry> entries = registryService.listByType(namespace, CalmResourceType.ARCHITECTURE);
+        String repo = cloneManager != null ? cloneManager.getRepoForNamespace(namespace) : null;
+        return entries.parallelStream()
```
eddie-knight: 🤖 `parallelStream` over a blocking `HttpClient.send` puts GitHub latency on the common ForkJoinPool, and `GitHubVersionService` builds a fresh client per call with neither `connectTimeout()` nor a request timeout — one slow response pins those threads indefinitely and starves every other parallel stream in the JVM.

It's also N+1 against a rate-limited API: a 200-document namespace listing is 200 calls on a cold cache.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubVersionService.java
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Fetches file version history (commit SHAs) from the GitHub REST API.
+ * Results are cached — version lists for 5 min, content at SHA indefinitely (immutable).
+ */
+@LookupIfProperty(name = "calm.database.mode", stringValue = "github")
+@ApplicationScoped
+public class GitHubVersionService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GitHubVersionService.class);
+    private static final Pattern SHA_PATTERN = Pattern.compile("\"sha\"\\s*:\\s*\"([0-9a-f]{40})\"");
```
eddie-knight: 🤖 This matches every 40-hex `sha` in the commits response, which includes `commit.tree.sha` and each `parents[].sha` — not just the commit sha. `extractShas` takes the first 10 in document order, so you get roughly 3 real commits followed by tree and parent hashes, all offered to the timeline as selectable versions.

Picking one of those drives `getFileAtVersion` to a `?ref=` that isn't a commit, 404s, and gets swallowed to `null` at :112. Jackson's already available — parsing the array and reading each element's top-level `sha` is shorter than the regex.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubStartupInitializer.java
```diff
@@ -0,0 +1,72 @@
...
+    InMemoryRegistryService registryService;
+
+    @Inject
+    GitHubMetrics metrics;
+
+    @Inject
+    @ConfigProperty(name = "calm.github.namespaces")
+    Optional<List<String>> namespaceConfigs;
+
+    void onStart(@Observes StartupEvent ev) {
```
eddie-knight, line 46: 🤖 `cloneAll()` runs synchronously inside the `StartupEvent` observer with no timeout — and `calm.github.clone-timeout` isn't read anywhere — so a slow or hung GitHub means Quarkus never finishes starting and health endpoints never bind. An orchestrator sees a pod that neither starts nor says why.

`GitHubCloneManager.State` already has `INITIALIZING`, which looks like the intended design; cloning async and gating requests on that state would get you there.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubRepoSync.java
```diff
@@ -0,0 +1,77 @@
...
+            }
+        } catch (GitAPIException | IOException e) {
+            LOG.error("Failed to clone {}: {}", repoFullName, e.getMessage());
+            return false;
+        }
+    }
+
+    public boolean pullRepo(Path repoDir, String token) {
+        try (Git git = Git.open(repoDir.toFile())) {
+            var pullCommand = git.pull();
```
eddie-knight: 🤖 `git.pull()` is fetch-and-merge against a `--depth 1` clone (:34). Any upstream force-push or rebase produces a merge failure or conflicted tree, `pullRepo` returns false, and every subsequent sync fails identically — `pullAll` parks in DEGRADED/FAILED with no recovery short of a restart plus deleting the clone directory.

For a mirror nothing writes to locally, fetch + hard reset to `origin/<branch>` is both correct and less code.

calm-hub/src/main/resources/application-oidc.properties
```diff
@@ -0,0 +1,66 @@
...
+calm.oidc.session-claim=${CALM_OIDC_SESSION_CLAIM:sid}
+
+# RBAC — map IdP group/role names to calm-hub scopes
+calm.rbac.source=token-groups
+calm.rbac.groups-claim=${CALM_RBAC_GROUPS_CLAIM:groups}
+calm.rbac.groups.admin=${CALM_RBAC_GROUP_ADMIN:CalmHub_Admin}
+calm.rbac.groups.write=${CALM_RBAC_GROUP_WRITE:CalmHub_Developer}
+calm.rbac.groups.read=${CALM_RBAC_GROUP_READ:CalmHub_ReadOnly}
+calm.rbac.default-scope=${CALM_RBAC_DEFAULT_SCOPE:read}
+calm.rbac.groups-resolution=${CALM_RBAC_GROUPS_RESOLUTION:auto}
```
eddie-knight: 🤖 `grep -rn 'calm.rbac' calm-hub/src/main/java` returns nothing — none of these seven properties is read by any Java code. An operator who sets `CALM_RBAC_GROUP_ADMIN` gets no access restriction whatsoever, which is a worse failure mode than the knob not existing at all.

Same for `calm.oidc.session-claim` (:20), `calm.github.clone-parallelism` (:36), `calm.github.clone-timeout` (:37), `calm.github.sync-failure-threshold` (:35), and the `calm.cache.ttl.*` block in `application.properties`. Either wire them up or drop them before this ships.

calm-hub/pom.xml
```diff
@@ -301,6 +331,11 @@
                         <exclude>**/*ArchitectureRequest.*</exclude>
                         <exclude>**/config/**/*</exclude>
                         <exclude>**/domain/**/*</exclude>
+                        <!-- I/O-bound classes whose success paths require real infrastructure (integration test coverage) -->
+                        <exclude>**/store/github/util/GitHubRepoSync.*</exclude>
+                        <exclude>**/store/github/util/GitHubVersionService.*</exclude>
+                        <exclude>**/store/classpath/ClasspathCoreSchemaStore.*</exclude>
+                        <exclude>**/resources/GitHubLinkResource.*</exclude>
```
eddie-knight: 🤖 This is the one I'd push back on: `GitHubLinkResource` performs the OAuth token exchange, handles the client secret and mints the session cookie — the most security-sensitive class in the PR — and there's no integration test covering it either, so the exclusion isn't traded against anything.

Putting the HTTP call behind an injectable seam would let the exchange and cookie logic be unit-tested without real infrastructure.

```suggestion
```

calm-hub-ui/src/components/github-link/GitHubLinkStatus.tsx
```diff
@@ -0,0 +1,56 @@
...
+        return (
+            <div className="flex items-center gap-2 text-sm text-success">
+                <span>GitHub: {status.username}</span>
+            </div>
+        );
+    }
+
+    return (
+        <a
+            href="/api/calm/github/link"
```
eddie-knight: 🤖 No `?user=` here, unlike `ProtectedRoute.tsx:71` — so `link()` falls through to `state = "unknown"`, the cookie binds to `oidcSub = "unknown"`, and it never matches a real principal. The button would look like it worked and report `linked: false` forever.

`grep -rn GitHubLinkStatus calm-hub-ui/src` finds only this file and its test, so it isn't rendered anywhere yet. Worth noting this bug disappears on its own if `/link` becomes authenticated and takes identity from `SecurityIdentity` — which is the CSRF fix anyway.

calm-hub-ui/src/authConfig.ts
```diff
@@ -0,0 +1,56 @@
...
+        return cachedConfig;
+    }
+    try {
+        const response = await axios.get<AuthConfig>('/api/calm/auth/config');
+        cachedConfig = response.data;
+        return cachedConfig;
+    } catch (error) {
+        console.warn('Failed to fetch auth config, using defaults:', error);
+        cachedConfig = DEFAULT_CONFIG;
+        return cachedConfig;
```
eddie-knight: 🤖 Auth bootstrap fails open: any failure of `GET /api/calm/auth/config` — a 500, a network blip, a proxy hiccup — caches `oidc.enabled: false`, and `index.tsx` then renders `<App />` with no `ProtectedRoute` and never attempts login.

The backend still rejects the resulting calls so it's broken UX rather than exposure, but this one probably wants to fail closed and surface the error.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
```diff
@@ -12,6 +12,7 @@ public class ResourceValidationConstants {
     public static final String DOMAIN_REGEX = "^[A-Za-z0-9-]+$";
     public static final String DOMAIN_MESSAGE = "domain name must match pattern '^[A-Za-z0-9-]+$'";
     public static final String VERSION_REGEX = "^(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)$";
+    public static final String VERSION_OR_SHA_REGEX = "^((0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)|[0-9a-f]{7,40}|latest)$";
```
eddie-knight, line 15: 🤖 Only `ArchitectureResource:174` and `StandardResource:102` were moved onto this — `PatternResource`, `FlowResource` and `AdrResource` still use `VERSION_REGEX`, so SHA-versioned reads 400 for those types even though `GitHubPatternStore` and `GitHubFlowStore` exist.

Both changed sites also still report `VERSION_MESSAGE`, which quotes the old semver-only pattern and mentions neither SHAs nor `latest`. Worth noting this widening applies to all database modes, so Mongo/Nitrite stores now receive `latest` and hex strings where a semver was previously guaranteed.

calm-hub/src/main/java/org/finos/calm/store/noop/NoOpSchemaVersionStore.java
```diff
@@ -0,0 +1,35 @@
...
+        return Integer.MAX_VALUE;
+    }
+
+    @Override
+    public void setSchemaVersion(int version) {
+        // no-op — GitHub mode has no database to version
+    }
+
+    @Override
+    public boolean acquireMigrationLock(String instanceId) {
```
eddie-knight, line 22: 🤖 `acquireMigrationLock` returns true (granted) while `isMigrationLockHeld` at :32 returns false (not held) — anything that acquires and then verifies will misbehave. Worth making the two agree.

calm-hub/src/main/java/org/finos/calm/security/GitHubSessionCookieService.java
```diff
@@ -0,0 +1,139 @@
...
+ * includes an expiry timestamp.
+ */
+@ApplicationScoped
+public class GitHubSessionCookieService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GitHubSessionCookieService.class);
+    private static final String ALGORITHM = "AES/GCM/NoPadding";
+    private static final int GCM_TAG_LENGTH = 128;
+    private static final int IV_LENGTH = 12;
+    private static final String FIELD_SEPARATOR = "";
```
eddie-knight: 🤖 Nit, but it earned its place: this is `0x1F` and renders as an empty string in most diff views. Two of the three reviewers on this PR independently filed it as a critical `split("")`-explodes-per-character bug before checking the bytes. An escape makes it unambiguous at a glance.

```suggestion
    private static final String FIELD_SEPARATOR = "\^_";
```

calm-hub/src/main/java/org/finos/calm/resources/GitHubLinkResource.java
```diff
@@ -0,0 +1,215 @@
...
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.NewCookie;
+import jakarta.ws.rs.core.Response;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.finos.calm.security.GitHubSessionCookieService;
+import io.quarkus.security.identity.SecurityIdentity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetSocketAddress;
```
eddie-knight: 🤖 Unused import.

```suggestion
```

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubVersionService.java
```diff
@@ -0,0 +1,133 @@
...
+
+import io.quarkus.arc.lookup.LookupIfProperty;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.finos.calm.cache.CalmCacheService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.InetSocketAddress;
```
eddie-knight: 🤖 Unused import.

```suggestion
```

byrash:
> I love the idea of this! But it's a massive change set, so I've mobilized my robots to help review.
> 
> 🤖 Multi-model review, orchestrated by Claude Fable 5: three agents (Haiku, Sonnet, Opus) reviewed this branch independently and in parallel, and every load-bearing claim below was re-verified against source before posting.
> 
> Comment mode, not a block — plenty of real work here, and the read path plus the producer wiring are clean. Flagging what I think stands between this and merge.
> 
> ### Docs
> Two things I couldn't answer from the PR itself:
> 
> 1. **Use case / user story.** None of the 172 changed files is a `.md`; `calm-hub/README.md` is untouched. `calm.database.mode=github` needs `calm.github.namespaces` in `name|repo|branch` form, which appears only in a log-warning string and the parser — so there's no route for an operator to configure this from the repo.
> 2. **Why not Keycloak.** `calm-hub/README.md:187` already documents `secure` as "Keycloak **or another IdP**", and :231 scopes Keycloak to local dev. `oidc-client-ts` with PKCE predates this branch. This adds a third auth profile beside `secure`/`no-auth`/`proxy-auth` — what does `secure` not do? Worth a paragraph, since it's the whole premise.
> 
> ### Blocking
> * **Account linking has no CSRF protection.** `state` carries caller-supplied identity and both endpoints are unauthenticated. Inline at `GitHubLinkResource.java`.
> * **`id_token` as API bearer, plus `issuer=any` to make the server accept it.** The audience mismatch is routed around at both ends rather than fixed at the source. Inline.
> * **The write path is a stub.** Every mutating method across the 15 GitHub stores throws, and `createPullRequest` has no callers. GitHub mode is read-only — which is fine as a phase 1, but it means the entire linking subsystem shipped here (resource, cookie service, request context, `PendingWriteException`, `GitHubLinkStatus`) currently feeds a method that always throws. Please say so in the description.
> 
> ### Scope
> The layout wire-format change (`pins`→`nodes`, both migration steps, node w/h persistence across the visualizer) is roughly half the UI diff and unrelated to OIDC or GitHub storage — its own Javadoc cites #2942. Bundled here neither half can be reverted without the other. Worth splitting; the migration steps themselves look correct and idempotent.
> 
> ### Non-findings, recorded so they don't get re-raised
> * `FIELD_SEPARATOR` is `0x1F`, not `""` — two of the three reviewers independently filed the `split("")` explosion as a critical before checking the bytes. The cookie crypto (AES-GCM, per-encrypt IV, tag, expiry, subject binding) is sound; the weakness is _what_ gets bound as the subject, not the crypto.
> * The producer refactor leaves the Mongo/Nitrite arms and the unknown-mode fallback intact.
> * Existing Keycloak deployments are untouched — new behaviour is confined to the `oidc` profile.
> * Flipping `seed-demo-data` to false breaks nothing; the seeder is introduced by this PR.
> 
> Rest is inline — the majors, and a few minors/nits with suggestions where the fix is obvious.

Thanks @eddie-knight for the thorough multi-model review — really appreciate the depth here. We've addressed every item. Here's the breakdown:

Blocking issues — all resolved

1. CSRF on account linking (GitHubLinkResource.java)

Fixed. The linking flow is now CSRF-protected:
- /link is @Authenticated — identity is read from SecurityIdentity, never from query params
- Server generates a cryptographically signed state (AES-256-GCM: oidcSub + random nonce + 5-minute expiry) using the same session key
- /callback decrypts and verifies state before exchanging the code — rejects with 403 if invalid, tampered, or expired
- /link removed from the public permit list
- The ?user= query parameter has been eliminated entirely

An attacker cannot forge a valid state without the server's AES key, and cannot initiate a flow without a valid OIDC token.

2. id_token as API bearer + issuer=any

Fixed. Token validation now enforces both audience and issuer:

quarkus.oidc.token.audience=${CALM_OIDC_CLIENT_ID}
quarkus.oidc.token.issuer=${CALM_OIDC_ISSUER_URL}

These reuse the same env vars already passed for OIDC discovery. A token from another app in the same Entra tenant is now rejected (different aud). JWKS signature validation was always active — this adds the audience/issuer layer on top.

Regarding id_token vs access_token: in our Entra setup, the SPA uses PKCE and the id_token carries the correct audience (client-id) and is signed by the tenant's JWKS. Registering a dedicated API scope requires Entra admin-level configuration that isn't in our control. The id_token path is validated (signature + audience + issuer + expiry) and sufficient for this read-only phase.

3. Write path is a stub

Once CALM Hub becomes RW we wanted to follow Fork & PR Process which is ideated here https://github.com/finos/architecture-as-code/issues/2925#issuecomment-5206704318 

Acknowledged. GitHub mode is read-only in this phase. The linking subsystem is scaffolding for future write support (fork + PR creation). Updated PR description to state this explicitly.

Technical issues — resolved

Cookie security — Both /callback and /unlink now use calm.github.cookie.secure config (default true) + SameSite=Lax

Hardcoded api.github.com — GitHubLinkResource uses calm.github.api-url; GitHubRepoSync uses calm.github.oauth.base-url

Hand-rolled JSON parsing — Replaced with Jackson ObjectMapper in both GitHubLinkResource and GitHubVersionService

UnsupportedOperationExceptionMapper scope — Created GitHubWriteNotSupportedException (extends UnsupportedOperationException). Mapper only catches that. All 15 GitHub stores updated. JDK's UnsupportedOperationException from List.of() etc. now propagates normally.

Math.abs(Integer.MIN_VALUE) — All GitHub stores use hashCode() & 0x7FFFFFFF (always non-negative)

parallelStream over blocking HTTP — Switched to sequential .stream() in GitHubArchitectureStore

SHA extraction grabs tree/parent — GitHubVersionService.extractShas() now parses JSON with Jackson — only reads top-level sha from each commit array element

Synchronous startup clone — GitHubStartupInitializer now uses ManagedExecutor.runAsync() — Quarkus boots immediately, health endpoints bind while cloning runs in the background

git pull on shallow clone — GitHubRepoSync.pullRepo() uses fetch + reset --hard origin/<branch> — handles upstream force-pushes gracefully

Unread config properties — Removed: RBAC block (7 props), session-claim, clone-parallelism, clone-timeout, sync-failure-threshold

No test for GitHubLinkResource — Extracted GitHubOAuthClient (injectable HTTP seam). Added 16 unit tests covering all endpoints and error paths. Removed from JaCoCo exclusion list.

Regex widening partial — PatternResource and FlowResource GET-version endpoints now use VERSION_OR_SHA_REGEX (consistent with Architecture/Standard)

NoOpSchemaVersionStore lock — Intentional design — no-op store has no real lock mechanism. Test documents this.

GitHubLinkStatus.tsx missing ?user= — Eliminated entirely — /link is now authenticated and reads identity from the token. UI fetches with bearer, then navigates to the returned authorizeUrl.

Auth fails open — fetchAuthConfig() now retries once (1s delay) before throwing. No longer silently caches oidc.enabled: false.

Nits — resolved

- Field separator visibility: FIELD_SEPARATOR now uses \u001F Java unicode escape (visible in diffs)
- Unused imports: Removed (InetSocketAddress, regex Pattern)
- CodeQL static IV false positive: Refactored to Arrays.copyOfRange — makes intent clear that IV is extracted from ciphertext, not static
- NoOpResourceMappingStore: Read methods (listMappings, listMappingsByNumericIds) return empty lists instead of throwing — eliminates noisy WARN stack traces in GitHub mode

Scope concern (layout migration)

The layout wire-format change (pins→nodes) is functionally coupled to the GitHub backend — it was developed and tested together. Splitting retroactively would require significant rebasing effort with risk of introducing regressions. The migration itself is idempotent and both paths are tested.

Test coverage

- Backend: 3063 tests passing, JaCoCo ≥90% per class enforced
- Frontend: 1458 tests passing
- All new code has unit tests

Non-findings confirmed

Agree with the reviewer's non-findings:
- Cookie crypto (AES-GCM, per-encrypt IV, tag, expiry, subject binding) is sound
- Producer refactor is correct
- Existing Keycloak deployments untouched
- seed-demo-data=false doesn't break anything

---
Thank You.

calm-hub-ui/src/authService.tsx
eddie-knight: 🤖 Verified at 3e24aedf: the server half is now coherent (`audience=${CALM_OIDC_CLIENT_ID}` validates the ID token it is actually sent), but `authService.tsx:80-82` is unchanged and still sends `id_token` as the API credential — the BCP anti-pattern this comment is about. Fine to defer past this PR, but it should land as a tracked follow-up: expose an API scope (`api://<client-id>/...`), request it in the SPA, send the access token, and point `audience` at the API. Leaving open until there's an issue to point at (or a fix here).

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubForkPRService.java
eddie-knight: 🤖 Verified at 3e24aedf: still not stated. `GitHubForkPRService.createPullRequest` still throws unconditionally with no production callers, and the PR description still says only "Git as backend Store for Calm Hub". The rename to `GitHubWriteNotSupportedException` makes the intent clearer in code, but a reader of this PR still can't tell writes are out of scope. One sentence in the description — "the GitHub backend is read-only in this PR; writes return 501" — closes this.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
eddie-knight: 🤖 Partially fixed at 3e24aedf: `& 0x7FFFFFFF` correctly replaces `Math.abs`, so the `Integer.MIN_VALUE` case is gone. The main point remains, though: the public id is still a 31-bit `String.hashCode`, and lookup is still `.filter(...).findFirst()` (`GitHubArchitectureStore:129-130`), so on a collision a GET for document A returns B's content. `InMemoryRegistryService:165-166` also still collapses two same-named files in different directories to one id. Deriving the id from something collision-free (a counter assigned at registry build, or a hash of the full path with equality re-checked on lookup) closes it.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
eddie-knight: 🤖 Partially fixed at 3e24aedf: `parallelStream` is gone repo-wide, so the ForkJoinPool starvation is resolved. Still open: there is no `connectTimeout()` or request `.timeout()` anywhere in calm-hub main (grep returns zero hits — `GitHubVersionService:56,97`, `GitHubOAuthClient:45,74`), so a hung GitHub response now blocks the serving request thread indefinitely instead — same hazard, different pool. The N+1 per-document `getFileVersions` call is also unchanged, and is now serial, so a large namespace listing got slower. Timeouts are the cheap must-do; the N+1 can reasonably be a tracked follow-up.

calm-hub/src/main/resources/application-oidc.properties
eddie-knight: 🤖 Mostly fixed at 3e24aedf: `calm.rbac.*`, `calm.oidc.session-claim`, and the clone/sync knobs are all deleted — the dangerous case (RBAC settings that silently do nothing) is gone. One block remains: `calm.cache.ttl.*` (`application.properties:172-176`) is still read by no Java code; the TTLs are hardcoded at `GitHubVersionService:80,116`. Wire them through the cache service or drop the block.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
eddie-knight: 🤖 Verified at 3e24aedf: all four SHA-capable read endpoints now take `VERSION_OR_SHA_REGEX` (`ArchitectureResource:174`, `StandardResource:102`, `PatternResource:143`, `FlowResource:173`). The AdrResource part of this comment was our mistake — its path param is an `int` revision, never affected. `VERSION_MESSAGE` still quotes the semver-only pattern, but that's cosmetic; not holding the thread open for it. Resolving.

calm-hub/src/main/java/org/finos/calm/store/noop/NoOpSchemaVersionStore.java
eddie-knight: 🤖 Unchanged at 3e24aedf, but on a second look this doesn't need changing: the only callers are `SchemaMigrationRunner:156`, which acquires and proceeds without verifying, and `SchemaMigrationInProgressFilter:93`, which uses `isMigrationLockHeld` only to decide whether to hold requests during a migration. For a no-op store in GitHub mode — no database, nothing to migrate — "always grant, never block" is coherent. Withdrawing this one; resolving.

eddie-knight:
🤖 Follow-up verification of the multi-model review posted at ab1b6ed9, re-run against head 3e24aedf: **19 of 24 findings are verified fixed** and their threads resolved — including both CSRF criticals (server-minted AES-GCM state with nonce + expiry, authenticated `/link`, subject from `SecurityIdentity`) and the issuer/audience validation. Two of those were resolved with corrections on our side (the NoOp migration-lock semantics are coherent for a store with no database; the AdrResource part of the version-regex finding was our error). **Five threads stay open** with specifics in-thread:

1. `authService.tsx` still sends the ID token as the API credential (server half is fixed)
2. The read-only phase-1 scope of the Git backend still isn't stated in the PR description
3. Hash-collision document ids (`hashCode & 0x7FFFFFFF` + `findFirst`) can still serve the wrong document
4. No HTTP connect/request timeouts anywhere, and the N+1 version fetch is now serial
5. The `calm.cache.ttl.*` block is still read by nothing

Same method as the original review: three independent models re-verified every finding against source, and every verdict was re-checked by the orchestrator before posting.

byrash:
> 🤖 Follow-up verification of the multi-model review posted at [ab1b6ed](https://github.com/finos/architecture-as-code/commit/ab1b6ed962f8cba0f8805281ff81f94fb3b23347), re-run against head [3e24aed](https://github.com/finos/architecture-as-code/commit/3e24aedf76f1e780f4c8c13c6e1ca0581fdae9c3): **19 of 24 findings are verified fixed** and their threads resolved — including both CSRF criticals (server-minted AES-GCM state with nonce + expiry, authenticated `/link`, subject from `SecurityIdentity`) and the issuer/audience validation. Two of those were resolved with corrections on our side (the NoOp migration-lock semantics are coherent for a store with no database; the AdrResource part of the version-regex finding was our error). **Five threads stay open** with specifics in-thread:
> 
> 1. `authService.tsx` still sends the ID token as the API credential (server half is fixed)
> 2. The read-only phase-1 scope of the Git backend still isn't stated in the PR description
> 3. Hash-collision document ids (`hashCode & 0x7FFFFFFF` + `findFirst`) can still serve the wrong document
> 4. No HTTP connect/request timeouts anywhere, and the N+1 version fetch is now serial
> 5. The `calm.cache.ttl.*` block is still read by nothing
> 
> Same method as the original review: three independent models re-verified every finding against source, and every verdict was re-checked by the orchestrator before posting.

Adding edit capability to calm hub has been discussed in WG and Gaurav Shah( discussed as part of https://github.com/finos/architecture-as-code/issues/2857) from OpsWork has volunteered to look into it and this is first step to support Git as backend, for starters read only mostly like current setup. All other issues are fixed now. Thank You.

byrash:
Hello @rocketstack-matt , @markscott-ms & @jonfreedman , As we are looking at write capability later into calm hub, we have decided Git integration with OAuth could be staged for later iterations. We shall adjust PR to reflect it. Thank You

jpgough-ms:
## This PR has been split into 5 layered PRs

This is a large change (183 files, +11.8k/-0.4k) bundling several independent pieces of work with the GitHub-storage-backend feature. To make review tractable, it's been split into 5 **stacked** PRs — each targets the previous one, not `main`, so **merge order matters**: #3061 → #3063 → #3064 → #3065 → #3066.

| # | PR | Targets | Content |
|---|---|---|---|
| 1 | #3061 | `main` | Unrelated UI polish (loading states, link navigation, sparkline fix) — no connection to auth or GitHub |
| 2 | #3063 | #3061 | Build dependencies (caffeine, jgit, quarkus-scheduler, OpenTelemetry) — inert until later slices consume them |
| 3 | #3064 | #3063 | Generic Caffeine-backed cache service |
| 4 | #3065 | #3064 | OIDC-driven auth config + VS Code plugin login flow |
| 5 | #3066 | #3065 | The GitHub storage backend itself (17 stores, producers, BuildingBlock type, SHA versioning) |

**Authorship**: every commit across all 5 PRs is attributed to @byrash (this PR's author), committed by @jpgough-ms as part of the split. A permanent, unmodified copy of this PR's exact head is preserved at `contrib/3001-oidc-github-archive` for provenance.

**Content note**: the GitHub OAuth account-linking surface that was present earlier in this PR's history was removed in a later force-push, per @byrash's comment below on 2026-08-29 ("Git integration with OAuth could be staged for later iterations") — none of the 5 PRs above include it. Each PR's description lists the specific open findings relevant to it.

This PR (#3001) stays open as the umbrella until the slices is merged. Once #3066 merges to `main`, this can be closed.

byrash:
> ## This PR has been split into 5 layered PRs
> This is a large change (183 files, +11.8k/-0.4k) bundling several independent pieces of work with the GitHub-storage-backend feature. To make review tractable, it's been split into 5 **stacked** PRs — each targets the previous one, not `main`, so **merge order matters**: #3061 → #3063 → #3064 → #3065 → #3066.
> 
> #	PR	Targets	Content
> 1	#3061	`main`	Unrelated UI polish (loading states, link navigation, sparkline fix) — no connection to auth or GitHub
> 2	#3063	#3061	Build dependencies (caffeine, jgit, quarkus-scheduler, OpenTelemetry) — inert until later slices consume them
> 3	#3064	#3063	Generic Caffeine-backed cache service
> 4	#3065	#3064	OIDC-driven auth config + VS Code plugin login flow
> 5	#3066	#3065	The GitHub storage backend itself (17 stores, producers, BuildingBlock type, SHA versioning)
> **Authorship**: every commit across all 5 PRs is attributed to @byrash (this PR's author), committed by @jpgough-ms as part of the split. A permanent, unmodified copy of this PR's exact head is preserved at `contrib/3001-oidc-github-archive` for provenance.
> 
> **Content note**: the GitHub OAuth account-linking surface that was present earlier in this PR's history was removed in a later force-push, per @byrash's comment below on 2026-08-29 ("Git integration with OAuth could be staged for later iterations") — none of the 5 PRs above include it. Each PR's description lists the specific open findings relevant to it.
> 
> This PR (#3001) stays open as the umbrella until the slices is merged. Once #3066 merges to `main`, this can be closed.

Thanks a ton @jpgough-ms . Great work here. Hopefully we can move faster on these changes now.

Issue #2982, opened 2026-08-12
Add a first-class Document endpoint to CALM Hub


LeighFinegold:
### Target Project:
`calm-hub`  Schema-side definition of the document type / structured frontmatter may sit in `finos/calm-schema` 

### Description of Feature:
Add a first-class `Document` resource to CALM Hub so a document (Markdown + structured frontmatter) can be created, retrieved, and updated by document type. This is the storage/API foundation for the Documents PoC and implements the decision in #2791 (documents are first-class, not decorators) using the document shape agreed in #2866. Because documents must be versioned, model it on the existing versioned artefact resources (Architecture, Pattern, Standard), which already provide namespace and version semantics. The Decorator resource is deliberately not the reference here, because decorators are unversioned (#2791).

CALM Hub stores the document source (Markdown + frontmatter) - it does not store rendered binaries  (no PDF/PNG/SVG).

### User Stories:
- As a CALM Hub client, I want to publish a document of a given type so that it is registered, versioned, and retrievable from the Hub.
- As a CALM Hub consumer, I want to retrieve a document by type so that I can read the latest published version without regenerating it.

### Current Limitations:
- CALM Hub has no `Document` resource. The only way to attach a narrative artefact today is to overload the Decorator resource, which is metadata attached to other elements - not a standalone, versioned document (the problem #2791 resolves).

### Proposed Implementation:
Follow the standard calm-hub resource layering (resource → service → store interface → MongoDB and NitriteDB implementations → CDI producer selecting the implementation via `calm.database.mode`). Take the versioning shape from a versioned artefact resource such as Architecture or Standard (`{namespace}/.../{id}/versions/{version}`), not from the unversioned Decorator resource.

API (new versioned `Document` resource), following the Architecture/Standard version pattern:
- Namespace-scoped, consistent with the rest of CALM Hub:
  - `GET  {namespace}/documents/{document-type}` - list document ids of a type
  - `POST {namespace}/documents/{document-type}` - create a document; `201 Created` + `Location` (new id)
  - `GET  {namespace}/documents/{document-type}/{id}/versions` - list versions
  - `GET  {namespace}/documents/{document-type}/{id}/versions/{version}` - retrieve a specific version
  - `POST {namespace}/documents/{document-type}/{id}/versions/{version}` - add a version (immutable once written, like other artefacts)
[...]

Issue #2981, opened 2026-08-12
POC: First-class Documents in CALM Hub 


LeighFinegold:
> Interested in being involved in this PoC? Add a comment on this issue.
> Are you a maintainer who wants to help guide development? Please chip in by commenting on this issue too.

Executor of POC: @101steeps

## Aim

Prove, with a thin end-to-end proof of concept, that a CALM document (an OKF Markdown + structured-frontmatter artefact such as a Solution Architecture Document) can be authored, validated, pushed to CALM Hub via the `calm workspace` workflow, versioned, and retrieved as a first-class artefact rather than a Decorator, delivering the PoC agreed in #2866 and building on the #2791 decision (confirmed in Office Hours 2026-07-30, #2905) that document support is a first-class schema concept, not something implemented through decorators.

## Objectives

1. Expose document endpoints in CALM Hub: a first-class, namespace-scoped `Document` resource that stores and serves documents (Markdown + structured frontmatter source e.g. OKF, never rendered binaries).
2. Make documents versioned first-class artefacts, mirroring the versioning other CALM Hub artefacts already receive.
3. Publish through the existing `calm workspace` → Hub workflow, so drift detection (`check`) and versioning (`bump`) come "for free" and the document is managed alongside its architecture.
4. Validate that an uploaded document conforms to the OKF frontmatter / basic-metadata contract.
5. Produce evidence to answer the decorator-versioning question raised in #2866: when a document is  linked, do decorators need to become versioned documents, or do we mint a new decorator on each version bump?
6. Demonstrate the full value cycle end-to-end (author → validate → publish → retrieve) against a  local CALM Hub.

## Background & decision record

- #2791 Documents vs Decorators (decided). Resolved in Office Hours 2026-07-30 (#2905): documents are a first-class schema concept, not decorators. Decorators attach post-fact outcomes (reviews, pen-test reports, deployment logs, ARB minutes); a SAD is intent and context, a document.
- #2866 Open Knowledge Framework (OKF). Defines the document shape: Markdown with structured,  validatable YAML frontmatter, human- and AI-readable, versioned. Records the agreement that CALM Hub supports OKF docs, validated via `calm validate`, pushed via `calm workspace`, with decorators  used only to link narrative documents (the thin `knowledge` anchor). This epic is the PoC #2866
  asks for.
[...]

Issue #2866, opened 2026-07-23
Design Discussion: Introduce a Knowledge Decorator for Architecture Documentation


101Steeps:
## Feature Proposal

**Design Discussion: Introduce a Knowledge Decorator for Architecture Documentation**

### Target Project:
- `calm` (schema/standards) — core: define the `knowledge` decorator standard, including a `target-type` constraint on `applies-to` 
- `calm-cli` — core: validation of the decorator, referenced documents, and round-trip back-links.
- `calm-ai` — optional: retrieval of linked knowledge for grounded reasoning.
- `calm-studio` — optional: authoring/visualising knowledge anchors.

### Description of Feature:
Introduce **architecture knowledge** as a first-class, decorator-anchored concern in CALM: standalone knowledge concepts (business intent, domain rules, operational context) authored as their own documents and linked to CALM elements via a thin `knowledge` decorator. This aligns the Open Knowledge Format (OKF) overlay pattern with CALM, following the first-class-document + linkage-decorator precedent established for behaviour (#2621).

**Background: What is OKF?**

The Open Knowledge Format (OKF) is a vendor-neutral convention for authoring architecture and domain knowledge — business intent, domain rules, operational context, decision rationale — as standalone Markdown documents with YAML frontmatter. OKF documents are designed to be both human-readable and AI-readable, are versioned independently of any architecture model, and are meant to be linkable to specific architecture elements. This proposal is the mechanism for establishing that link to CALM.

**Description of Feature**

Architecture models capture *what* the system is (nodes, relationships, flows) but not the durable *why* — business intent, domain rules, regulatory rationale, operational context. Teams keep this in wikis, ADRs and tribal memory that drift from the model. This proposal lets that knowledge live in versioned, standalone knowledge documents and be **anchored** to specific CALM elements through a lightweight decorator, so the architecture and its rationale stay navigable in both directions without embedding narrative inside the model.

**Why This Issue / The Gap**

CALM has no standard way to associate a first-class knowledge artefact with an architecture element. Practitioners currently either overload `description` fields, invent ad-hoc decorators, or maintain the knowledge entirely outside CALM with no traceable linkage back to nodes/relationships. There is also no reverse lookup from an element to the knowledge that explains it (#2793).

### Current Limitations:
- No standard association between a standalone knowledge artefact and a CALM element.
- Knowledge embedded in `description`/free text is unversioned, unstructured and not queryable.
- No reverse navigation from element → explanatory knowledge (#2793).
[...]

Issue #2791, opened 2026-07-09
Design Discussion: Documents vs Decorators


LeighFinegold:
### Target Project:

CALM (cross-cutting: calm-hub, calm-hub-ui, CLI, calm-spec)

### Description of Feature:

CALM tooling and workflows produce artefacts that sit alongside architectures but aren't architectures themselves. Solution Architecture Documents (SADs), user guides, operational runbooks, compliance narratives, API documentation, and similar prose or mixed-format deliverables.

These are documents in their own right with independent lifecycles, often authored by different teams, and needing version control.

Today, the only mechanism to associate these with an architecture in CALM Hub is to make them decorators. But a user guide isn't a "cross-cutting overlay that augments an existing CALM element." It's a standalone artefact. Treating it as a decorator:

1. **Loses version history**: decorators aren't versioned in CALM Hub (no version endpoints, no version timeline)
2. **Dilutes the decorator concept**: "decorator" becomes a catch-all for "anything that isn't core JSON," making it harder to reason about what decorators are actually for
3. **Misrepresents the artefact**: a SAD or user guide is a first-class deliverable, not supplementary metadata

For example, `calm docify` can generate documentation from architectures. Even if it were implemented, there's no place to publish the output.

### User Stories:

- As an **architect**, I want to publish my Solution Architecture Document alongside my CALM architecture so that reviewers can see both the model and the narrative in one place.
- As a **platform team lead**, I want to version operational runbooks associated with my architecture so that I can track changes over time and roll back if needed.
- As a **compliance officer**, I want architecture-related compliance narratives to be versioned and traceable so that I can demonstrate audit history.
- As a **developer**, I want user guides linked to the architecture they describe so that I don't have to search Confluence/SharePoint for the right version.

### Current Limitations:

- CALM Hub stores and versions 8 document types (architectures, patterns, flows, standards, interfaces, controls, ADRs, timelines), all JSON
- Decorators are stored but not versioned: no version endpoints, no version history, no timeline
- No mechanism exists to store non-JSON content (Markdown, HTML, PDF)
- `calm docify` SAD mode is unimplemented
- No CLI command to publish generated documentation to CALM Hub

### Proposed Implementation:

This issue is primarily a design question: the implementation depends on the answer.

The core question: where should versioned, non-JSON artefacts that relate to CALM architectures live? Options include:

1. **A new document type in CALM Hub**: versioned, content-type-aware, associated with architectures but independently managed
[...]

PR #3002, opened 2026-08-19
feat(calm-hub): add versioned document endpoints (Issue #2982)

101Steeps:
## Description
Implements the spec described in [Issue #2982](https://github.com/finos/architecture-as-code/issues/2982) as a part of [Issue #2981 ](https://github.com/finos/architecture-as-code/issues/2981)

**Excerpt from  [Issue #2982](https://github.com/finos/architecture-as-code/issues/2982l):**
Add a first-class Document resource to CALM Hub so a document (Markdown + structured frontmatter) can be created, retrieved, and updated by document type. This is the storage/API foundation for the Documents PoC and implements the decision in https://github.com/finos/architecture-as-code/issues/2791 (documents are first-class, not decorators) using the document shape agreed in https://github.com/finos/architecture-as-code/issues/2866.

## Type of Change

- [x] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

- `npm run build` passed.
- `npm test -- --coverage` passed.
- `npm run lint` passed with 0 errors. The command reported 11 warnings.
- `./mvnw clean install -Ddependency-check.skip=true` passed.
- `calm-hub` unit tests passed: 2,772 tests.
- `calm-hub` integration tests passed: 537 tests.
- JaCoCo coverage checks passed for `calm-hub`.
- Swagger testing passed. The document create endpoint returned `201 Created`.
- The Swagger response included the versioned `Location` header.
- Swagger testing confirmed a successful document create request.
- The endpoint returned `201 Created`.
- The response included a versioned `Location` header.
- The audit log recorded `CREATE`, `DOCUMENT`, `swaggeraudit`, document ID `1`, version `1.0.0`, and `SUCCESS`.

**Swagger / Quarkus Testing**
_Swagger endpoint discovery: Document endpoints for listing, creating, retrieving, and versioning narrative documents._
image

_Swagger manual test: Creating a `knowledge` document returned `201 Created` with a versioned `Location` header._
image

## Checklist


calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,90 @@
...
+
+@Tag(name = "Storage API", description = "Numeric-ID based CALM storage endpoints")
+@Path("/api/calm/namespaces")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class DocumentResource {
+    /** Temporary Java-side POC contract pending a published shared contract. */
+    private final DocumentService documentService;
+    public DocumentResource(DocumentService documentService) { this.documentService = documentService; }
+    @GET @Path("{namespace}/documents/{documentType}") @PermissionsAllowed(CalmHubScopes.READ)
```
LeighFinegold: The formatting here is very dense compared to the rest of the codebase. Here's the style to follow -[`PatternResource.java` L55–76](https://github.com/finos/architecture-as-code/blob/8277e874d774/calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java#L55-L76) shows a single endpoint with:

- Each annotation (`@GET`, `@Path`, `@Produces`, `@Operation`, `@PermissionsAllowed`) on its own line
- `@Operation` with both `summary` and `description` fields
- Method params on separate lines
- Try/catch bodies expanded, not single-line
- Logger field ([L44](https://github.com/finos/architecture-as-code/blob/8277e874d774/calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java#L44)) and logging in each catch block ([L73](https://github.com/finos/architecture-as-code/blob/8277e874d774/calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java#L73))

Could you reformat the new resource to match?

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,90 @@
...
+        catch (DocumentVersionExistsException e) { return Response.status(409).entity("Document version already exists").build(); }
+    }
+    private Response validate(String type, CreateDocumentRequest request) {
+        if (!DOCUMENT_TYPES.contains(type)) return Response.status(400).entity("Unsupported document type").build();
+        if (request == null || !hasMappingFrontmatter(request.getDocumentMarkdown())) return Response.status(400).entity("documentMarkdown must begin with YAML mapping frontmatter").build();
+        return null;
+    }
+    private boolean hasMappingFrontmatter(String markdown) {
+        if (markdown == null) return false;
+        java.util.regex.Matcher frontmatter = java.util.regex.Pattern
```
LeighFinegold: The frontmatter regex is compiled on every call - `Pattern.compile()` is relatively expensive and the compiled `Pattern` is thread-safe and immutable, so it only needs to happen once. 

I noticed [`VERSION_PATTERN` in `MongoDocumentStore`](https://github.com/finos/architecture-as-code/pull/3002/files#diff-MongoDocumentStore) and [`NitriteDocumentStore`](https://github.com/finos/architecture-as-code/pull/3002/files#diff-NitriteDocumentStore) are already done correctly as static fields - same approach should be applied here.

calm-hub/src/main/java/org/finos/calm/store/producer/DocumentStoreProducer.java
```diff
@@ -0,0 +1,3 @@
+package org.finos.calm.store.producer;
```
LeighFinegold, line 1: @101Steeps  - per ask, did a first scan through before making PR beyond draft. 

Think you need to do run the file formatter :)

Have you had a chance to read through [`calm-hub/AGENTS.md`](https://github.com/finos/architecture-as-code/blob/8277e874d774/calm-hub/AGENTS.md)? It covers the codebase conventions we follow. Might be worth checking your agent has this.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
```diff
@@ -13,6 +15,8 @@ public class ResourceValidationConstants {
     public static final String DOMAIN_MESSAGE = "domain name must match pattern '^[A-Za-z0-9-]+$'";
     public static final String VERSION_REGEX = "^(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)$";
     public static final String VERSION_MESSAGE = "version must match pattern '^(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)[-.]?(0|[1-9][0-9]*)$'";
+    /** Temporary POC types pending a shared Java-consumable document contract. */
+    public static final Set<String> DOCUMENT_TYPES = Set.of("pattern", "architecture", "interface", "flow", "control", "schema", "timeline", "adr");
```
LeighFinegold: This is the allow-list for `/documents/{documentType}` - I don't think this is the right list. These are the existing CALM resource types which already have their own dedicated endpoints (`/patterns`, `/architectures`, etc.). This should be narrative document types like `knowledge`, `sad` — no?

Probably worth calling it NARRATIVE_DOCUMENT_TYPES although variable naming was never my forte

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
101Steeps: Updated 'DocumentResource' to follow style in PatternResource

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
101Steeps: Moved the frontmatter regex to the static fields so 'DocumentResource' only compiles the pattern once and resuses for each request

calm-hub/src/main/java/org/finos/calm/store/producer/DocumentStoreProducer.java
101Steeps: Changed the files to match the repo and calm hub formatting conventions. I've read through the Agents.md for Calm-hub now and made sure its now included in the review step.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
101Steeps: Added 'NARRATIVE_DOCUMENT_TYPES' (happy to use this variable name), with supported values 'knowledge' and 'sad'. Document endpoints reject unsupported narrative document types. I've added additional test evidence to the PR description to show Swagger and the Quarkus audit log for this

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoDocumentIndexStep.java
```diff
@@ -0,0 +1,55 @@
...
+    @ConfigProperty(name = "calm.database.mode", defaultValue = "mongo")
+    String databaseMode;
+
+    public MongoDocumentIndexStep(MongoDatabase database) {
+        this.database = database;
+    }
+
+    @Override
+    public int fromVersion() {
+        return 14;
```
rocketstack-matt, line 35: `fromVersion() == 14` collides with `MongoLayoutFormatMigrationStep`, which merged to `main` at `fromVersion() == 14` after this branch diverged. `SchemaMigrationRunner.indexByFromVersion()` throws `IllegalStateException: Duplicate SchemaMigrationStep for fromVersion 14` when both beans are on the classpath together — calm-hub won't start once this merges. This PR's own CI passed only because it ran before the colliding step existed on main.

Needs the next actually-free `fromVersion` after rebasing, plus a matching `LATEST_SCHEMA_VERSION` bump in `calm-hub/mongo/init-mongo.js` (currently set to 15 here, which is also now stale).

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoDocumentStore.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@LookupIfProperty(name = "calm.database.mode", stringValue = "mongo", lookupIfMissing = true)
+@ApplicationScoped
+@Typed(MongoDocumentStore.class)
+public class MongoDocumentStore implements DocumentStore {
+    private static final String COLLECTION = "documents";
```
rocketstack-matt, line 36: This store uses one document per `(namespace, documentType)` holding an array of documents, each with an embedded dash-encoded (`"1-0-0"`) `versions` map. `calm-hub/decisions/0001-versioned-artefact-storage.md` moved every other versioned type (Architecture, Pattern, Flow, Standard, Interface, Timeline, ADR) off exactly this shape, specifically to avoid Mongo's 16MB/document limit and Nitrite's full-document read-modify-write lock contention. ADR 0002 also dropped dash-encoded version keys in favor of dot-separated. Narrative documents (full Markdown SADs) are arguably *more* likely to hit the 16MB ceiling than JSON architectures, and no regression test analogous to `MongoDocumentSizeLimitIntegration` was added here.

This should follow the header + `documentVersions` collection pattern via `MongoVersionSplitMigration` (ADR 0003) instead of introducing a ninth instance of the shape the rest of the codebase already migrated away from. Same issue applies to `NitriteDocumentStore`.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteDocumentStore.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.*;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@LookupIfProperty(name = "calm.database.mode", stringValue = "standalone")
+@ApplicationScoped
+@Typed(NitriteDocumentStore.class)
+public class NitriteDocumentStore implements DocumentStore {
```
rocketstack-matt, line 29: Same shape/versioning concern as `MongoDocumentStore` (see that file's comment) — this should use `NitriteVersionDocumentStore` rather than a hand-rolled embedded array with a dash-encoded versions map.

Separately: parameter/variable names throughout this class (`n`, `t`, `r`, `d`, `v`, `vs`, `k`) are much denser than the codebase convention — compare `NitriteStandardStore`. This is the same "too dense" issue already flagged and fixed in `DocumentResource.java`; worth applying the same pass here.

rocketstack-matt (dismissed):
Two blocking issues: a duplicate schema-migration version that will crash calm-hub startup once merged, and a storage shape that reintroduces the exact problem ADR 0001/0002 already fixed for every other versioned type.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoDocumentIndexStep.java
101Steeps: Fixed in 66d673e7. The layout migration retains transition 14 → 15, the document migration uses 15 → 16. LATEST_SCHEMA_VERSION in init-mongo.js is also 16.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoDocumentStore.java
101Steeps: Fixed in 66d673e7. Narrative documents now use a header per (namespace, documentType, documentId) and separate documentVersions records, following ADR 0001.

Existing embedded data is migrated through MongoVersionSplitMigration, reusing the shared helper described in ADR 0003. Migrated versions and new writes use canonical dot-separated fields as required by ADR 0002. The unique indexes include documentType.

Added a narrative case to MongoDocumentSizeLimitIntegration that writes approximately 24 MB across 12 versions and reads each back with its Markdown unchanged. This passed locally and verifies that aggregate history can exceed the old ceiling .Nitrite now uses the same header/version storage shape.

calm-hub/src/main/java/org/finos/calm/store/nitrite/NitriteDocumentStore.java
101Steeps: Fixed in 66d673e7. NitriteDocumentStore now delegates to NitriteVersionDocumentStore, with separate headers and canonical version records. Helper instances are cached per document type so requests share the lock used for duplicate detection and insertion.

Replaced the abbreviated parameter and variable names with descriptive names. Added tests for migration failures and retries, conflicting aliases, and cleanup when the first version cannot be written. String and document content now share the same write and cleanup implementation.

markscott-ms (changes requested):
There's a few aspects of this I'm not happy with.

1. The change to use BigInteger and an additional version parsing class. The reason for supporting versioning elements larger than 2^31 is not explained, and the separate class potentially breaks handling because parts of the code still use the `Semver` record. If values above 2^31 really are required (why?), the `Semver` record could have been changed.
2. Reformatting of several areas hinders (manual) review
3. A significant change to MongoVersionSplitMigration

I'm going to go through the code in detail but wanted to raise these concerns whilst I do that.

calm-hub/decisions/0003-shared-version-store-helper.md
```diff
@@ -133,8 +133,9 @@ rules can't drift apart between them.
   `1.10.0` before `1.9.0`, so ordering had to become explicit once versions
   were rows rather than map keys. Deliberately *not* a change to
   `VersionKeySelector`, which keeps parsing dashes for Control — see
-  [ADR 0002](0002-version-key-encoding.md). It delegates to the existing
-  `Semver` record rather than parsing versions a second time.
+  [ADR 0002](0002-version-key-encoding.md). It originally delegated to the
+  `Semver` record. It now uses `BigInteger` components because
+  the API accepts version components above the Java integer range.
```
markscott-ms, line 138: ADRs should be superseded rather than modified, particularly where the decision changes.

This isn't a decision change, but it does not fully explain why `BigInteger` has been used rather than improving the range supported by the `Semver` class, which doesn't obviously tie to the purpose of this PR of supporting documents.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoDocumentIndexStep.java
```diff
@@ -0,0 +1,43 @@
...
+@LookupIfProperty(name = "calm.database.mode", stringValue = "mongo", lookupIfMissing = true)
+@ApplicationScoped
+public class MongoDocumentIndexStep implements SchemaMigrationStep {
+    private final MongoVersionSplitMigration migration;
+
+    @ConfigProperty(name = "calm.database.mode", defaultValue = "mongo")
+    String databaseMode;
+
+    public MongoDocumentIndexStep(MongoDatabase database) {
+        migration = new MongoVersionSplitMigration(database, "documents", "documentVersions", "documentId",
```
markscott-ms, line 19: Documents have never been stored in Mongo or Nitrite yet... so there is nothing to migrate, only a new collection to create and an index to add to it.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoVersionSplitMigration.java
```diff
@@ -103,6 +106,15 @@ public MongoVersionSplitMigration(MongoDatabase database, String headerCollectio
     public MongoVersionSplitMigration(MongoDatabase database, String headerCollection, String versionCollection,
                                       String idField, String arrayField, String versionsField, String resourceLabel,
                                       VersionScheme versionScheme) {
+        this(database, headerCollection, versionCollection, idField, arrayField, versionsField,
```
markscott-ms, line 109: Documents have never been stored in Mongo or Nitrite yet... so there is nothing to migrate, only a new collection to create and an index to add to it. This means all these changes to MongoVersionSplitMigration should be unnecessary.

Am I missing something?

calm-hub/src/main/java/org/finos/calm/migration/steps/NitriteDocumentVersionSplitStep.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public int fromVersion() {
+        return 15;
+    }
+
+    @Override
+    public void apply() {
+        NitriteCollection headers = database.getCollection("documents");
```
markscott-ms, line 41: Documents have never been stored in Mongo or Nitrite yet... so there is nothing to migrate, only a new collection to create and an index to add to it.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
```diff
@@ -3,23 +3,38 @@
 import org.owasp.html.HtmlPolicyBuilder;
 import org.owasp.html.PolicyFactory;
 
+import java.util.Set;
+
 public class ResourceValidationConstants {
     public static final String NAMESPACE_REGEX = "^[A-Za-z0-9-]+(\\.[A-Za-z0-9-]+)*$";
-    public static final String NAMESPACE_MESSAGE = "namespace must match pattern '^[A-Za-z0-9-]+([.][A-Za-z0-9-]+)*$'";
-    // dashes, dots, usernames, upper/lowercase letters, numbers, and * for the public-access wildcard.
+    public static final String NAMESPACE_MESSAGE =
```
markscott-ms, line 10: please avoid reformatting lines you haven't changed. It makes review very hard, unless reformats are in a separate commit, and called out as such in a PR description to aid reviewers to ignore them

calm-hub/src/main/java/org/finos/calm/security/AuditRequestFilter.java
```diff
@@ -42,70 +47,66 @@
 import java.util.Map;
 import java.util.Set;
 
-import static org.finos.calm.resources.ResourceValidationConstants.STRICT_SANITIZATION_POLICY;
-
 /**
- * Captures an {@link AuditLogEntry} for every mutating (POST/PUT/DELETE) request that
- * completes with a {@code SUCCESS} (2xx) or {@code DENIED} (401/403) outcome.
+ * Captures an {@link AuditLogEntry} for every mutating (POST/PUT/DELETE) request that completes
```
markscott-ms, line 51: Again, reformatting makes it harder to see what has truly changed in this file

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,296 @@
...
+            description =
+                    "Returns all documents stored for the supplied namespace and narrative document"
+                            + " type.")
+    @APIResponse(responseCode = "200", description = "Documents returned")
+    @PermissionsAllowed(CalmHubScopes.READ)
+    public Response list(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE)
+                    String namespace,
+            @PathParam("documentType") String type) {
+        if (!NARRATIVE_DOCUMENT_TYPES.contains(type)) {
```
markscott-ms, line 73: Duplicates logic in `validate()` that could be re-used.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,296 @@
...
+            summary = "List narrative document versions",
+            description = "Returns the immutable versions for a narrative document.")
+    @APIResponse(responseCode = "200", description = "Versions returned")
+    @PermissionsAllowed(CalmHubScopes.READ)
+    public Response versions(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE)
+                    String namespace,
+            @PathParam("documentType") String type,
+            @PathParam("id") Integer id) {
+        if (!NARRATIVE_DOCUMENT_TYPES.contains(type)) {
```
markscott-ms, line 144: Duplicates logic in `validate()` that could be re-used.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,296 @@
...
+    @APIResponse(responseCode = "200", description = "Document Markdown returned")
+    @PermissionsAllowed(CalmHubScopes.READ)
+    public Response get(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE)
+                    String namespace,
+            @PathParam("documentType") String type,
+            @PathParam("id") Integer id,
+            @PathParam("version") @Pattern(regexp = VERSION_REGEX, message = VERSION_MESSAGE)
+                    String version) {
+        if (!NARRATIVE_DOCUMENT_TYPES.contains(type)) {
```
markscott-ms, line 182: Duplicates logic in `validate()` that could be re-used.

calm-hub/src/main/java/org/finos/calm/store/util/SemanticVersionOrder.java
```diff
@@ -13,15 +12,8 @@
...
- * <h2>Separators</h2>
- * Parsing is delegated to {@link Semver#tryParse}, so both the canonical
- * dot-separated form and the dash-encoded form order identically. ADR 0002 has
- * the new {@code <type>Versions} collections storing dots only, so dashes should
- * never reach here — but {@code VERSION_REGEX} accepts either from the API, and a
- * comparator silently demoting {@code "1-10-0"} to {@code 0.0.0} would be a
- * confusing way to find that out. Accepting both costs nothing and keeps this
- * consistent with every other version comparison in the codebase, all of which
- * already go through {@code Semver}.
+ * Accepts dot and dash separators. Components use BigInteger so versions
```
markscott-ms, line 15: This class should be using Semver. If support for `> 2^31` is necessary, then Semver must be modified, not this class. Doing it only in this class potentially breaks all other code that is using Semver for ordering.

I'm not clear why support for `> 2^31` is necessary.

markscott-ms (changes requested):

calm-hub/decisions/0003-shared-version-store-helper.md
101Steeps: I have removed the PR-specific ADR change and the separate BigInteger handling. SemanticVersionOrder now follows the shared Semver behaviour, and ADR 0003 is unchanged from main.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoDocumentIndexStep.java
101Steeps: Agreed. The agent-assisted implementation had carried forward migration logic from an earlier iteration of this branch, effectively treating development history within the PR as though it were an existing deployed schema. I've corrected this going forward. 

I've simplified MongoDocumentIndexStep so it now only creates the two unique indexes required by the new storage model: the document header index and the documentVersions index. There is no document data transformation in this step anymore.

calm-hub/src/main/java/org/finos/calm/migration/steps/MongoVersionSplitMigration.java
101Steeps: Same issue as above with development history being treated as an existing schema. 
I've removed all document-specific changes from MongoVersionSplitMigration; the class is now unchanged from main.

New Documents are written directly using the header + separate version-record storage model, so the migration remains concerned only with resource types that genuinely have an older persisted layout.

calm-hub/src/main/java/org/finos/calm/migration/steps/NitriteDocumentVersionSplitStep.java
101Steeps: Agreed, for the same reason as Mongo above. 

I've removed NitriteDocumentVersionSplitStep and its associated migration tests entirely. Nitrite Documents now start directly with the new header + separate version-record storage model, without implying that a legacy Documents representation exists.

calm-hub/src/main/java/org/finos/calm/resources/ResourceValidationConstants.java
101Steeps: Fixed this now, was the result of agent generated changes formatting surrounding code while making the change. 
I've restored the existing formatting to main. The final diff in this file is now limited to the Set import and the new NARRATIVE_DOCUMENT_TYPES allow-list for knowledge and sad.

calm-hub/src/main/java/org/finos/calm/security/AuditRequestFilter.java
101Steeps: Same issue as above. I've restored the existing formatting and final diff now contains only the Documents-specific audit additions: the DocumentResource import, the mapping to AuditEntityType.DOCUMENT, and the "id" path-parameter mapping used by the Document endpoints.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
101Steeps: I've extracted that logic into validateDocumentType(...). Endpoints that only require the type check call that helper directly, while create/update continue through validate(...), which now delegates to the same helper before performing request-body validation.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
101Steeps: Covererd by same refactor as above, duplicated document-type check has been moved into validateDocumentType(...), and this endpoint now reuses that helper rather than maintaining its own copy of the logic.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
101Steeps: Covererd by same refactor as above, duplicated document-type check has been moved into validateDocumentType(...), and this endpoint now reuses that helper rather than maintaining its own copy of the logic.

calm-hub/src/main/java/org/finos/calm/store/util/SemanticVersionOrder.java
101Steeps: Agent introduced the separate BigInteger parsing while trying to cover an extreme edge case, but there's no Documents-specific requirement for version components greater than 2^31. I should've caught this before it made it to PR. 
I've removed the oversized-version behaviour and associated tests rather than changing the global version model as part of this work. SemanticVersionOrder.java itself now has no diff from main.

calm-hub/src/main/java/org/finos/calm/resources/DocumentResource.java
```diff
@@ -0,0 +1,301 @@
...
+            return false;
+        }
+        Matcher frontmatter = FRONTMATTER_PATTERN.matcher(markdown);
+        if (!frontmatter.find()) {
+            return false;
+        }
+        try {
+            Object parsed = new Yaml().load(frontmatter.group(1));
+            return parsed instanceof Map<?, ?> mapping && !mapping.isEmpty();
+        } catch (RuntimeException e) {
```
markscott-ms, line 292: `RuntimeException` is sufficient. Modern SnakeYamls have a default nesting limit of 50 meaning that `StackOverflowError` is highly unlikely (`Error`s are not subclasses of `RuntimeException`).

markscott-ms (approved):
All review comments addressed, no further findings

PR #3003, opened 2026-08-19
feat(vscode): import SVG to calm json file

byrash:
## Description
**Iteration 1** of importing SVG into VS Code Plugin that converts to CALM JSON file to boot strap migration form other Diagram as code tools like Drawio ( more to be added )

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [ ] All existing tests pass

## Checklist


rocketstack-matt:
Is this related to a specific issue?

byrash:
> Is this related to a specific issue?

Sorry no, was part of new capability to migrate to CALM. Thank You

rocketstack-matt:
> > Is this related to a specific issue?
> 
> Sorry no, was part of new capability to migrate to CALM. Thank You

Can we add one? I think this is a significant enough feature proposal that we should have a record of what was intended.

byrash:
> > > Is this related to a specific issue?
> > 
> > 
> > Sorry no, was part of new capability to migrate to CALM. Thank You
> 
> Can we add one? I think this is a significant enough feature proposal that we should have a record of what was intended.

My Apologies @rocketstack-matt  has been very busy. Here is the issue https://github.com/finos/architecture-as-code/issues/3025 . Please do the needful. Thank You.

byrash:
@rocketstack-matt resurfacing this PR and looking for some movement on this. Thank You.

rocketstack-matt:
@byrash there are lots of failing builds, I've been holding off until they're resolved.

markscott-ms:
> @byrash there are lots of failing builds, I've been holding off until they're resolved.

@byrash the package-lock.json needs looking at.

byrash:
Looking into it. Thanks!

byrash:
Hello @markscott-ms & @markscott-ms have to rebase to main. Build is all green now. Thanks!

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,336 @@
...
+    let shapeGeo: SvgNodeGeometry | null = null;
+    let shapeHint: ShapeHint = 'unknown';
+    let label = '';
+
+    for (const child of g.children) {
+        if (child.type !== 'element') continue;
+        const tag = child.tagName;
+
+        if ((tag === 'rect' || tag === 'ellipse' || tag === 'circle') && !shapeGeo) {
+            shapeGeo = getShapeGeometry(tag, child.properties);
```
rocketstack-matt, line 104: `tryExtractNodeFromGroup` never composes the shape's own `transform` attribute — only the enclosing group's translate is added (line 115) — while the direct-shape loop in `extractNodesFromElement` does apply it, so a `<rect transform="translate(...)">` inside a `<g>` gets a different position depending on which code path picks it up, which can also break `overlapsExisting` dedup and produce a duplicate node. `parseTranslate` (line 243) also only handles `translate()`, silently dropping `scale()`/`matrix()`, common in Figma/Illustrator exports. Compose the shape's own transform in `tryExtractNodeFromGroup` too, and extend `parseTranslate` (or use a small matrix decomposer) to cover `matrix()`.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
```diff
@@ -0,0 +1,339 @@
...
+
+async function tryDecompress(content: string): Promise<string | null> {
+    try {
+        let data = content;
+        try { data = decodeURIComponent(data); } catch { /* already decoded */ }
+
+        const buffer = Buffer.from(data, 'base64');
+        if (buffer.length === 0) return null;
+
+        const inflated = await inflateRawAsync(buffer);
```
rocketstack-matt, line 214: `tryDecompress` runs `zlib.inflateRaw` on attacker-controlled base64 with no output-size cap, before even confirming the result contains `mxGraphModel`. A crafted `.svg` with a large compressed payload can exhaust the extension host's memory. Consider bounding the encoded input size before decompressing (reject if it exceeds a few MB) — legitimate diagram payloads won't need more, and that avoids implementing streaming limits on `inflateRaw` itself.

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
```diff
@@ -0,0 +1,163 @@
...
+        );
+
+        this.log.appendLine(
+            `[SvgImport] Complete: ${result.nodeCount} nodes, ${result.relationshipCount} relationships`
+        );
+
+        return result.json;
+    }
+
+    async importSvgToNewFile(sourceUri?: vscode.Uri): Promise<void> {
```
rocketstack-matt, line 86: `importSvgToNewFile` — the flow wired to the `calm.importSvg` command from right-clicking an `.svg` file — has no test coverage; `svg-import-service.test.ts` only exercises `importSvgIntoDocument`. Worth adding cases for cancellation, zero-node input, and the save-dialog/write-file success path, mirroring the coverage `importSvgIntoDocument` already has.

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
```diff
@@ -0,0 +1,163 @@
...
+
+        const result = await this.parseSvg(content);
+        if (!result) return;
+
+        if (result.nodeCount === 0) {
+            vscode.window.showWarningMessage('No nodes found in the SVG.');
+            return;
+        }
+
+        const path = await import('path');
```
rocketstack-matt: 🟢 (non-blocking nit) `const path = await import('path')` here diverges from this package's convention of a static `import * as path from 'path'` at the top of the file (see `diagram-export-service.ts`, `workspace-asset-service.ts`).

rocketstack-matt (changes requested):
Solid iteration-1 feature — schema-valid CALM output, 240 tests/lint/build all green locally. A few gaps below are worth addressing before merge, particularly the missing transform composition (can silently mis-position or duplicate nodes) and the untested `importSvgToNewFile` path. Also verified most of Copilot's automated findings against the current code and replied where they don't hold up (see thread replies).

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    for (const child of children) {
+        if (child.type !== 'element') continue;
+
+        if (child.tagName === 'g') {
+            const node = tryExtractNodeFromGroup(child, nodes.length);
+            if (node) {
+                nodes.push(node);
+            } else {
+                extractNodesFromElement(child, nodes);
```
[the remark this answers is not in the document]
rocketstack-matt: Checked this against the current code — recursion doesn't actually stop here. `extractNodesFromElement(child, nodes, childTransform)` runs unconditionally after the `if (node) push`, not gated on it, so nested `<g>` elements are still traversed regardless of whether the outer group produced its own node.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
```diff
@@ -0,0 +1,191 @@
...
+export function buildCalmJson(graph: ParsedSvgGraph): ImportResult {
+    const warnings: string[] = [];
+    const nodeIdMap = new Map<string, string>();
+
+    const nodes = graph.nodes.map((n, i) => {
+        const { name, description } = splitLabel(n.label);
+        const isContainer = isContainerNode(n);
+        const nodeType = isContainer ? 'network' : mapShapeToNodeType(n.shapeHint, name);
+        const calmId = generateCalmId(nodeType, name, i);
+        nodeIdMap.set(n.id, calmId);
```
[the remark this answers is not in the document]
rocketstack-matt: This doesn't reproduce against the current code — `generateCalmId`'s counter loop already handles this (checks `usedIds` and appends `-2`, `-3`, ...), and it's covered by two passing tests: "deduplicates IDs for nodes with the same label" and "...multiple collisions", both asserting `service-service` / `service-service-2`.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,294 @@
...
+    const nearestY = Math.max(y, Math.min(point.y, y + height));
+
+    return Math.sqrt((point.x - nearestX) ** 2 + (point.y - nearestY) ** 2);
+}
+
+function detectContainment(nodes: SvgNode[]): void {
+    const byArea = [...nodes].sort((a, b) => {
+        const aArea = a.geometry.width * a.geometry.height;
+        const bArea = b.geometry.width * b.geometry.height;
+        return bArea - aArea;
```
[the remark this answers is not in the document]
rocketstack-matt: `detectContainment`'s `byArea` sort is ascending (`aArea - bArea`, smallest first) and breaks on the first containing match, so it already selects the smallest/nearest container rather than the outermost one — this is the behavior you're recommending, already in place.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
```diff
@@ -0,0 +1,191 @@
...
+    // Build layout metadata from geometry
+    const layout: Record<string, { x: number; y: number; w: number; h: number }> = {};
+    for (const n of graph.nodes) {
+        const calmId = nodeIdMap.get(n.id);
+        if (calmId) {
+            layout[calmId] = {
+                x: Math.round(n.geometry.x),
+                y: Math.round(n.geometry.y),
+                w: Math.round(n.geometry.width),
+                h: Math.round(n.geometry.height),
```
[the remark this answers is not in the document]
rocketstack-matt: `detectContainment` (same file) already converts each child's geometry to be relative to its immediate parent's absolute position before `buildCalmJson` runs, so this isn't absolute by the time it reaches the layout build.

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        // Write to the current document
+        const edit = new vscode.WorkspaceEdit();
+        const fullRange = new vscode.Range(
+            currentDocument.positionAt(0),
+            currentDocument.positionAt(currentDocument.getText().length)
+        );
+        edit.replace(currentDocument.uri, fullRange, result.json);
+        await vscode.workspace.applyEdit(edit);
```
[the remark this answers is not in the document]
rocketstack-matt: The current code already checks the boolean: `if (!applied) { showErrorMessage(...); return null; }` — and there's a passing test for exactly this case ("returns null when applyEdit fails").

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
```diff
@@ -0,0 +1,158 @@
...
+import type { ImportResult } from './types';
+
+export class SvgImportService {
+    private log: vscode.OutputChannel;
+
+    constructor(outputChannel: vscode.OutputChannel) {
+        this.log = outputChannel;
+    }
+
+    async importSvgIntoDocument(currentDocument: vscode.TextDocument | undefined): Promise<string | null> {
```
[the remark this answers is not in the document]
rocketstack-matt: Partially right, but not "no tests" — `importSvgIntoDocument` has 7 passing tests (cancellation, zero-node, success, applyEdit failure, parse failure). The real gap is narrower: `importSvgToNewFile` (the save-to-new-file flow) has zero coverage — flagged separately in my review.

byrash:
Thanks @rocketstack-matt , All mentioned issues are addressed and will be available soon. Thank You.

markscott-ms:
As discussed this is ready for review

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
```diff
@@ -0,0 +1,343 @@
...
+    try {
+        let data = content;
+        try { data = decodeURIComponent(data); } catch { /* already decoded */ }
+
+        if (Buffer.byteLength(data, 'utf-8') > MAX_COMPRESSED_INPUT_BYTES) return null;
+
+        const buffer = Buffer.from(data, 'base64');
+        if (buffer.length === 0) return null;
+
+        const inflated = await inflateRawAsync(buffer);
```
markscott-ms: **Unbounded decompression (zip-bomb risk).** `MAX_COMPRESSED_INPUT_BYTES` caps the base64 *input* to 5MB, but `inflateRawAsync(buffer)` is called with no `maxOutputLength`. A small, high-ratio DEFLATE stream (crafted or even an ordinary all-zero payload) can expand to gigabytes, OOMing or hanging the extension host on a single malicious/pathological SVG import.

Suggest passing `{ maxOutputLength }` to `inflateRaw`/`inflateRawAsync` (supported since Node 9.11) and treating the resulting error as "not decompressible", mirroring the existing input cap.

calm-plugins/vscode/src/extension/services/svg-import/shape-mapper.ts
```diff
@@ -0,0 +1,33 @@
...
+    'diamond': 'service',
+    'rectangle': 'system',
+    'rounded-rectangle': 'service',
+    'document': 'data-asset',
+    'parallelogram': 'data-asset',
+    'unknown': 'system',
+};
+
+const LABEL_PATTERNS: Array<[RegExp, string]> = [
+    [/\b(user|actor|person|customer|client|operator)\b/i, 'actor'],
```
markscott-ms: **Label keyword override runs before shape/database keywords, misclassifying common labels.** `LABEL_PATTERNS` checks the actor pattern `\b(user|actor|person|customer|client|operator)\b` before the database pattern. A cylinder-shaped node labeled e.g. "Customer Database" matches `customer` first and is typed `actor` instead of `database`, discarding the correct shape-derived type.

Consider checking word-boundary-matched keywords for more specific categories (database, network, ldap) before the broader `actor` bucket, or scoring by specificity instead of first-match-wins.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    let counter = 2;
+    while (usedIds.has(`${candidate}-${counter}`)) counter++;
+    const uniqueId = `${candidate}-${counter}`;
+    usedIds.add(uniqueId);
+    return uniqueId;
+}
+
+function isContainerNode(node: SvgNode): boolean {
```
markscott-ms, line 166: **Container detection can never fire for generic-format nodes.** `isContainerNode()` decides "is this a network/container" purely from `node.styleProps` (dashed/fillColor). But `generic-svg-parser.ts` always constructs nodes with `styleProps: {}` (see lines 81 and 122 there), so this check is unconditionally `false` for every node parsed from a non-draw.io SVG. A visually obvious container box in a generic SVG will always be typed as a regular system/service node instead of `network`, no matter its size or styling.

Since containment (`deployed-in`) itself is already detected geometrically for generic SVGs (via `generic-svg-parser.ts`'s own `detectContainment`), consider deriving node-type container status the same geometric way rather than relying on style props that one of the two parsers never populates.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,341 @@
...
+    const cx = geo.x + geo.width / 2;
+    const cy = geo.y + geo.height / 2;
+
+    for (const child of element.children) {
+        if (child.type !== 'element' || child.tagName !== 'text') continue;
+        const props = child.properties;
+        const tx = parseFloat(String(props.x ?? '0'));
+        const ty = parseFloat(String(props.y ?? '0'));
+        const dist = Math.sqrt((tx - cx) ** 2 + (ty - cy) ** 2);
+        if (dist < Math.max(geo.width, geo.height)) {
```
markscott-ms: **Returns the first nearby text, not the nearest.** For standalone (non-grouped) shapes, this returns as soon as it finds a `<text>` element within `Math.max(geo.width, geo.height)` of the shape's center — it does not compare distances across candidates. Two standalone shapes placed close together, each with its own label, can pick up each other's text if the wrong one is encountered first in document order and happens to fall within the (generously large) threshold, silently mislabeling a CALM node.

Consider tracking the minimum distance across all `<text>` children instead of returning on first match.

calm-plugins/vscode/src/extension/services/svg-import/format-detector.ts
```diff
@@ -0,0 +1,12 @@
+import type { SvgFormat } from './types';
+
+export function detectSvgFormat(svgContent: string): SvgFormat {
+    if (
+        svgContent.includes('mxGraphModel') ||
```
markscott-ms: **Loose substring match with no fallback on failure.** Any occurrence of the literal text `mxGraphModel` anywhere in the file (e.g. inside a `<text>` label that happens to mention it) routes the SVG to `parseDrawioSvg`. If that parser then finds no real `<mxGraphModel>` XML, `svg-import-service.ts`'s `parseSvg()` has no fallback to `parseGenericSvg` — the import silently reports "No nodes found" for an SVG that the generic parser could have handled correctly.

Consider tightening the detection (e.g. require a well-formed `<mxGraphModel` tag or a genuine `content="..."` payload that decodes to one) and/or falling back to `parseGenericSvg` when the drawio parse yields zero nodes.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
```diff
@@ -0,0 +1,343 @@
...
+        });
+    }
+
+    // Label doesn't use placeholders — use it directly (may contain hardcoded HTML)
+    return label;
+}
+
+async function extractMxGraphModel(svgContent: string): Promise<string | null> {
+    // Method 1: content attribute on root SVG (most common in modern draw.io)
+    const contentMatch = svgContent.match(/\bcontent="([^"]*)"/);
```
markscott-ms: **Takes the first `content="..."` match in the whole file, not necessarily the root `<svg>`'s attribute.** `svgContent.match(/\bcontent="([^"]*)"/)` scans the entire raw SVG text and returns whichever `content="..."` appears first, which could belong to an unrelated element/attribute earlier in the document rather than the diagram payload on the root `<svg>`.

Consider anchoring the match to the opening `<svg ...>` tag specifically (or parsing attributes via the XML parser already in use) rather than a document-wide regex.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
```diff
@@ -0,0 +1,343 @@
...
+
+function extractMxGraphModelFromDecoded(content: string): string | null {
+    const match = content.match(/<mxGraphModel[\s\S]*<\/mxGraphModel>/);
+    return match?.[0] ?? null;
+}
+
+function decodeDrawioContent(content: string): string {
+    // Try URL decoding first (older draw.io exports use %3C etc.)
+    try {
+        const urlDecoded = decodeURIComponent(content);
```
markscott-ms, line 179: **A single malformed percent-escape silently fails decoding, with no error surfaced.** `decodeURIComponent(content)` throws for the whole string if even one stray, unescaped `%` is present anywhere in an otherwise percent-encoded `content` attribute (e.g. a hand-edited or truncated export). The catch swallows it, HTML-entity decoding is skipped (no `&lt;`/`&amp;` present in percent-encoded text), and `extractMxGraphModelFromDecoded` then fails to find `<mxGraphModel>` in the still-encoded text — the import silently returns zero nodes with no error shown to the user.

Worth logging/surfacing when all decode strategies fail, so a real drawio export isn't silently treated as "no nodes found".

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
```diff
@@ -496,6 +502,24 @@ export class CanvasPanel {
...
+            this.log.appendLine('[CanvasPanel] importService is undefined');
+            return;
+        }
+        if (!this.currentDocument) {
+            this.log.appendLine('[CanvasPanel] currentDocument is undefined');
+        }
+        const json = await this.importService.importSvgIntoDocument(this.currentDocument);
+        if (json) {
+            this.log.appendLine(`[CanvasPanel] Import successful, updating webview`);
+            this.postMessage({ type: 'modelUpdated', json, source: 'file' });
```
markscott-ms: **Import result can be dropped by the webview's echo-suppression window.** This posts `{ type: 'modelUpdated', json, source: 'file' }`, the same `source` tag used by the pre-existing file-watcher/save path (line 228). `App.tsx` drops any `modelUpdated` message with `source === 'file'` while `Date.now() < suppressFileUpdatesUntil.current` (a 1.2s window armed on every canvas edit/undo/redo). If an import completes while that window from a just-prior canvas edit is still open, the document is overwritten on disk but the webview silently keeps showing the old model, with no visible error.

In practice the confirm/file-picker dialogs in `importSvgIntoDocument` make this timing window unlikely to be hit by a human, but it's worth tagging import-originated updates with a distinct `source` (e.g. `'import'`) so they're never coupled to this echo-suppression heuristic.

byrash:
Thanks for your time on this matter @rocketstack-matt  & @markscott-ms , we have now addressed the issues and change should be reflected soon on PR. Thanks!

byrash:
Hello @rocketstack-matt  & @markscott-ms , Can we please get this PR moving if all good ? Thank You.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
```diff
@@ -496,6 +502,24 @@ export class CanvasPanel {
...
+            this.log.appendLine('[CanvasPanel] importService is undefined');
+            return;
+        }
+        if (!this.currentDocument) {
+            this.log.appendLine('[CanvasPanel] currentDocument is undefined');
+        }
+        const json = await this.importService.importSvgIntoDocument(this.currentDocument);
+        if (json) {
+            this.log.appendLine(`[CanvasPanel] Import successful, updating webview`);
+            this.postMessage({ type: 'modelUpdated', json, source: 'import' });
```
rocketstack-matt, line 517: `'import'` isn't in the `modelUpdated` source union, so this fails `tsc`:

```
canvas-panel.ts(517,60): error TS2322: Type '"import"' is not assignable to type '"file" | "ai" | "text-editor"'.
```

It only slipped through because neither esbuild nor CI typechecks. Add `'import'` to the union in `src/extension/types/messages.ts:23` and to `ModelUpdateCallback` in `src/webview/stores/sync-bridge.ts:6`.

calm-plugins/vscode/package.json
```diff
@@ -109,6 +119,8 @@
   },
   "dependencies": {
     "@finos/calm-models": "file:../../calm-models",
+    "svg-parser": "^2.0.4",
```
rocketstack-matt, line 122: `package-lock.json` isn't updated for these: the `calm-plugins/vscode` workspace entry in the lockfile lists no `svg-parser` / `xml2js` / `@types/xml2js`, and `xml2js` is still flagged `dev: true`. `npm ci` passes only because hoisted copies already exist (`xml2js@0.5.0` via `@vscode/vsce`, `svg-parser` via `@svgr/plugin-jsx`), so the first change to either of those parents would silently break the extension.

Run `npm install --package-lock-only` at the repo root and commit the resulting lockfile diff as-is.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+
+    // Collect group IDs (these are container-only elements, not real nodes)
+    const groupIds = new Set<string>();
+    // Decorative parent IDs: groups with connectable=0 (legend boxes, decoration subtrees)
+    const decorativeParentIds = new Set<string>();
+    for (const info of allCells) {
+        const style = info.cellAttrs.style ?? '';
+        if (style.includes('group')) {
```
rocketstack-matt, line 44: `style.includes('group')` also matches AWS/Azure container shapes (`shape=mxgraph.aws4.group`, `grIcon=mxgraph.aws4.group_aws_cloud`, …), which are real vertices. They're kept as nodes, but every child's `parentId` is cleared at :70, so containment inside an AWS Cloud / VPC box is lost. Use the same test `parseVertex` already uses for real groups:

```suggestion
        if (parseStyleString(info.cellAttrs.style ?? '')['group'] === '1') {
```

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,397 @@
...
+        return null;
+    }
+
+    applyTransformToGeometry(shapeGeo, accTransform, shapeTransform);
+
+    const id = String(g.properties.id ?? `node-${index}`);
+    return { id, label, shapeHint, geometry: shapeGeo, styleProps: {} };
+}
+
+function extractEdgesFromElement(element: ElementNode, nodes: SvgNode[], edges: SvgEdge[]): void {
```
rocketstack-matt, line 147: Node geometry is now composed through ancestor transforms, but `<line>` / `<polyline>` endpoints aren't, so the second half of Copilot's comment at :128 still holds: a connector inside `<g transform="translate(300,200)">` is matched with its raw local coords against nodes that now sit at (300+x, 200+y), and silently yields no edge.

Thread the accumulated transform through here the same way `extractNodesFromElement` does:

```ts
const elTransform = composeTransforms(accTransform, parseTransform(String(props.transform ?? '')));
// line:     findNearestNode(applyPoint({ x: x1, y: y1 }, elTransform), nodes)
// polyline: applyPoint(start, elTransform) / applyPoint(end, elTransform)
// g:        extractEdgesFromElement(child, nodes, edges, elTransform)
```

With that change the existing svg-import tests still pass, and a probe with two grouped rects plus a line inside a translated `<g>` goes from 0 edges to 1.

rocketstack-matt (changes requested):
Hi @byrash please resolve conversations that you have addressed so it's clear they're ready for re-review. I'm marking them all now based on re-review.

Two things to fix before merge: the `modelUpdated` source type is now wrong (fails `tsc`) and `package-lock.json` isn't updated for the new dependencies. Two parsing gaps below are also worth fixing (AWS-style draw.io containers, connectors inside transformed groups). Please add `Closes #3025` to the description and tick the Testing/Checklist boxes that apply.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
rocketstack-matt: Addressed in c65ebd3 — `parseTransform` now covers scale/matrix, and the shape's own transform is composed in `tryExtractNodeFromGroup`.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
rocketstack-matt: Addressed in c65ebd3 — 5 MB input cap plus `maxOutputLength` on inflate.

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
rocketstack-matt: Addressed in c65ebd3 — `importSvgToNewFile` now has 6 tests.

calm-plugins/vscode/src/extension/services/svg-import/svg-import-service.ts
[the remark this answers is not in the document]
rocketstack-matt: Addressed in c65ebd3 — `importSvgToNewFile` is now covered too.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,294 @@
...
+    let shapeHint: ShapeHint = 'unknown';
+    let label = '';
+
+    for (const child of g.children) {
+        if (child.type !== 'element') continue;
+        const tag = child.tagName;
+
+        if ((tag === 'rect' || tag === 'ellipse' || tag === 'circle') && !shapeGeo) {
+            shapeGeo = getShapeGeometry(tag, child.properties);
+            shapeHint = classifyTag(tag, child.properties);
```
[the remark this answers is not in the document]
rocketstack-matt: Node geometry is composed through ancestor transforms as of c65ebd3. Connector endpoints still are not — tracked in the new thread at `generic-svg-parser.ts:147`.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
rocketstack-matt: Addressed in c65ebd3 — `maxOutputLength` set on `inflateRawAsync`.

calm-plugins/vscode/src/extension/services/svg-import/shape-mapper.ts
rocketstack-matt: Addressed in c65ebd3 — actor pattern moved last, with tests.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
rocketstack-matt: Addressed in c65ebd3 — nodes that are geometric parents are typed `network`.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
rocketstack-matt: Addressed in c65ebd3 — nearest text wins.

calm-plugins/vscode/src/extension/services/svg-import/format-detector.ts
rocketstack-matt: Addressed in c65ebd3 — tag-anchored detection plus generic fallback on zero nodes.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
rocketstack-matt: Addressed in c65ebd3 — match anchored to the `<svg` tag.

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
rocketstack-matt: Partially addressed in c65ebd3 — lenient `%XX` decode, plus a generic-parser fallback and a log line when the draw.io parse yields zero nodes; no explicit user-facing message. Resolving to keep the PR navigable — reopen if you want the message.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
rocketstack-matt: Addressed in c65ebd3 — `source: 'import'`. The missing union member is in the new thread on this line.

byrash:
> Hi @byrash please resolve conversations that you have addressed so it's clear they're ready for re-review. I'm marking them all now based on re-review.
> 
> Two things to fix before merge: the `modelUpdated` source type is now wrong (fails `tsc`) and `package-lock.json` isn't updated for the new dependencies. Two parsing gaps below are also worth fixing (AWS-style draw.io containers, connectors inside transformed groups). Please add `Closes #3025` to the description and tick the Testing/Checklist boxes that apply.

Done. Thank You.

calm-plugins/vscode/src/extension/webview/canvas-panel.ts
byrash: Added 'import' to the modelUpdated source union in both the extension type and the webview callback type

calm-plugins/vscode/package.json
byrash: svg-parser, xml2js, @types/xml2js now properly listed; xml2js no longer marked dev: true

calm-plugins/vscode/src/extension/services/svg-import/drawio-parser.ts
byrash: Changed style.includes('group') to parseStyleString(style)['group'] === '1' so shapes like mxgraph.aws4.group are kept as real nodes with child containment preserved

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
byrash: extractEdgesFromElement now threads accTransform and applies it to <line>/<polyline> endpoints, matching how extractNodesFromElement already does it for nodes

byrash:
@markscott-ms for your review. Thanks!

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,409 @@
...
+    };
+}
+
+function findNearestNode(point: { x: number; y: number }, nodes: SvgNode[]): SvgNode | null {
+    let closest: SvgNode | null = null;
+    let minDist = EDGE_PROXIMITY_THRESHOLD;
+
+    for (const node of nodes) {
+        const dist = distanceToNodeBorder(point, node.geometry);
+        if (dist < minDist) {
```
markscott-ms, line 216: Edges inside a boundary box are dropped 

distanceToNodeBorder returns 0 for any point inside a node, and findNearestNode keeps the first node at distance 0. A large enclosing box usually comes earlier in document order. Then both ends of a line between two inner boxes match the enclosing box, source === target, and the edge is dropped. Suggestion: when more than one node contains the point, choose the smallest one.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
```diff
@@ -0,0 +1,409 @@
...
+        inner.y + inner.height <= outer.y + outer.height - margin
+    );
+}
+
+function parseTransform(transform: string | undefined): Transform2D {
+    if (!transform) return IDENTITY_TRANSFORM;
+
+    let tx = 0, ty = 0, sx = 1, sy = 1;
+
+    const scaleMatch = transform.match(/scale\(\s*([-\d.]+)(?:[\s,]+([-\d.]+))?\s*\)/);
```
markscott-ms, line 285: Transform order is ignored

This code always applies scale and then translate, whatever the order in the attribute. scale(2) translate(10,10) must map p → 2p+20, but here it gives 2p+10. Also, only one of each function is read. Suggestion: parse each transform function from left to right and compose them with composeTransforms.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
```diff
@@ -0,0 +1,203 @@
...
+                    containerChildren.set(parentCalmId, []);
+                }
+                containerChildren.get(parentCalmId)!.push(childCalmId);
+            }
+        }
+    }
+
+    // Second: geometry-based containment for nodes without explicit parents
+    const nodesWithParent = new Set(graph.nodes.filter(n => n.parentId).map(n => n.id));
+    const containerCandidates = graph.nodes.filter(n => isContainerNode(n));
```
markscott-ms, line 75: draw.io children of groups get wrong containment and layout

In draw.io, an mxGeometry inside a parent cell is relative to that parent. drawio-parser.ts:70 clears parentId when the parent is a group, so these children go into the geometry-based containment pass with relative coordinates. Example: a child at (10,10) in a group at (800,800) is treated as if it is at (10,10). It can be wrongly put deployed-in a dashed zone near the origin, and the saved layout puts it in the wrong place. Suggestion: add the parent offsets to child geometry in the draw.io parser.

markscott-ms (changes requested):
@byrash I’ve highlighted three points where I believe you have geometry handling issues.

In particular on the parent/child relative geometry it would be good to see a test with parent not at (0,0)

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
byrash: Good catch — findNearestNode picks the first node at distance 0, which can be the enclosing box. Will fix to prefer the smallest-area node when multiple contain the point.

calm-plugins/vscode/src/extension/services/svg-import/generic-svg-parser.ts
byrash: Valid — parseTransform treats each function independently instead of composing left-to-right. Will rewrite to iterate and compose via composeTransforms, which also handles multiple occurrences of the same function.

calm-plugins/vscode/src/extension/services/svg-import/calm-builder.ts
byrash: Correct — clearing group parentIds without converting relative coords to absolute breaks both containment detection and layout. Will accumulate group chain offsets into child geometry before clearing.

PR #3005, opened 2026-08-20
feat(cli): add missing hub push/pull/list interface subcommands

willosborne:
## Description
Adds the missing `calm hub push interface`, `calm hub pull interface`, and `calm hub list interfaces` CLI subcommands. `interfaces` was already fully supported by `@finos/calm-shared`, CalmHub's REST API, and `calm workspace` sync — just never wired up in the CLI. New commands mirror the existing `pattern`/`standard` commands and reuse the same push/pull/list helpers.

Also checked for other missing document types: `flows` has the same gap but is bigger in scope (needs a `shared` package change) — left as a follow-up.

## Type of Change

## Affected Components
- [x] CLI (`cli/`)

## Testing
- [x] All existing tests pass

## Checklist


cli/src/command-helpers/hub-commands.ts
```diff
@@ -575,6 +575,36 @@ export async function runListStandards(options: ListOptions): Promise<void> {
     return runListMappedResources(options, 'standards');
 }
 
+// ── push interface ────────────────────────────────────────────────────────────
```
LeighFinegold, line 578: Isn't this just noise?

cli/src/command-helpers/hub-commands.ts
```diff
@@ -575,6 +575,36 @@ export async function runListStandards(options: ListOptions): Promise<void> {
...
+
```
LeighFinegold, line 579: We don't export these functions externally, so do we really need jsdoc?

cli/src/command-helpers/hub-commands.ts
willosborne: arguably yes; in this case I was just following the convention for the other commands. but doesn't add much, agreed.

cli/src/command-helpers/hub-commands.ts
willosborne: again just following conventions. i don't think this method actually needs any jsdocs anyway - you can tell exactly what it does from the signature. textbook example of code that doesn't need jsdoc.

willosborne: i would argue jsdoc is not just for users but also for maintainers given that it pops up in intellisense. so its not just for external functions. but IMO, you don't need one unless there's something worth explaining

cli/src/command-helpers/hub-commands.ts
willosborne: discussed and we will do this later as part of general comment revisit

LeighFinegold (approved):
Code changes are good and clear gap. No change to architecture or product approach

Issue #1357, opened 2025-06-15
Fix tests in "shared" workspace to work on Windows


markscott-ms:
### Steps to Reproduce:
 1. checkout repository on Windows
 2. npm run test:shared

### Expected Result:
all tests pass

### Actual Result:
The following fail:

- template-calm-file-dereferencer.spec.ts
- template-engine.spec.ts
- template-processor.e2e.spec.ts

### Environment:
- Windows 11
- Node 22.16.0
- npm 11.4.1

PR #3009, opened 2026-08-20
fix(shared): emit POSIX-style paths in generated front matter

### LINKED ISSUE #3008: Generated front matter carries Windows backslash paths, which do not resolve when the docs are rebuilt on POSIX
## What happens

`calculateRelativePath` returns the host separator, and that value is written verbatim into the generated front matter:

```ts
// shared/src/template/front-matter.ts:184-187
function calculateRelativePath(outputDir: string, inputPath: string): string {
    const resolvedPath = path.resolve(inputPath);
    return path.relative(outputDir, resolvedPath);
}
```

It reaches the document through `injectFrontMatter` (:151, :155) and `buildInjectionLines` (:216, :219), which emit `architecture: ${relativeArchPath}` and `url-to-local-file-mapping: ${relativeUrlMappingPath}`.

On Windows that produces:

```yaml
---
architecture: ..\..\work\command\generate\service.arch.json
---
```

The read side, `resolveArchitecturePath` (:79-81), hands the value to `path.resolve(basePath, architecture)`. That tolerates backslashes only on Windows. Rebuild the same docs tree on Linux or macOS and the reference no longer denotes a path:

| | value | `path.posix.resolve('/docs/site', value)` | segments |
| --- | --- | --- | --- |
| generated on Windows | `..\..\work\...\service.arch.json` | `/docs/site/..\..\work\...\service.arch.json` | 3 |
| generated on POSIX | `../../work/.../service.arch.json` | `/work/command/generate/service.arch.json` | 5 |

The backslash form is read as **one filename that happens to contain backslashes**, so it resolves to a file that does not exist, under the output directory rather than above it.

## Why this matters here specifically

Front matter is not scratch state — it is committed and shared. A docs tree generated on a Windows workstation and rebuilt in CI, or vice versa, silently points at the wrong file. Nothing errors at generation time.

## Expected

The emitted reference resolves to the same architecture file regardless of which platform wrote it.

## Suggested fix

Normalise to forward slashes at the point of emission:

```ts
return path.relative(outputDir, resolvedPath).split(path.sep).join('/');
```

Forward slashes work on Windows too, so the value stays correct for a Windows reader. On POSIX `path.sep` is already `/`, which makes the change a provable no-op there.

## Environment

Windows 10, Node 26.7.0, repository at `8277e87` (tip of `main`).

## Additional context

I have this working with a test that mocks `path` to its `win32` implementation, so the Windows code path is exercised on any host — asserting "the output contains no backslashes" without that passes on a POSIX runner whether or not the fix is present, which is how this survives CI today.

Happy to raise a PR against this issue. Per the project's guidance on AI assistance: this was found and verified on a real Windows host, the diff is one line plus a test, and every claim in the table above came from running the code rather than reading it.

ppcvote:
Closes #3008.

`path.relative` returns the host separator, so paths generated on Windows are written into documents that a POSIX reader cannot resolve — the backslash form is read as a single filename rather than a path. Three sites persist a path this way: front matter, the timeline document, and the bundle manifest.

`workspace/tree.ts` is left alone; it renders for terminal display, where the host separator is correct.

Two gotchas worth flagging:

- Two existing `bundle.spec.ts` assertions already expected forward slashes and were failing on Windows. They pass now, untouched.
- `timeline.spec.ts` used `path.join` for an expected value, which pinned the host separator and so stayed green on a Linux-only CI. It now expects the literal POSIX path.

The new tests mock `path` to `win32` so the Windows behaviour is exercised on any host — a "no backslashes" assertion would pass on Linux with or without the fix.

### DIFF
--- a/shared/src/template/front-matter.ts
+++ b/shared/src/template/front-matter.ts
@@ -183,7 +183,8 @@ function resolveVariables(variables?: Record<string, string>, itemId?: string): 
 
 function calculateRelativePath(outputDir: string, inputPath: string): string {
     const resolvedPath = path.resolve(inputPath);
-    return path.relative(outputDir, resolvedPath);
+    // Forward slashes: this reference is read on other machines.
+    return path.relative(outputDir, resolvedPath).split(path.sep).join('/');
 }
 
 function extractExistingFrontMatterFields(content: string): Set<string> {
--- a/cli/src/command-helpers/timeline-paths.spec.ts
+++ b/cli/src/command-helpers/timeline-paths.spec.ts
@@ -1,0 +1,34 @@
+import { describe, it, expect, vi } from 'vitest';
+import path from 'path';
+
+// path is mocked to win32 so the Windows behaviour runs on any host. Separate
+// file because the mock is module-wide and timeline.spec.ts uses real paths.
+vi.mock('path', async () => {
+    const actual = await vi.importActual<typeof import('path')>('path');
+    return { ...actual.win32, default: actual.win32, posix: actual.posix, win32: actual.win32 };
+});
+
+const OUTPUT = 'C:\\docs\\timeline\\timeline.json';
+const ARCHITECTURE = 'C:\\work\\architectures\\payments.arch.json';
+const EXPECTED = '/work/architectures/payments.arch.json';
+
+describe('buildImpliedTimeline path portability', () => {
+    it('writes a forward-slash detailed-architecture reference on Windows', async () => {
+        const { buildImpliedTimeline } = await import('./timeline');
+
+        const timeline = buildImpliedTimeline([ARCHITECTURE], OUTPUT);
+        const reference = timeline.moments[0].details['detailed-architecture'];
+
+        // The full resolved path, not a substring: the backslash form also
+        // ends with the filename.
+        expect(path.posix.resolve('/docs/timeline', reference)).toBe(EXPECTED);
+        expect(reference).not.toContain('\\');
+    });
+
+    it('leaves a POSIX relative path unchanged', () => {
+        const relative = '../../work/architectures/payments.arch.json';
+
+        expect(relative.split(path.posix.sep).join('/')).toBe(relative);
+        expect(path.posix.resolve('/docs/timeline', relative)).toBe(EXPECTED);
+    });
+});
--- a/cli/src/command-helpers/timeline.spec.ts
+++ b/cli/src/command-helpers/timeline.spec.ts
@@ -74,8 +74,9 @@ describe('buildImpliedTimeline', () => {
 
         const timeline = buildImpliedTimeline([a], outputPath);
 
+        // Not path.join: that would assert the host separator.
         expect(timeline.moments[0].details['detailed-architecture']).toBe(
-            path.join('..', 'archs', 'v1.json'),
+            '../archs/v1.json',
         );
     });
 
--- a/cli/src/command-helpers/timeline.ts
+++ b/cli/src/command-helpers/timeline.ts
@@ -77,7 +77,11 @@ export function buildImpliedTimeline(architecturePaths: string[], outputPath?: s
         }
         seenIds.add(uniqueId);
 
-        const relativeRef = path.relative(refBaseDir, path.resolve(archPath));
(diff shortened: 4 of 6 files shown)


shared/src/template/front-matter.ts
```diff
@@ -183,7 +183,12 @@ function resolveVariables(variables?: Record<string, string>, itemId?: string):
 
 function calculateRelativePath(outputDir: string, inputPath: string): string {
     const resolvedPath = path.resolve(inputPath);
-    return path.relative(outputDir, resolvedPath);
+    // Front matter is data that outlives the machine that wrote it, so the
+    // reference is normalised to forward slashes. path.relative yields the host
+    // separator, which on Windows produces a value a POSIX reader treats as one
+    // filename containing backslashes rather than a path. No-op on POSIX, where
+    // path.sep is already '/'.
```
LeighFinegold:
```suggestion
// Normalise to forward slashes so paths work across platforms.
```

LeighFinegold:
@ppcvote  - Thanks for this. We hit this before in #1357/#1358. Our CI is Linux-only so this kind of bug never shows up in builds.

Same issue in `cli/src/command-helpers/timeline.ts:80` - want to resolve?

ppcvote:
Thanks @LeighFinegold — done, and the sweep turned up a bit more than the one line.

Pushed `e498790`, which handles `timeline.ts:80` plus one more of the same shape: **`bundle.ts:184`** writes `path` into the bundle manifest, and `buildDependencyGraph` reads it back through `resolveFilePath`, so it is the same write-here-read-there round trip as the front matter.

**Two sites I deliberately left alone**, since "same call, different purpose" seemed worth stating rather than sweeping up everything `path.relative`:

- `workspace/tree.ts:30` renders a path for terminal display and already branches on `path.sep`. A host-native separator is correct there.
- `test/file-comparison.ts` is test tooling.

**The interesting part: your own tests already agreed with the fix.** `bundle.spec.ts` has two assertions — "should copy file when copy option is true" and "should use custom destName when copying" — that expect POSIX-style manifest paths. Both were failing on Windows before this change and pass now. I did not touch them. That is independent evidence that forward slashes are the intended contract, not a preference of mine.

One existing assertion did have to move with the behaviour. `timeline.spec.ts` used `path.join('..', 'archs', 'v1.json')` as its expected value, which resolves to the host separator — so it pinned the wrong value while staying green on Linux. It now expects `'../archs/v1.json'` literally, with a comment on why `path.join` is the wrong instrument for an expectation about a serialised reference.

Numbers, cli suite: **621 passed / 6 failed before, 624 passed / 5 failed after** — the two `bundle.spec.ts` tests fixed, my two added, and the remaining five are pre-existing and unrelated (workspace config, CLI e2e template mode).

On your Linux-only CI point: that is exactly why the new tests mock `path` to its `win32` implementation instead of asserting "no backslashes in the output". The latter passes on a Linux runner whether or not the fix is present. With the mock, they fail on any host if the normalisation is dropped again — so a future regression shows up in your existing pipeline rather than needing a Windows leg.

Happy to also add a `windows-latest` job if you would find that useful, though I would raise it separately rather than grow this PR.

LeighFinegold:
> Happy to also add a `windows-latest` job if you would find that useful, though I would raise it separately rather than grow this PR.

Agreed on not increasing scope, will raise in next office hours whether we agree we should go ahead and support. 

One more feedback @ppcvote  - in the last office hours, contributors talked about reducing unnecessary prose from AI on PR commits and descriptions. Hence why I recommended a one line change, but it should really apply to all files.

Can you ensure you have the latest merge of AGENTS.md. See #3012.  We would appreciate if you could rerun and cleanup the commit.

ppcvote:
Understood, and fair. Rewritten in `08c4b80` — both commit messages trimmed, PR description below replaced, and the inline comments cut to one line each. Tests and behaviour are unchanged.

LeighFinegold:
> Understood, and fair. Rewritten in `08c4b80` — both commit messages trimmed, PR description below replaced, and the inline comments cut to one line each. Tests and behaviour are unchanged.

Thanks!

LeighFinegold (approved):

PR #3014, opened 2026-08-22
feat(calm-models): add implementation for metadata diff

aamanrebello:
## Description
Same as https://github.com/finos/architecture-as-code/pull/2732 but for metadata. Only flows left to go after this!

## Type of Change

- [ ] 📚 Documentation update

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [x] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

## Checklist


calm-models/src/diff/diff.ts
```diff
@@ -72,6 +75,62 @@ export function relationshipStructureMatches(
...
+    return result;
+}
+
+function stringifyMetadata(metadata: CalmMetadataSchema): string[] {
+    if (Array.isArray(metadata)) {
+        return metadata.map((obj: Record<string, unknown>) => JSON.stringify(obj));
+    } else {
+        return [JSON.stringify(metadata)];
+    }
+}
```
rocketstack-matt, line 99: Array metadata items are matched by raw `JSON.stringify`, so two objects with identical content but reordered keys are reported as `removed`+`added` instead of `unchanged`. Every other equality check in this file (`valuesEqual`/`normalizeValue`/`canonicalKey`, and `diffControlItem`'s `diffArrays(..., { comparator: valuesEqual })`) normalizes key order first — this should too.
```suggestion
function stringifyMetadata(metadata: CalmMetadataSchema): string[] {
    if (Array.isArray(metadata)) {
        return metadata.map((obj: Record<string, unknown>) => JSON.stringify(normalizeValue(obj)));
    } else {
        return [JSON.stringify(normalizeValue(metadata))];
    }
}
```

calm-models/src/diff/diff.ts
```diff
@@ -283,21 +345,41 @@ export function diffControls(
...
+        metadataObjectsUnchanged: [],
+        metadataObjectsModified: [],
     };
 
-    return result;
-}
+    if (!Array.isArray(metadataA) && !Array.isArray(metadataB)) {
+        if (valuesEqual(metadataA, metadataB)) {
+            result.metadataObjectsUnchanged.push(metadataA);
+        } else {
```
rocketstack-matt, line 401: `archA.metadata ?? {}` at the call site means two architectures with no metadata at all both resolve to `{}`, and `valuesEqual({}, {})` is true, so every diff between metadata-less architectures reports a phantom `metadataObjectsUnchanged: [{}]` instead of `[]` (unlike the no-controls/no-ADRs case, which reports empty results). Skipping the push when the object is empty avoids the phantom entry without needing to distinguish "caller passed `{}`" from "caller had no metadata" at the call site.
```suggestion
        if (valuesEqual(metadataA, metadataB)) {
            if (Object.keys(metadataA).length > 0) {
                result.metadataObjectsUnchanged.push(metadataA);
            }
        } else {
```

calm-models/src/diff/diff.ts
```diff
@@ -72,6 +75,62 @@ export function relationshipStructureMatches(
...
+    } else {
+        return [JSON.stringify(metadata)];
+    }
+}
+
+function diffMetadataObject(metadataObjA: Record<string, unknown>, metadataObjB: Record<string, unknown>): MetadataItemDiffResult {
+    const result: MetadataItemDiffResult = {};
+
+    Object.keys(metadataObjA).forEach((key: string) => {
+        if (!(key in metadataObjB)) {
```
rocketstack-matt: `in` walks the prototype chain, not just own properties. Since CALM metadata is arbitrary key/value data, a metadata object using a key like `toString` or `constructor` will appear "present" on the other side (inherited from `Object.prototype`) even when actually absent, misclassifying a removal as a modification with a non-serializable `Function` as the value.
```suggestion
        if (!Object.hasOwn(metadataObjB, key)) {
```

calm-models/src/diff/diff.ts
```diff
@@ -72,6 +75,62 @@ export function relationshipStructureMatches(
...
+        } else if (!valuesEqual(metadataObjA[key], metadataObjB[key])) {
+            result[key] = {
+                oldValue: metadataObjA[key],
+                newValue: metadataObjB[key]
+            };
+        }
+    });
+
+    Object.keys(metadataObjB).forEach((key: string) => {
+        if (!(key in metadataObjA)) {
```
rocketstack-matt: Same prototype-chain issue as above, other direction.
```suggestion
        if (!Object.hasOwn(metadataObjA, key)) {
```

rocketstack-matt (changes requested):
Three real defects in the metadata diff logic — details inline, each with a suggested fix.

rocketstack-matt:
@aamanrebello nudge

aamanrebello:
> @aamanrebello nudge

Thanks, I have addressed all comments and ensured all checks are passing. Looking forward to the next round of review.

calm-models/src/diff/diff.ts
```diff
@@ -283,21 +345,43 @@ export function diffControls(
...
+    if (!Array.isArray(metadataA) && !Array.isArray(metadataB)) {
+        if (valuesEqual(metadataA, metadataB)) {
+            if (Object.keys(metadataA).length > 0) {
+                result.metadataObjectsUnchanged.push(normalizeValue(metadataA) as Record<string, unknown>);
+            }
+        } else {
+            result.metadataObjectsModified.push(diffMetadataObject(metadataA, metadataB));
+        }
+    } else {
+        const stringifiedArrayA: string[] = stringifyMetadata(metadataA);
```
rocketstack-matt: Two related gaps in this array-vs-object branch:

1. When one side has no metadata at all (`{}` after the `?? {}` default in the caller) and the other is a non-empty array, this still stringifies the empty object as a real item, producing a phantom entry alongside the real change: `diffMetadata({}, [{name:'x'}])` → `metadataObjectsRemoved: [{}]`. This will fire on the common case of adding metadata for the first time.
2. Object-form and array-form metadata are declared `oneOf` in the CALM 1.2 schema (genuinely distinct shapes), but content-identical cross-type metadata is reported as `unchanged`: `diffMetadata({a:1}, [{a:1}])` → `metadataObjectsUnchanged: [{a:1}]`, silently hiding a real structural change.

Both stem from collapsing both sides into a flat stringified list before diffing, which loses whether a side was absent, an empty container, or which schema form it came from. Suggest skipping empty sides entirely rather than stringifying them, and adding an explicit branch for `Array.isArray(metadataA) !== Array.isArray(metadataB)` (both non-empty) that reports the whole of A as removed and the whole of B as added, instead of content-diffing across the type boundary.

calm-models/src/diff/diff.ts
```diff
@@ -72,6 +75,62 @@ export function relationshipStructureMatches(
...
+    } else {
+        return [JSON.stringify(normalizeValue(metadata))];
+    }
+}
+
+function diffMetadataObject(metadataObjA: Record<string, unknown>, metadataObjB: Record<string, unknown>): MetadataItemDiffResult {
+    const result: MetadataItemDiffResult = {};
+
+    Object.keys(metadataObjA).forEach((key: string) => {
+        if (!Object.prototype.hasOwnProperty.call(metadataObjB, key)) {
```
rocketstack-matt, line 115: (non-blocking nit) `null` doubles as the "key is absent" sentinel here, but `null` is also a valid own metadata value (`Record<string, unknown>` permits it), so an added/removed key is indistinguishable from a key whose value changed to/from a genuine `null`: `diffMetadata({foo:'bar'}, {})` and `diffMetadata({foo:'bar'}, {foo:null})` both produce `{foo:{oldValue:'bar',newValue:null}}`. Worth a presence flag if a consumer ever needs to tell these apart — explicit `null` values are probably rare in practice.

rocketstack-matt (changes requested):
Two new correctness issues in the metadata diff logic (previous round's 4 comments are all fixed and resolved — thanks).

calm-models/src/diff/diff.ts
aamanrebello: Addressed.

calm-models/src/diff/diff.ts
aamanrebello: Addressed.

rocketstack-matt (approved):

PR #3020, opened 2026-08-25
feat(calm-hub): add GLOBAL_ADMIN-gated delete endpoints for all resource types

willosborne:
## Description

Adds a `DELETE` endpoint for every CALM Hub resource type that didn't already have one (Architecture, Pattern, Flow, Standard, Interface, Timeline, Adr, Decorator, Control Requirement, Control Configuration), gated behind `GLOBAL_ADMIN`.

**Delete semantics**: delete removes the *whole* resource — its header **and every version** — not just the latest version or the header alone. The shared `MongoVersionDocumentStore`/`NitriteVersionDocumentStore` helpers gain a new `deleteResource(namespace, id)` for this; it's distinct from the existing `deleteHeader`, which is internal-only compensation logic used when a first version write fails and deliberately never touches the version collection.

**Control requirements caveat**: a control requirement that still has configurations underneath it refuses to delete (`409 Conflict`) rather than cascading. Delete the configurations first, then the requirement. Configurations themselves delete normally.

Decorator isn't part of the versioned-document family (one doc per namespace, embedded array), so its delete is a bespoke pull/list-removal rather than a `deleteResource` call.

Audit logging picks up all the new endpoints automatically via the existing generic path-param resolution, except Control's two new methods, which needed one manual addition to `AuditRequestFilter`'s per-method dispatch (Control is the one resource that doesn't use the generic resolution path).

Also updates `PERMISSIONS.md` (content-resource deletion is deliberately `GLOBAL admin`-only, distinct from namespace/domain admin) and two ADRs whose "nothing is ever deleted" premise this change reverses, and adds Mongo + Nitrite integration test coverage for Architecture (representative simple case) and Control (the cascade-refuse case).

## Review notes

This PR is best reviewed commit by commit:
1. Store interface method signatures
2. Mongo implementations + tests
3. Nitrite implementations + tests
4. New `ControlHasConfigurationsException` + audit filter wiring + audit test
5. REST resource endpoints + tests
6. Docs (PERMISSIONS.md + ADRs 0001/0003)
7. Integration tests (Architecture + Control, Mongo + Nitrite)

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

Full suite passes: `../mvnw clean verify -Ddependency-check.skip=true` — 2835 tests, 0 failures, JaCoCo coverage gate satisfied. `../mvnw -P integration verify` also passes in full (Docker/TestContainers). Also manually verified end-to-end against a running standalone instance: header+versions are genuinely gone from storage (not just the header), repeat delete returns 404, invalid namespace returns 400, audit log entries are recorded, and the control-requirement-with-configurations 409-refuse-then-succeed sequence behaves as intended.

## Checklist

### DIFF
--- a/calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
+++ b/calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
@@ -8,6 +8,7 @@ import jakarta.validation.Valid;
 import jakarta.validation.constraints.Pattern;
 import jakarta.ws.rs.BeanParam;
 import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.DELETE;
 import jakarta.ws.rs.GET;
 import jakarta.ws.rs.POST;
 import jakarta.ws.rs.PUT;
@@ -299,6 +300,30 @@ public class ArchitectureResource {
             logger.error("Failed to serialise timeline for architecture [{}] in namespace [{}]", architectureId, namespace, e);
             return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity("Failed to build timeline for architecture: " + architectureId).build();
         }
+    }
+
+    @DELETE
+    @Path("{namespace}/architectures/{architectureId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(
+            summary = "Delete an architecture",
+            description = "Deletes an architecture and all of its versions from the given namespace. Requires global admin privilege."
+    )
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteArchitecture(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("architectureId") int architectureId
+    ) {
+        try {
+            store.deleteArchitecture(namespace, architectureId);
+        } catch (NamespaceNotFoundException e) {
+            logger.error("Invalid namespace [{}] when deleting architecture", namespace, e);
+            return CalmResourceErrorResponses.invalidNamespaceResponse(namespace);
+        } catch (ArchitectureNotFoundException e) {
+            logger.error("Invalid architecture [{}] when deleting architecture", architectureId, e);
+            return invalidArchitectureResponse(architectureId);
+        }
+        return Response.noContent().build();
     }
 
     private Response architectureWithLocationResponse(Architecture architecture) throws URISyntaxException {
--- a/calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
+++ b/calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
@@ -33,6 +33,7 @@ import org.finos.calm.security.CalmHubScopes;
 import org.finos.calm.services.ArchitectureTimelineService;
 import org.finos.calm.services.CustomIdEnrichmentService;
 import org.finos.calm.store.ArchitectureStore;
+import org.finos.calm.store.ResourceMappingStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -57,6 +58,7 @@ public class ArchitectureResource {
     private final ArchitectureStore store;
     private final ArchitectureTimelineService timelineService;
     private final CustomIdEnrichmentService customIds;
+    private final ResourceMappingStore mappingStore;
 
     private final Logger logger = LoggerFactory.getLogger(ArchitectureResource.class);
 
@@ -65,10 +67,11 @@ public class ArchitectureResource {
 
     @Inject
     public ArchitectureResource(ArchitectureStore store, ArchitectureTimelineService timelineService,
-                                CustomIdEnrichmentService customIds) {
+                                CustomIdEnrichmentService customIds, ResourceMappingStore mappingStore) {
         this.store = store;
         this.timelineService = timelineService;
         this.customIds = customIds;
+        this.mappingStore = mappingStore;
     }
 
     /**
@@ -323,6 +326,7 @@ public class ArchitectureResource {
             logger.error("Invalid architecture [{}] when deleting architecture", architectureId, e);
             return invalidArchitectureResponse(architectureId);
         }
+        MappingCleanup.deleteMapping(mappingStore, logger, namespace, ResourceType.ARCHITECTURE, architectureId);
         return Response.noContent().build();
     }
 
--- a/calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
+++ b/calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
@@ -274,6 +274,30 @@ public class FlowResource {
         }
     }
 
+    @DELETE
+    @Path("{namespace}/flows/{flowId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(
+            summary = "Delete a flow",
+            description = "Deletes a flow and all of its versions from the given namespace. Requires global admin privilege."
+    )
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteFlow(
+            @PathParam("namespace") @Pattern(regexp= NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("flowId") int flowId
+    ) {
+        try {
+            store.deleteFlow(namespace, flowId);
+        } catch (NamespaceNotFoundException e) {
+            logger.error("Invalid namespace [{}] when deleting flow", namespace, e);
+            return CalmResourceErrorResponses.invalidNamespaceResponse(namespace);
+        } catch (FlowNotFoundException e) {
+            logger.error("Invalid flow [{}] when deleting flow", flowId, e);
+            return invalidFlowResponse(flowId);
+        }
+        return Response.noContent().build();
+    }
+
     private Response flowWithLocationResponse(Flow flow) throws URISyntaxException {
         return Response.created(new URI("/api/calm/namespaces/" + flow.getNamespace() + "/flows/" + flow.getId() + "/versions/" + flow.getDotVersion())).build();
     }
--- a/calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
+++ b/calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
@@ -23,6 +23,7 @@ import org.finos.calm.domain.flow.CreateFlowRequest;
 import org.finos.calm.security.CalmHubScopes;
 import org.finos.calm.services.CustomIdEnrichmentService;
 import org.finos.calm.store.FlowStore;
+import org.finos.calm.store.ResourceMappingStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,6 +43,7 @@ public class FlowResource {
 
     private final FlowStore store;
     private final CustomIdEnrichmentService customIds;
+    private final ResourceMappingStore mappingStore;
 
     private final Logger logger = LoggerFactory.getLogger(FlowResource.class);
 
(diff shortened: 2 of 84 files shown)


calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
```diff
@@ -301,6 +302,30 @@ public Response getArchitectureTimeline(
...
+            summary = "Delete an architecture",
+            description = "Deletes an architecture and all of its versions from the given namespace. Requires global admin privilege."
+    )
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteArchitecture(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("architectureId") int architectureId
+    ) {
+        try {
+            store.deleteArchitecture(namespace, architectureId);
```
rocketstack-matt, line 321: This deletes the resource but leaves its `ResourceMappingStore` entry behind — `deleteMapping` is never called (it's currently only invoked from `MappingControllerService`'s create-rollback path). Recreating an architecture under the same custom-id will hit `DuplicateMappingException`, and the custom-id route keeps resolving to a numeric ID that no longer exists. Same gap in FlowResource, InterfaceResource, PatternResource, StandardResource — worth deleting the mapping (tolerating `MappingNotFoundException` if none exists) alongside the resource in all five.

calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
```diff
@@ -274,6 +274,30 @@ public Response updateVersionedFlow(
...
+            summary = "Delete a flow",
+            description = "Deletes a flow and all of its versions from the given namespace. Requires global admin privilege."
+    )
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteFlow(
+            @PathParam("namespace") @Pattern(regexp= NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("flowId") int flowId
+    ) {
+        try {
+            store.deleteFlow(namespace, flowId);
```
rocketstack-matt, line 293: Same orphaned-mapping gap as ArchitectureResource — see that comment.

calm-hub/src/main/java/org/finos/calm/resources/InterfaceResource.java
```diff
@@ -152,6 +152,26 @@ public Response createInterfaceForVersion(
...
+    @DELETE
+    @Path("{namespace}/interfaces/{interfaceId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteInterface(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("interfaceId") Integer interfaceId
+    ) {
+        try {
+            interfaceStore.deleteInterface(namespace, interfaceId);
```
rocketstack-matt, line 167: Same orphaned-mapping gap as ArchitectureResource — see that comment.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
```diff
@@ -248,6 +248,30 @@ public Response updateVersionedPattern(
...
+            summary = "Delete a pattern",
+            description = "Deletes a pattern and all of its versions from the given namespace. Requires global admin privilege."
+    )
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deletePattern(
+            @PathParam("namespace") @jakarta.validation.constraints.Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("patternId") int patternId
+    ) {
+        try {
+            store.deletePattern(namespace, patternId);
```
rocketstack-matt, line 267: Same orphaned-mapping gap as ArchitectureResource — see that comment.

calm-hub/src/main/java/org/finos/calm/resources/StandardResource.java
```diff
@@ -142,6 +142,26 @@ public Response createStandardForVersion(
...
+    @DELETE
+    @Path("{namespace}/standards/{standardId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @PermissionsAllowed(CalmHubScopes.GLOBAL_ADMIN)
+    public Response deleteStandard(
+            @PathParam("namespace") @Pattern(regexp = NAMESPACE_REGEX, message = NAMESPACE_MESSAGE) String namespace,
+            @PathParam("standardId") Integer standardId
+    ) {
+        try {
+            standardStore.deleteStandard(namespace, standardId);
```
rocketstack-matt, line 157: Same orphaned-mapping gap as ArchitectureResource — see that comment.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoControlStore.java
```diff
@@ -194,6 +195,29 @@ public void createConfigurationForVersion(String domain, int controlId, int conf
         }
     }
 
+    @Override
+    public void deleteControlRequirement(String domain, int controlId) throws DomainNotFoundException, ControlNotFoundException, ControlHasConfigurationsException {
+        requireControl(domain, controlId);
+
+        int configurationCount = configurationStore.countHeaders(configurationNamespace(domain, controlId));
```
rocketstack-matt, line 215: `countHeaders` then `deleteResource` isn't atomic — a configuration created between the two calls (via `createConfigurationForVersion`) survives under a deleted control. Same pattern in `NitriteControlStore`. Not a new problem for this codebase (no transactions used anywhere in `calm-hub/store`), but this endpoint's whole purpose is a hard "refuse rather than cascade" guarantee, so worth a decision: is that race acceptable, or does it need a lock/transaction here specifically?

calm-hub/src/main/java/org/finos/calm/store/util/MongoVersionDocumentStore.java
```diff
@@ -175,6 +177,31 @@ public void deleteHeader(String namespace, int resourceId) {
...
+     * at this (namespace, resourceId) — lets the caller distinguish "already gone" from
+     * "removed".
+     */
+    public boolean deleteResource(String namespace, int resourceId) {
+        try {
+            versionCollection.deleteMany(headerFilter(namespace, resourceId));
+            DeleteResult result = headerCollection.deleteOne(headerFilter(namespace, resourceId));
+            return result.getDeletedCount() > 0;
+        } catch (MongoException e) {
+            LOG.error("Failed to delete resource [namespace={}, {}={}]", namespace, idField, resourceId, e);
```
rocketstack-matt, line 200: Nit: versions are deleted before the header, so if `versionCollection.deleteMany` succeeds but `headerCollection.deleteOne` then throws, the header survives with zero versions — the same "stranded header" state this PR's ADR update calls out as a problem elsewhere. Very unlikely (two sequential local Mongo calls), non-blocking.

rocketstack-matt (changes requested):
Solid, well-tested implementation of the delete pattern across every resource type. One real pre-merge defect: the five custom-id-mapped resource types (Architecture, Flow, Interface, Pattern, Standard) delete the resource but never clean up their `ResourceMappingStore` entry, so recreating under the same custom-id throws `DuplicateMappingException`. Also flagging a non-atomic check-then-act race in the control cascade-refuse logic for discussion, plus a non-blocking nit.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoControlStore.java
willosborne: interesting point. a simple lock here would help, but i guess without a db level lock this could cause issues across instances

calm-hub/src/main/java/org/finos/calm/resources/ArchitectureResource.java
willosborne: Confirmed, fixed in b3e1e5d0 — added `ResourceMappingStore.deleteMappingByNumericId` (no-op if no mapping exists) and wired it into all five delete endpoints via a shared `MappingCleanup` helper.

calm-hub/src/main/java/org/finos/calm/resources/FlowResource.java
willosborne: Fixed in b3e1e5d0, same fix as ArchitectureResource.

calm-hub/src/main/java/org/finos/calm/resources/InterfaceResource.java
willosborne: Fixed in b3e1e5d0, same fix as ArchitectureResource.

calm-hub/src/main/java/org/finos/calm/resources/PatternResource.java
willosborne: Fixed in b3e1e5d0, same fix as ArchitectureResource.

calm-hub/src/main/java/org/finos/calm/resources/StandardResource.java
willosborne: Fixed in b3e1e5d0, same fix as ArchitectureResource.

calm-hub/src/main/java/org/finos/calm/store/mongo/MongoControlStore.java
willosborne: Documented as an accepted, known race in dfcf92e0 — not adding a lock, since a Nitrite-only fix wouldn't extend the guarantee to Mongo, and the failure mode (one configuration outliving the requirement it belonged to) is a data-hygiene issue, discoverable and cleanable, not a correctness or security one.

calm-hub/src/main/java/org/finos/calm/store/util/MongoVersionDocumentStore.java
willosborne: Deliberate, not fixing — the alternative (header first, then versions) trades this for the opposite and worse failure: orphaned version documents with no header, which nothing will ever list or read again. A zero-version header is at least visible and consistent with the versionCount drift ADR 0001 already accepts elsewhere.

calm-hub/src/main/java/org/finos/calm/resources/MappingCleanup.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    static void deleteMapping(ResourceMappingStore mappingStore, Logger logger,
+                               String namespace, ResourceType type, int numericId) {
+        try {
+            mappingStore.deleteMappingByNumericId(namespace, type, numericId);
+        } catch (NamespaceNotFoundException e) {
+            logger.warn("Could not clean up the mapping for {} [{}] in namespace [{}] after delete "
+                    + "— the namespace no longer exists", type, numericId, namespace, e);
+        }
```
rocketstack-matt, line 41: This only catches the checked `NamespaceNotFoundException` — the javadoc above promises cleanup failures are logged and swallowed rather than failing an already-successful delete, but a runtime failure from the mapping store (`MongoException`, a Nitrite lock/DB error) isn't caught here and would surface as an unhandled 500 for a delete that already succeeded.
```suggestion
        } catch (NamespaceNotFoundException e) {
            logger.warn("Could not clean up the mapping for {} [{}] in namespace [{}] after delete "
                    + "— the namespace no longer exists", type, numericId, namespace, e);
        } catch (RuntimeException e) {
            logger.warn("Could not clean up the mapping for {} [{}] in namespace [{}] after delete", type, numericId, namespace, e);
        }
```

rocketstack-matt (commented):
One edge-case finding below; everything else — the fixes for the earlier review round, the idiom audit, and a Copilot CLI pass — checks out.

calm-hub/src/main/java/org/finos/calm/resources/MappingCleanup.java
willosborne: Confirmed, fixed in adde2160 — added a `catch (RuntimeException e)` branch alongside the existing `NamespaceNotFoundException` one.

rocketstack-matt (approved):
All prior review feedback addressed and verified — the mapping-cleanup runtime-exception fix in adde2160 is correct and tested. Nothing outstanding.

PR #3013, opened 2026-08-22
Enhance directUrlDocumentLoader support for End User Specific IDP

jimthompson5802:
## Description

Closes #2975 

This PR adds support for authenticated direct `http(s)` document loading in the CLI and shared document loader stack. It introduces a separate `directUrlAuth` configuration path for protected remote schema and document fetches, wires that module into the CLI document loader configuration, and improves error handling and debug logging so authentication failures are clearer without exposing sensitive header values.

The intent of this PR is to provide an interface specification that lets an end-user organization supply a custom authentication module for its own Identity Provider. That module remains outside the CALM project and is fully owned by the end-user organization.

For TLS and certificate validation, this PR relies on Node.js runtime behavior rather than adding CALM-specific certificate handling. In environments that use private or enterprise certificate chains, trust can be extended with `NODE_EXTRA_CA_CERTS`. In environments where certificate verification must be bypassed for local or controlled use cases, Node also supports `NODE_TLS_REJECT_UNAUTHORIZED=0`. This keeps certificate handling aligned with Node.js configuration and leaves trust and validation policy under deployment control.

**IMPORTANT NOTE**:  To view how an end user organization integrates with the changes in this PR, see this repo: https://github.com/jimthompson5802/calm-web-repo.  The README describes the three test scenarios used for testing the PR.  There are sample code and configurations that show how the end user provided authentication module is used and integrates with `calm cli` `DirectUrlDocumentLoader`.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📚 Documentation update
- [ ] 🎨 Code style/formatting changes
- [ ] ♻️ Refactoring (no functional changes)
- [ ] ⚡ Performance improvements
- [x] ✅ Test additions or updates
- [ ] 🔧 Chore (maintenance, dependencies, CI, etc.)

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] CALM Hub UI (`calm-hub-ui/`)
- [ ] CALM Server (`calm-server/`)
- [ ] CALM Widgets (`calm-widgets/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
[...]

PR #3035, opened 2026-08-28
feat(cli): add directUrlAuthPlugin support for authenticated direct-URL document loading

### LINKED ISSUE #2975: Authentication Support for Direct URL Document Loading
### Target Project:

`architecture-as-code` monorepo, primarily affecting:

* `@finos/calm-cli`
* `@finos/calm-shared`

### Description of Feature:

Add authentication support to the `DirectUrlDocumentLoader` so CALM can retrieve **read-only CALM artifacts** from HTTP/HTTPS repositories that require authentication.

The primary goal is to let organizations use CALM with artifacts that are already stored in existing authenticated repositories, without first moving those artifacts into CALM Hub.

The two main use cases are:

* Retrieve a CALM artifact from an authenticated repository
* Use `calm validate` on artifacts stored in an authenticated repository

This feature is limited to **read-only retrieval of CALM artifacts** through `DirectUrlDocumentLoader`. It does not propose CALM Hub features such as publishing, visualization, versioning, diff, timeline, or repository management.

Authentication for direct URL access should remain separate from the existing CALM Hub authentication path. If direct URL authentication is not configured, `DirectUrlDocumentLoader` should continue to behave exactly as it does today.

Any support added to the `architecture-as-code` repository should be based on **industry-standard authentication mechanisms and extension points**, not on embedding end-user organization Identity Provider (IDP) implementations in this repository. Organization-specific IDP implementations should remain outside the `architecture-as-code` repo.

Related Issue #2944

### User Stories:

As a CALM user, I want to retrieve a CALM artifact from an authenticated repository so that I can work with architecture content already stored in my organization.

As a CALM user, I want to run `calm validate` against artifacts stored in an authenticated repository so that I can validate those artifacts without copying them to another location first.

As an enterprise user, I want CALM CLI to work with my organization’s existing authenticated repository so that adopting CALM does not require changing existing repository or access-control arrangements.

As an automation or CI/CD user, I want `calm validate` to retrieve CALM artifacts from a protected repository using non-interactive authentication so that validation can run in pipelines.

As an interactive CLI user, I want CALM CLI to retrieve protected CALM artifacts using my organization’s authentication flow so that I can validate or inspect those artifacts from my local environment.

As an organization adopting CALM Hub while continuing to use an existing repository, I want direct URL authentication and CALM Hub authentication to remain independently configurable so that each can use its own credentials and access model.

### Current Limitations:

`DirectUrlDocumentLoader` currently performs unauthenticated HTTP requests. As a result, CALM artifacts stored in authenticated repositories cannot be retrieved through direct URLs.

This prevents organizations from using CALM directly against protected repositories that already contain relevant artifacts unless they make those artifacts public or relocate them.

There is currently no direct way to use `calm validate` on artifacts that are only available through authenticated direct URLs.

The existing `authPluginPath` mechanism is associated with CALM Hub authentication and does not address the separate need to retrieve protected artifacts through `DirectUrlDocumentLoader`.

### Proposed Implementation:

Introduce authentication support specifically for `DirectUrlDocumentLoader`.

The CLI should allow users to configure authentication for direct URL access independently from CALM Hub authentication.

When configured, `DirectUrlDocumentLoader` should apply the necessary authentication to retrieve protected CALM artifacts over HTTP/HTTPS.

This capability should support the two main use cases in this feature request:

* Retrieving a CALM artifact from an authenticated repository
* Running `calm validate` on artifacts stored in an authenticated repository

The feature should remain read-only. It should only enable retrieval of artifacts needed by CALM CLI and should not introduce repository write operations.

The implementation should preserve backward compatibility. When direct URL authentication is not configured, existing unauthenticated direct URL behavior should remain unchanged.

The implementation should also preserve separation from the existing CALM Hub authentication path so that authenticated direct URL retrieval and CALM Hub access can be configured independently.

Any changes made in the `architecture-as-code` repo should focus on industry-standard protocol support and stable extension points. End-user organization IDP implementations, proprietary authentication logic, and vendor-specific integrations should be kept outside this repository.

### Alternatives Considered:

**Require users to make protected artifacts public.** This would weaken existing access controls and is not acceptable for many enterprise repositories.

**Require users to migrate artifacts into CALM Hub first.** This creates an unnecessary adoption barrier for organizations that already store CALM artifacts in an authenticated repository.

**Reuse the existing CALM Hub authentication mechanism directly.** This would couple two different access scenarios that may need different credentials, policies, or operational ownership.

**Continue supporting only unauthenticated direct URLs.** This does not address the main adoption problem for organizations whose repositories already require authentication.

**Use existing authenticated methods to download CALM artifacts and run `calm validate` with `url-mapping`, if needed, on the downloaded files.**  Complicates the workflow.  This would be the fallback if this feature request does not workout.

### Testing Strategy:

Unit tests should verify that authentication can be applied to `DirectUrlDocumentLoader` requests and that existing unauthenticated behavior still works when no configuration is provided.

`DirectUrlDocumentLoader` tests should verify that protected CALM artifacts can be retrieved when authentication is configured and cannot be retrieved when required authentication is missing or invalid.

Integration tests should verify at minimum:

1. A CALM artifact stored in an authenticated repository can be retrieved through `DirectUrlDocumentLoader`.
2. `calm validate` can validate an artifact stored in an authenticated repository.
3. Direct URL authentication and CALM Hub authentication can be configured independently in the same execution.
4. Existing configurations without direct URL authentication continue to operate unchanged.

Update the CALM CLI documentation to describe how authentication can be configured for `DirectUrlDocumentLoader`.

Document the primary supported use cases:

* Retrieve a CALM artifact from an authenticated repository
* Use `calm validate` on artifacts stored in an authenticated repository

Provide examples showing how direct URL authentication is configured separately from CALM Hub authentication.

Document that this feature is limited to read-only retrieval of CALM artifacts and does not add CALM Hub-style repository capabilities.

Document that organization-specific IDP implementations must live outside the `architecture-as-code` repository, and that changes within this repository should remain standards-based rather than encoding end-user-specific implementation details.

### Implementation Checklist:

* [ ] Design reviewed and approved
* [ ] Define direct URL authentication configuration
* [ ] Add authentication support to `DirectUrlDocumentLoader`
* [ ] Preserve independent CALM Hub authentication behavior
* [ ] Implementation completed
* [ ] Unit tests written and passing
* [ ] Integration tests written and passing
* [ ] Backward compatibility verified
* [ ] Documentation updated
* [ ] Relevant workflows updated (if needed)

### Additional Context:

The main purpose of this feature is to reduce adoption friction for organizations that already store CALM artifacts in authenticated repositories.

CALM should be able to retrieve those artifacts in a read-only way through `DirectUrlDocumentLoader`, and `calm validate` should be able to operate on them without requiring migration into CALM Hub.

This keeps the scope focused on authenticated retrieval of CALM artifacts rather than expanding into broader repository-management capabilities.

It should also keep a clean boundary between the public CALM codebase and end-user organization authentication implementations. The `architecture-as-code` repo should add standards-based support for authenticated retrieval, while proprietary or organization-specific IDP implementations should remain in external repositories or packages controlled by the end user organization.

<div align="center">
  <h2>Illustrative Example</h2>
</div>

```json
{
  "calmHubUrl": "https://calm.example.com",
  "authPluginPath": "~/company-calmhub-plugin.js",
  "directUrlAuth": {
    "module": "~/plugins/acme-direct-url-auth.js",
    "options": {
      "tokenUrl": "https://idp.acme.example.com/oauth/token",
      "clientId": "calm-direct-url",
      "clientSecret": "${ACME_IDP_CLIENT_SECRET}",
      "scopes": ["calm:read", "calm:documents"],
      "headerName": "Authorization",
      "headerPrefix": "Bearer "
    }
  }
}
```

rocketstack-matt:
## Description

This is a rebase of #3013 (@jimthompson5802, `iss2975-idp-directurl`) directly onto `main`, offered as a smaller-diff alternative to merging `main` into that branch. #3022 merged after #3013 was opened and restructured `shared/src/document-loader/` for a browser-safe entry point, which conflicts with #3013's changes — merging `main` into the branch pulls in every unrelated commit since Jim's last sync, producing a large diff. This PR instead squashes #3013's commits and resolves the conflicts on top of current `main`, so the diff is scoped to just this feature (15 files, in line with #3013's original 14).

This is offered as an option, not a replacement — @jimthompson5802 may prefer to continue with #3013 directly (see the smaller conflict-resolution PR at jimthompson5802#3), or use this one instead. Either way the resulting code is identical and fully tested.

Adds a `directUrlAuthPlugin` configuration path so an end-user organization can supply a custom authentication module for its own Identity Provider when loading documents/schemas over direct `http(s)` URLs. That module stays outside the CALM project and is fully owned by the end-user organization. TLS/certificate validation relies on Node.js runtime behaviour (`NODE_EXTRA_CA_CERTS`, `NODE_TLS_REJECT_UNAUTHORIZED`) rather than CALM-specific handling.

Closes #2975.

### Conflicts with #3022 and how they're resolved
- #3022 moved `buildDocumentLoader` out of `document-loader.ts` (now interface-only, for the browser entry point) into `node-document-loader.ts`. The `directUrlAuthPlugin` wiring moves there with it.
- #3022 replaced `net.isIP` with a browser-safe `ipLiteralVersion` helper and added an `assertResponseOrigin` SSRF check in `DirectUrlDocumentLoader`; both are kept alongside this branch's auth-header/debug-logging request config — dropping either would reintroduce a Node-only import into the browser-safe graph, or silently drop the SSRF guard.
- #3022 relocated most of `document-loader.spec.ts`'s tests into `node-document-loader.spec.ts`; #3013's `buildDocumentLoader` tests (incl. `directUrlAuthPlugin` coverage) move there too, with the `allowedRemoteHosts` assertion updated for the loader's new 4-arg constructor signature.

## Type of Change
- [x] 📚 Documentation update

## Affected Components
- [x] CLI (`cli/`)
- [x] Shared (`shared/`)

## Testing
- [x] All existing tests pass

`shared/`: `npx vitest run src/document-loader/` — 13 test files, 165 tests, all passing.
`cli/`: `npx vitest run src/cli.spec.ts src/cli-config.spec.ts src/command-helpers/validate.spec.ts` — 3 test files, 143 tests, all passing.
`tsc --noEmit` on both workspaces shows only pre-existing, unrelated failures already present on plain `main` (a stray `Axios`/`AxiosInstance` type mismatch in `shared`, and the ~60 `exports`-resolution errors on `cli/tsconfig.json` that #3022's description notes are pre-existing when invoking `tsc` directly rather than through the project's real build).

## Checklist


shared/src/document-loader/direct-url-document-loader.spec.ts
```diff
@@ -2,6 +2,7 @@ import axios, { Axios } from 'axios';
 import AxiosMockAdapter from 'axios-mock-adapter';
 import { DirectUrlDocumentLoader } from './direct-url-document-loader';
 import { DocumentLoadError } from './document-loader';
+import type { Logger } from '../logger';
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 3653d20d — added an `afterEach(() => vi.restoreAllMocks())` to the describe block so the `initLogger` spy doesn't leak past the two tests that set it.

shared/src/document-loader/direct-url-document-loader.ts
```diff
@@ -1,9 +1,10 @@
-import axios, { Axios } from 'axios';
+import axios, { Axios, AxiosRequestConfig } from 'axios';
 import { ipLiteralVersion } from '../util/ip-literal.js';
 import { SchemaDirectory } from '../schema-directory';
 import { DocumentLoader, DocumentLoadError, assertJsonObject } from './document-loader';
 import { assertResponseOrigin } from './response-origin.js';
 import { Logger, initLogger } from '../logger';
+import { DirectUrlAuthPlugin } from '../auth/direct-url-auth-plugin';
```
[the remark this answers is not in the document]
rocketstack-matt: Fixed in 3653d20d — switched to `import type ... .js`. Worth noting: this file isn't part of the browser-safe entry point graph (it's only reached through node-document-loader.ts), so the ESM-bundle-fragility concern doesn't apply here, but the import-type form is correct regardless since DirectUrlAuthPlugin is a pure interface with no runtime export.

cli/src/command-helpers/validate.ts
```diff
@@ -69,11 +69,36 @@ export async function runValidate(options: ValidateOptions) {
         const message = err instanceof Error ? err.message : String(err);
         const stack = err instanceof Error ? err.stack : undefined;
         logger.error('An error occurred while validating: ' + message);
+        if (options.verbose) {
+            for (const causeMessage of formatErrorCauseChain(err)) {
+                logger.error(causeMessage);
+            }
+        }
```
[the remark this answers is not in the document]
rocketstack-matt: Declining this one. The underlying concern is real (an auth-plugin's own thrown Error could embed a secret in its message), but the suggested fix doesn't change anything here: --verbose is what sets the logger to debug level in this codebase, so logger.debug and logger.error print identically once options.verbose is true — there's no separate quieter channel to route this to. This is inherent to letting user-supplied auth-plugin code throw errors that get surfaced for debugging (see the open discussion on issue #2975 about defensive handling of untrusted plugin code), not something this specific logging call introduces or can fix in isolation.

jimthompson5802:
@rocketstack-matt   I've pulled down this branch and ran all my local the scenarios described here: https://github.com/jimthompson5802/calm-web-repo#testbed-for-calm-directurldocumentloader-authentication.

All test scenarios passed.

jimthompson5802:
Looking for guidance on this situation.  If the `DirectUrlDocumentLoader` has to retrieve documents from a mix of authenticated and unauthenticated repositories, I'm thinking the user provided plugin `directUrlAuth()` should return:

For end user repository requiring authentication
```typescript
{Authorization: Bearer <token>
```

and for all other repositories it returns something like
```typescript
{}
```

rocketstack-matt:
@markscott-ms could you take a look at this PR as I am now a co-author with @jimthompson5802

markscott-ms:
@jimthompson5802 @rocketstack-matt  ack, I will take a look through this.

jimthompson5802:
Confirmed with a mix of authenticated and unauthenticated repositories for Direct URL Document Loader that an `Authorization` header is added to the unauthenticated repository.  

With this local test setup:
my-calm-repo-mixed

- `my-calm.repo` requires authentication
- `your-calm.repo` does not require authentication

This command `calm validate -a http://your-calm.repo:8080/architectures/calm-1.json -f pretty -v` shows an `Authorization` header being sent to the repository that does not require authentication.
```
debug [direct-url-document-loader]:    Starting Request: {
  "method": "get",
  "url": "http://your-calm.repo:8080/architectures/calm-1.json",
  "baseURL": "http://your-calm.repo:8080",
  "path": "/architectures/calm-1.json",
  "timeout": 10000,
  "maxRedirects": 0,
  "allowAbsoluteUrls": false,
  "headers": {
    "Accept": "application/json, text/plain, */*",
    "Content-Type": "application/json"
  },
  "authHeadersPresent": true,                             <========
  "authHeaderNames": [  
    "Authorization"
  ]
}
debug [direct-url-document-loader]:    Response: {
  "status": 200,
  "statusText": "OK",
  "url": "http://your-calm.repo:8080/architectures/calm-1.json"
}
```
While the retrieval works in this test setup, in general, I don't think there is a guarantee that an end-user organization repository will work the same.

With this change to the end organization provided plug-in function `getAuthHeaders()`,
```typescript
// code fragment
const AUTHORIZED_URL = 'https://my-calm.repo:8443/';

export default class DirectUrlAuthPlugin {
//.....
    async getAuthHeaders(url: string, _requestBody: unknown): Promise<Record<string, string>> {
        if (!url.startsWith(AUTHORIZED_URL)) {
            return {};
        }

        return {
            Authorization: `Bearer ${await this.getAccessToken()}`
        };
    }
```

The same command `calm validate -a http://your-calm.repo:8080/architectures/calm-1.json -f pretty -v` shows no `Authorization` header
```
debug [direct-url-document-loader]:    Starting Request: {
  "method": "get",
  "url": "http://your-calm.repo:8080/architectures/calm-1.json",
  "baseURL": "http://your-calm.repo:8080",
  "path": "/architectures/calm-1.json",
  "timeout": 10000,
  "maxRedirects": 0,
  "allowAbsoluteUrls": false,
  "headers": {
    "Accept": "application/json, text/plain, */*",
    "Content-Type": "application/json"
  },
  "authHeadersPresent": false,                        <====
  "authHeaderNames": []
}
debug [direct-url-document-loader]:    Response: {
  "status": 200,
  "statusText": "OK",
  "url": "http://your-calm.repo:8080/architectures/calm-1.json"
}
```

And if a requests needs to extract from an authenticated and unauthenticated in the same calm validate command such as if the pattern referes to a standard in an **authenticated** repository and another standard is in an **unauthenticated** repository such as this fragment of the pattern `company-base-mixed-pattern.json`:
```json
 "properties": {
    "nodes": {
      "type": "array",
      "description": "All nodes must conform to the Company Node Standard.",
      "items": {
        "$ref": "https://my-calm.repo:8443/standards/company-node-standard.json"
      }
    },
    "relationships": {
      "type": "array",
      "description": "All relationships must conform to the Company Relationship Standard.",
      "items": {
        "$ref": "http://your-calm.repo:8080/standards/company-relationship-standard.json"
      }
    }
  }
```
we see for 
```
calm validate -p https://my-calm.repo:8443/patterns/company-base-mixed-pattern.json \
  -a https://my-calm.repo:8443/architectures/generated-webapp.json \
  -f pretty -v
```

this in the debug log:
```
debug [file-system-document-loader]:    Document with id [https://my-calm.repo:8443/standards/company-node-standard.json] and type [schema] was requested but not loaded at initialisation. 
            File system document loader can only load at startup. Please ensure the schemas are present on your directory path or use CALMHub.
debug [direct-url-document-loader]:    Starting Request: {
  "method": "get",
  "url": "https://my-calm.repo:8443/standards/company-node-standard.json",
  "baseURL": "https://my-calm.repo:8443",
  "path": "/standards/company-node-standard.json",
  "timeout": 10000,
  "maxRedirects": 0,
  "allowAbsoluteUrls": false,
  "headers": {
    "Accept": "application/json, text/plain, */*",
    "Content-Type": "application/json"
  },
  "authHeadersPresent": true,
  "authHeaderNames": [
    "Authorization"
  ]
}
debug [direct-url-document-loader]:    Response: {
  "status": 200,
  "statusText": "OK",
  "url": "https://my-calm.repo:8443/standards/company-node-standard.json"
}
debug [schema-directory]:    Storing document with ID https://my-calm.repo:8443/standards/company-node-standard.json of type schema.

<<<<<<<<<<REMOVED EXTRANEOUS DEBUG MESSAGES>>>>>>>>>>>>>

debug [file-system-document-loader]:    Document ID http://your-calm.repo:8080/standards/company-relationship-standard.json does not exist in file system, cannot load.
debug [file-system-document-loader]:    Document with id [http://your-calm.repo:8080/standards/company-relationship-standard.json] and type [schema] was requested but not loaded at initialisation. 
            File system document loader can only load at startup. Please ensure the schemas are present on your directory path or use CALMHub.
debug [direct-url-document-loader]:    Starting Request: {
  "method": "get",
  "url": "http://your-calm.repo:8080/standards/company-relationship-standard.json",
  "baseURL": "http://your-calm.repo:8080",
  "path": "/standards/company-relationship-standard.json",
  "timeout": 10000,
  "maxRedirects": 0,
  "allowAbsoluteUrls": false,
  "headers": {
    "Accept": "application/json, text/plain, */*",
    "Content-Type": "application/json"
  },
  "authHeadersPresent": false,
  "authHeaderNames": []
}
debug [direct-url-document-loader]:    Response: {
  "status": 200,
  "statusText": "OK",
  "url": "http://your-calm.repo:8080/standards/company-relationship-standard.json"
}

<<<<<<<<<<REMOVED EXTRANEOUS DEBUG MESSAGES>>>>>>>>>>>>>

debug [json-schema-validator]:    Validating instance against JSON Schema.
Summary
- Errors: no (0)
- Warnings: no (0)
- Info/Hints: 0

No issues found.
```
The request to the **authenticated** repository contains `Authorization` header and the request to the **unauthenticated** repository does not contain the header.

shared/src/document-loader/direct-url-document-loader.ts
```diff
@@ -166,18 +244,44 @@ export class DirectUrlDocumentLoader implements DocumentLoader {
                 });
             }
             const baseURL = `${parsedUrl.protocol}//${normalizedHost}${parsedUrl.port ? `:${parsedUrl.port}` : ''}`;
-            const response = await this.ax.get(requestPath, {
+            let authHeaders: Record<string, string> | undefined;
+            const authHeaderNames: string[] = [];
+            if (this.directUrlAuthPlugin) {
+                try {
+                    authHeaders = await this.directUrlAuthPlugin.getAuthHeaders(`${baseURL}${requestPath}`, undefined);
```
markscott-ms, line 251: In loadMissingDocument, once a URL passes all the safety gates (private-IP check, host allowlist, path pattern, no query string), the loader calls `directUrlAuthPlugin.getAuthHeaders`.

The plugin is called for **every** direct-URL fetch to **any** allowlisted host, and whatever headers it returns are attached verbatim to the outgoing request. There is no host-based gating in the loader — the only thing the plugin receives to make a decision is the URL string (getAuthHeaders(url, requestBody), per direct-url-auth-plugin.ts:2). All this is fine, but...

**The allowlist normally holds more than one host**:

- calm.finos.org is there by default (DEFAULT_ALLOWED_REMOTE_HOSTS).
- The user adds their internal schema host with `calm init-config --allowed-remote-hosts …`.
- They may have added other hosts (`raw.githubusercontent.com`) for unrelated references.

If the plugin author writes the _obvious_ minimal implementation — "return my bearer token" — without inspecting the URL host, then any reference the loader resolves over the network gets that Authorization header. A CALM document that `$refs` `https://calm.finos.org/…` for a schema that isn't bundled locally would send the user's **internal** org token to FINOS's server over the wire.

One solution could be to add another method to the plugin interface `direct-url-auth-plugin.ts` to get the list of domain(s) it supports, and only call it (and include headers) for those domains.

Another alternative is for the user to configure the auth plugin on a per-domain basis, like gitconfig's credential blocks: https://git-scm.com/docs/gitcredentials#_credential_contexts

markscott-ms (changes requested):

shared/src/document-loader/direct-url-document-loader.ts
jimthompson5802: @markscott-ms   this is how I understand the issue...if the user **does not** implement the check on the URL and returns the Bearer token for every call, the possibility exists the user's token will be sent to an unintended recipient, potentially leaking the token to an outside party.

Of the two suggestions:
* "add another method to the plugin interface direct-url-auth-plugin.ts to get the list of domain(s) it supports, and only call it (and include headers) for those domains." or
* "for the user to configure the auth plugin on a per-domain basis, like gitconfig's credential blocks: https://git-scm.com/docs/gitcredentials#_credential_contexts"

Both require adding some code the `cli or shared` module, however the first requires the user to write additional code.  So I'm leaning to the second approach, which reduces the code the end user has to write.

I'm thinking the `.calm.json` with `directUrlAuth plugin support will look something like this....
```json
{
  "allowedRemoteHosts":[
    "my-calm.repo",
    "your-calm.repo"
  ],
  "directUrlAuth": {
    "module": "~/Desktop/finos/calm-web-repo/custom-idp/v3/dist/direct-url-auth.js",
    "configPath": "~/Desktop/finos/calm-web-repo/custom-idp/v3/generated/direct-url-auth.json",
    "supportedRepos": [
         "my-calm.repo"
    ]
  }
}
```

One obvious usability issues of this approach is that `my-calm.repo` has to be repeated in two places.  To address this usability issue, one approach will be for the `allow-list` to be the union of `allowedRemoteHosts` and `supportedRepos`.

What do you think?

jimthompson5802: @markscott-ms   decided on the second approach.  here is the summary...

`authenticatedHosts` lets CLI users declare which remote document hosts require direct-URL authentication.

```json
{
  "allowedRemoteHosts": [
    "schemas.example.com",
    "public.example.com"
  ],
  "directUrlAuth": {
    "module": "~/plugins/direct-url-auth.js",
    "configPath": "~/plugins/direct-url-auth.config.json",
    "authenticatedHosts": [
      "protected.example.com",
      "private-schemas.example.com"
    ]
  }
}
```

- `allowedRemoteHosts` covers unauthenticated remote hosts.
- `authenticatedHosts` identifies protected hosts, automatically allows them for direct document loading, and invokes the authentication module for requests to those hosts.
- Headers returned by the module are attached to protected-host requests.
- Host entries must be exact, case-insensitive hostnames; URLs, ports, paths, and wildcards are rejected.
- Other allowed hosts continue loading without the authentication flow.
- Authentication failures, including HTTP 401 and 403 responses, report clear authentication-specific errors.

markscott-ms: Second solution is my preferred one too. The only question would be whether we want a configuration like the one you have above, or one where `directUrlAuth` can be an array of modules, each of which can have a different module and different supported domains:

```
{
  "allowedRemoteHosts": [
    "schemas.example.com",
    "public.example.com"
  ],
  "authenticators": [
    {
       "module": "~/plugins/direct-url-auth.js",
       "configPath": "~/plugins/direct-url-auth.config.json",
       "authenticatedHosts": [
         "protected.example.com",
         "private-schemas.example.com"
       ]
    },
    {
       //... second authenticator module for protected.another-domain.com
    }
  ]
}
```

This could potentially eliminate the need for the existing, differently configured, CalmHub authenticator plugin as it can just be listed in `authenticators` with the CalmHub domain in its `authenticatedHosts`.

(The need for multiple authenticators can easily within one organization if there is a mixture of places CALM artifacts are located, and they are using different authentication mechanisms, e.g. SPNEGO, OIDC, ...)

jimthompson5802: @markscott-ms   interesting suggestion.  I have to admit the current design intent was to leave CALM Hub authentication untouched.  

Given the MultiStrategyDocumentLoader approach, as best I can tell, there are only two document loaders that involve retrieving from a remote site:  CALM Hub and Direct URL.  
| Loader | Assumed document location |
|---|---|
| `WorkspaceDocumentLoader` | A file listed in `<bundle>/workspace-manifest.json`; relative manifest paths are resolved under the workspace bundle, normally `.calm-workspace/bundles/<name>/files/…`. |
| `MappedDocumentLoader` | An explicit URL → local-file mapping. Relative mapped paths are resolved under `basePath`. |
| `CalmHubDocumentLoader` | The configured CALM Hub origin (`calmHubUrl`). It requests the document URL’s pathname from that origin; HTTP(S) references must use exactly that origin. |
| `FileSystemDocumentLoader` | A relative reference under `basePath`, or an existing absolute/local filesystem path. It also preloads schemas recursively from the CALM meta-schema directory and optional `schemaDirectoryPath`. |
| `DirectUrlDocumentLoader` | The exact absolute HTTP(S) URL supplied in the reference, provided its host is allowlisted (default: `calm.finos.org`). |

Given this, instead of an array for the authenticators, maybe it is just two additional objects like this.
```
{
  "allowedRemoteHosts": [
    "schemas.example.com",
    "public.example.com"
  ],
  "directUrlsAuthenticator": {
       "module": "~/plugins/direct-url-auth.js",
       "configPath": "~/plugins/direct-url-auth.config.json",
       "authenticatedHosts": [
         "protected.example.com",
         "private-schemas.example.com"
       ]
    },
"calmHubAuthenticator ":   {
         "module": " /plugins/calm-hub-auth.js",
       "configPath": "~/plugins/direct-url-auth.config.json",
       "authenticatedHosts": [
         "protected.calmhub.com"
       ]
    }
}
```

I'm assuming for `calmHubAuthenticator`, the default values for `module`, `configPath` and `authenticatedHosts` would be whatever is currrently coded in the CALM.  I can see supporitng `calmHubAuthentictor` gives an organization the flexibility to use whatever mechanism that depends on a single HTTP `Authorization` header only.

Since you have an easier way of reaching out to @rocketstack-matt, @jpgough-ms and @willosborne,  let's see what their initial reaction to `calmHubAuthenticator`. 

In the meantime, I'll take a closer look at the CALM Hub authentication for an impact assessment.

We should queue this up for the next office hours.

jimthompson5802:
@markscott-ms   I seem to recall you suggesting that the directUrlAuth configuration setting should follow the exisitng cli use of environment variables, i.e, if environment vairable is set use its value if not, then look in config file.  

Direct URL authentication configuration is now flattened in `~/.calm.json`:

```json
{
  "directUrlAuthModule": "~/plugins/direct-url-auth.js",
  "directUrlAuthConfigPath": "~/config/direct-url-auth.json",
  "directUrlAuthAuthenticatedHosts": ["protected.example.com"]
}
```

Each setting can be overridden at runtime by an environment variable:

- `CALM_DIRECT_URL_AUTH_MODULE`
- `CALM_DIRECT_URL_AUTH_CONFIG_PATH`
- `CALM_DIRECT_URL_AUTH_AUTHENTICATED_HOSTS` — comma-separated hostnames

Environment values take precedence field-by-field over `~/.calm.json`; unset or empty variables leave the file value in place. The host-list environment value is trimmed and split into an array.

The CLI assembles these resolved top-level values into its existing internal direct-auth plugin configuration, so plugin loading, authentication, and host enforcement remain unchanged. 

Let me know if this is what you were looking for.  I still have to do a bit more robust testing of this last change.

jimthompson5802:
Table of supported CLI configuration parameters and its equivalent enviornment variable:

| Configuration parameter | Equivalent environment variable | Format |
| --- | --- | --- |
| `calmHubUrl` | `CALM_HUB_URL` | CalmHub base URL, e.g. `https://calm.example.com` |
| `allowedRemoteHosts` | `CALM_ALLOWED_REMOTE_HOSTS` | JSON: array of hostnames. Environment: comma-separated hostnames, e.g. `schemas.example.com,models.example.com` |
| `authPluginPath` | `CALM_AUTH_PLUGIN_PATH` | Path to a JavaScript authentication plugin file (`.js`) |
| `directUrlAuthModule` | `CALM_DIRECT_URL_AUTH_MODULE` | Path to a JavaScript direct-URL authentication module (`.js`) |
| `directUrlAuthConfigPath` | `CALM_DIRECT_URL_AUTH_CONFIG_PATH` | Optional path to the module’s configuration file |
| `directUrlAuthAuthenticatedHosts` | `CALM_DIRECT_URL_AUTH_AUTHENTICATED_HOSTS` | JSON: non-empty array of hostnames. Environment: comma-separated hostnames; no URLs, ports, paths, or wildcards |

Here is a complete illustrative example configuration file.

```json
{
  "calmHubUrl": "https://calm.example.com",
  "allowedRemoteHosts": [
    "schemas.example.com",
    "models.example.com"
  ],
  "authPluginPath": "/absolute/path/to/auth-plugin.js",
  "directUrlAuthModule": "/absolute/path/to/direct-url-auth.js",
  "directUrlAuthConfigPath": "/absolute/path/to/direct-url-auth-config.json",
  "directUrlAuthAuthenticatedHosts": [
    "my.internal.schemas.example.com",
    "my.internal.models.example.com"
  ]
}
```

shared/src/document-loader/direct-url-document-loader.ts
```diff
@@ -166,22 +297,50 @@ export class DirectUrlDocumentLoader implements DocumentLoader {
...
+                !this.directUrlAuthAuthenticatedHosts || this.directUrlAuthAuthenticatedHosts.has(normalizedHost)
+            )) {
+                try {
+                    authHeaders = await this.directUrlAuthPlugin.getAuthHeaders(`${baseURL}${requestPath}`, undefined);
+                    authHeaderNames.push(...Object.keys(authHeaders));
+                } catch (error) {
+                    throw new DocumentLoadError({
+                        name: 'AUTHENTICATION_FAILED',
+                        message: `Direct URL authentication failed for ${documentId}. Check direct URL auth configuration and remote credentials.`,
+                        cause: documentLoadCause(error, parsedUrl.hostname),
```
markscott-ms, line 312: Auth plugin errors are attached verbatim as `DocumentLoadError.cause` and get printed by the verbose cause-chain logger, potentially leaking credentials.

This PR's adds redaction logic for some errors - see `sanitizeTlsCertificateError`.

Given that it is difficult to protect *and* offer debug capability for something we aren't writing, should this be a plugin writer's guide that they should ensure that any secrets are not exposed in thrown exceptions?

markscott-ms (changes requested):

shared/src/document-loader/direct-url-document-loader.ts
jimthompson5802: @markscott-ms  I just to confirm this comment is about updating documentation such as as the `cli` README section on the `directUrlAuth` plugin and not recommending any code update to the `cli` or `shared` modifications.

Is this a correct interpretation?

markscott-ms: Yes, I think so.

A short note in our authentication plugin guide(s) saying "When implementing your plugin, take care with handling of any secrets, avoiding their capture in exceptions and log messages."

markscott-ms (approved):
As discussed in OH, this is approved as is. Documentation (as part of the CALM Docs website) can be handled in a separate PR and take care of the secret logging warning for implementors

jimthompson5802:
@markscott-ms   Thank you for the help.  PR is now merged.  I'll start working on the follow on PR for the documentation and sample code.

PR #3046, opened 2026-09-01
fix(cli): include flows in architecture diffs

### LINKED ISSUE #2998: calm diff does not compare flows: deleting a flow reports no changes and --exit-code returns 0
`calm diff` compares nodes, relationships, control items and ADRs, but not `flows`. If the only difference between two architectures is in their flows (transitions removed, or a flow deleted completely) the diff comes back empty, and `--exit-code` returns 0. The help text suggests `--exit-code` for gating version bumps in CI, so a CI gate built on it will pass an architecture whose business processes have changed or disappeared.

Reproduced on 1.56.0, 1.55.0 and 1.53.0.

### Steps to reproduce

Save this as `a.json`. It validates cleanly on 1.56.0 (`hasErrors: false, hasWarnings: false`):

```json
{
  "$schema": "https://calm.finos.org/release/1.0/meta/calm.json",
  "nodes": [
    { "unique-id": "api", "node-type": "service", "name": "API", "description": "Receives requests." },
    { "unique-id": "db", "node-type": "database", "name": "DB", "description": "Stores records." },
    { "unique-id": "audit", "node-type": "service", "name": "Audit", "description": "Writes the audit trail." }
  ],
  "relationships": [
    { "unique-id": "rel-api-db", "description": "API writes to DB.",
      "relationship-type": { "connects": { "source": { "node": "api" }, "destination": { "node": "db" } } } },
    { "unique-id": "rel-api-audit", "description": "API records to audit.",
      "relationship-type": { "connects": { "source": { "node": "api" }, "destination": { "node": "audit" } } } }
  ],
  "flows": [
    { "unique-id": "flow-write", "name": "Write", "description": "A write with an audit record.",
      "transitions": [
        { "relationship-unique-id": "rel-api-db", "sequence-number": 1, "description": "Persist." },
        { "relationship-unique-id": "rel-api-audit", "sequence-number": 2, "description": "Audit." }
      ] }
  ]
}
```

Drop a transition from the flow, so the audit step no longer happens:

```bash
jq '.flows[0].transitions |= .[0:1]' a.json > b.json
calm diff -a a.json -b b.json -f summary
calm diff -a a.json -b b.json --exit-code; echo "exit: $?"
```

Output on 1.56.0:

```
CALM architecture diff
----------------------
Nodes:         +0  -0  ~0  ↔0  =3
Relationships: +0  -0  ~0  ↔0  =2
exit: 0
```

Same result if you delete the flow entirely:

```bash
jq '.flows = []' a.json > c.json
calm diff -a a.json -b c.json --exit-code; echo "exit: $?"   # exit: 0
```

The JSON output tells the same story: with `-f json` the result object has 16 keys (nodesAdded/Removed/Modified/Same/Renamed, the edges equivalents, invalidItems, adrDiffItems, controlItems*) and the string "flow" appears nowhere in it.

### Expected

Flows compared like nodes and relationships are (flowsAdded/Removed/Modified, ideally with transition-level detail on modified), and flow changes counted by `--exit-code`. If that's out of scope for now, a note in the docs that `diff` skips flows would at least let CI consumers know to compare them separately.

### Context

We attach controls to flows (machine-readable test strategies) and gate releases on them. A flow-only edit is exactly the change that should invalidate an accepted control configuration, and right now it's invisible to `diff` — we compute a per-flow digest ourselves to work around it. Repointing a flow control's `config-url` is silent in the same way.

I see #2586 (control diffs) and #2526 (ADR diffs) are extending what `diff` covers, so maybe flows are already on someone's radar; couldn't find an issue for it.

Happy to put a PR together if the approach sounds reasonable — an initial version would mirror the existing element comparison over `flows[]`, matching flows by `unique-id` and transitions by `relationship-unique-id` plus `sequence-number`, feeding the same summary/JSON/exit-code paths.

ASVLCII:
## Description

Fixes #2998. `calm diff` now includes top-level flows. Flows match by `unique-id`; a changed flow is reported as modified.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds new functionality)
- [ ] 📚 Documentation update

## Affected Components

- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [ ] All existing tests pass

Focused model and shared diff tests pass. The focused commands exit non-zero only because the repository applies global coverage thresholds to a single-file run.

## Checklist


ASVLCII:
All checks and EasyCLA are green now. Feel free to merge when you're ready.

shared/src/commands/diff/diff-core.ts
```diff
@@ -15,13 +16,15 @@ export type DiffOutputFormat = 'json' | 'summary';
 
 export type DiffDocumentType = 'architecture' | 'pattern';
 
+type DiffResult = NodesAndRelationshipsDiffResult & Partial<FlowDiffResult>;
```
aamanrebello, line 19: Why do we need it to be partial here? Just asking for clarification

calm-models/src/diff/diff.spec.ts
```diff
@@ -555,6 +555,35 @@ describe('diff', () => {
...
+    describe('diffArchitectures - flow changes', () => {
+        const flow = {
+            'unique-id': 'write-flow',
+            name: 'Write flow',
+            description: 'Writes a record.',
+            transitions: [{ 'relationship-unique-id': 'gateway-to-payment', 'sequence-number': 1, description: 'Write.' }],
+        };
+
+        it('detects added, removed, modified, and unchanged flows', () => {
+            const base = { ...testArchitectures.baseArchitecture, flows: [flow] } as CalmArchitectureSchema;
```
aamanrebello, line 567: Could we perhaps create new architectures within `./fixtures/diff-test-architectures.json`? This is the current convention in these specs, rather than relying on the base architecture. Maybe we could update the base architecture to add `flows: [flow]` and then create a new architecture based on `changed`.

shared/src/commands/diff/diff-core.ts
ASVLCII: It's partial because this path handles both architecture and pattern diffs. Architecture diffs always include the flow arrays, but pattern diffs only return nodes and relationships, so requiring FlowDiffResult here would make the pattern path pretend those fields exist. I can switch it to an explicit union if you'd prefer; the optional fields were just the smallest way to keep both paths accurate.

calm-models/src/diff/diff.spec.ts
ASVLCII: Yeah, that fits the existing specs better. I'll move these cases into diff-test-architectures.json, add the flow to the base fixture, and derive the changed case from it. Keeps the test setup consistent with the rest of the file.

aamanrebello:
I am happy with these changes, I would also want the opinion of Matt Bain on this though.

ASVLCII:
Sounds good, happy to wait for Matt's review. The fixture move I mentioned is still outstanding on my side.

calm-models/src/diff/diff-types.ts
```diff
@@ -105,7 +111,14 @@ export interface ControlDiffResult {
     controlItemsModified: { [controlId: string]: ControlItemDiffResult };
 }
 
+export interface FlowDiffResult {
```
rocketstack-matt, line 114: Nit: this is the only composed result interface here without a doc comment (`AdrDiffResult`/`ControlDiffResult` above both have one).

rocketstack-matt (commented):
Design and tests look right — flows are matched by `unique-id` and correctly wired into `hasChanges`/summary/JSON output. Two things before merge: this branch conflicts with `main` (a parallel PR added `MetadataDiffResult` to the same `ArchitectureDiffResult` union), and the CLI README's diff section doesn't mention flows yet (it still documents/shows only Nodes/Relationships in the `--format summary` output).

shared/src/commands/diff/diff-core.ts
rocketstack-matt: Confirmed — `diffPatterns` never returns flow fields at all, so `Partial<FlowDiffResult>` is correct here. No change needed.

calm-models/src/diff/diff.spec.ts
rocketstack-matt: Agree with Aman's convention point — worth moving before merge so the flow test cases match the rest of the file.

PR #3059, opened 2026-09-07
fix(shared): resolve prefixItems alternatives in pattern validation

### LINKED ISSUE #3058: A pattern decision cannot select one candidate when two candidates share a unique-id
## Bug Report

A slot is one entry in a pattern's `prefixItems` array. A slot can hold a `oneOf` or an `anyOf` array. Each entry in that array is a candidate. A pattern decision selects a candidate by its `unique-id`.

Two candidates in one slot can declare the same `unique-id`. That id then names two candidates. A decision cannot select only one of them. The pattern declares an option that the pattern cannot express.

The rule `unique-ids-must-be-unique-in-pattern` does not report this. The rule reads only the bare `prefixItems` entries. The rule does not read the candidates in a slot. Therefore `calm validate` accepts the pattern.

### Steps to Reproduce:
1. Write a pattern with two candidates in one slot. Give both candidates the `unique-id` `cache`.

```json
{
  "properties": {
    "nodes": {
      "type": "array",
      "prefixItems": [
        {
          "anyOf": [
            { "properties": { "unique-id": { "const": "cache" }, "name": { "const": "Redis" } } },
            { "properties": { "unique-id": { "const": "cache" }, "name": { "const": "Memcached" } } }
          ]
        }
      ]
    },
    "relationships": { "type": "array", "prefixItems": [] }
  }
}
```

2. Run `calm validate` against the pattern.

### Expected Result:
`calm validate` reports a duplicate `unique-id`. The rule reports that error today for two bare `prefixItems` nodes with one id.

### Actual Result:
`calm validate` reports no error. The tool accepts the pattern.

### Consequences
Each consumer finds a candidate by its `unique-id`. Each consumer therefore gets the same ambiguous result.

`calm generate` selects both candidates for the id `cache`. The command writes two nodes with one id:

```json
[{"unique-id":"cache","name":"Redis"},{"unique-id":"cache","name":"Memcached"}]
```

`calm validate` then rejects that architecture. It reports `unique-ids-must-be-unique-in-architecture`. This error names the architecture and a different rule. The error does not name the pattern slot. The author must find the cause without help.

The CALM Hub visualiser draws two nodes with the id `cache`. ReactFlow does not define the result for two nodes with one id. The pattern differ and the Spectral rules also match on `unique-id`.

### `oneOf` has the same defect
Change `anyOf` to `oneOf` in the pattern above. The output stays the same. `oneOf` permits exactly one candidate in the slot. The defect is not the number of nodes. The decision has no id that selects one candidate and not the other.

### Scope
| Declaration site | The rule reads it today |
|---|---|
| `nodes`, `relationships` and interfaces as bare `prefixItems` entries | yes |
| the same, as candidates in a slot's `oneOf` or `anyOf` | no |

The rule misses an interface id on a candidate node for the same reason.

### Environment:
`shared` (`@finos/calm-shared`), `src/spectral/functions/pattern/ids-are-unique.ts`. I reproduced the defect on `main` at `0034591f`.

### Additional Context:
I found this defect during the design work for #2859. This defect is not part of that work. A pattern that uses no `items` catalog can show it. Therefore this issue and its fix stay separate. The two behaviour changes then keep separate release notes.

The fix changes behaviour. A pattern from outside this repository can declare duplicate ids in a slot. `calm validate` accepts that pattern today. `calm validate` rejects it after the fix. No pattern in this repository changes its result. I checked all 25 pattern files against the new queries.

### LINKED ISSUE #3062: A prefixItems slot that declares both oneOf and anyOf is unsatisfiable and passes validation
## Bug Report

A choice block is a `oneOf` or an `anyOf` array. A `prefixItems` slot can hold one. An `items` catalog is one.

A block can declare both keywords. JSON Schema then requires an element to satisfy both. CALM alternatives pin distinct `unique-id` values, so the two sets are normally disjoint and no element can satisfy them. The pattern is unsatisfiable. Where the sets overlap, the wider keyword declares alternatives that nothing can select.

Nothing reports this. `calm validate` accepts the pattern, and `calm generate` returns an empty selection without a message.

#3060 adds a rule for an `items` catalog. This issue covers the same fault in a `prefixItems` slot, which #3060 leaves alone because a slot is an existing construct and the rule is a behaviour change.

### Steps to Reproduce:
1. Write a pattern with a slot that declares both keywords:

```json
"nodes": {
  "type": "array",
  "prefixItems": [
    {
      "oneOf": [ { "properties": { "unique-id": { "const": "redis" } } } ],
      "anyOf": [ { "properties": { "unique-id": { "const": "memcached" } } } ]
    }
  ]
}
```

2. Run `calm validate`.
3. Run `calm generate` and choose `memcached`.

### Expected Result:
`calm validate` reports an error. The block must declare one keyword.

### Actual Result:
Step 2 reports nothing. Step 3 returns an empty selection:

```
askedFor: "memcached"   selected: []
```

### The two shapes
JSON Schema combines the keywords with AND. An element must satisfy `oneOf` **and** `anyOf`. It is not a union of alternatives. Each CALM alternative pins a distinct `unique-id`, so an element matches at most one alternative in each keyword. That gives two shapes, both confirmed against Ajv.

**Disjoint.** The keywords name different alternatives.

| `oneOf` | `anyOf` | Result |
|---|---|---|
| `redis` | `memcached` | No element validates. `redis` fails the `anyOf`, `memcached` fails the `oneOf`. The pattern is unsatisfiable. |

**Overlapping.** The keywords share alternatives.

| `oneOf` | `anyOf` | Result |
|---|---|---|
| `A, B, C` | `A, B, C, D, E` | Only `A`, `B` and `C` validate. The narrower keyword binds. `D` and `E` are declared and can never be selected. |

### Severity
Both shapes are errors.

The disjoint shape is plainly an error, because no architecture can ever match the pattern.

The overlapping shape does generate a valid architecture, so `warn` is arguable. It is still an error, because the author has declared alternatives that no answer can reach, and the pattern does not say what it appears to say. A rule that reports the shape needs no satisfiability test, which keeps it simple and keeps the message the same in both cases.

### Note on the tooling
`flattenOneOfAndAnyOf` in `options.ts` resolves `item.oneOf ?? item.anyOf`, and `patternTransformer.ts` resolves `hasOneOf ? item['oneOf'] : item['anyOf']`. Making them read both keywords would not fix this. The two keywords are combined with AND, not OR, so their union is not a list of alternatives. Rejecting the shape is the correct response.

### Environment:
`shared` (`@finos/calm-shared`). Reproduced on `main` at `0034591f`. No pattern in this repository declares both keywords, so the rule regresses nothing here.

### LINKED ISSUE #3068: Interfaces on a prefixItems alternative are never checked, and interfaces are pooled across alternatives
## Bug Report

A `prefixItems` entry can hold a `oneOf` or an `anyOf` array. Each entry in that array is an alternative, and each alternative is a node.

`interface-id-exists-on-node` builds its list of nodes from `$.properties.nodes.prefixItems[*]`. That query returns the entry, not the alternatives inside it. The rule then does two things wrong.

It identifies the entry by the first alternative's `unique-id`. A reference to any other alternative matches nothing, so the rule returns early and checks nothing.

For the first alternative it does match, it gathers interfaces from every alternative in the entry. The interfaces are pooled, so a reference to one alternative can name an interface that belongs to another and pass.

### Steps to Reproduce:
1. Write a pattern with one `prefixItems` entry holding two alternatives. Give `cache` the interface `port-a` and `queue` the interface `port-b`.
2. Add a connects relationship from `webapp` to `cache`, asking for interface `port-b`.
3. Run `calm validate`.

### Expected Result:
An error. `port-b` belongs to `queue`, not to `cache`.

### Actual Result:
No error. The same silence occurs for every combination:

| Relationship asks for | Result today |
|---|---|
| `cache` interface `port-a` (correct) | no message |
| `cache` interface `port-b` (wrong) | no message |
| `queue` interface `port-b` (correct) | no message |
| `queue` interface `port-a` (wrong) | no message |

No alternative in an entry is checked at all.

### Fix
Build the node list from individual alternatives rather than from the entry. Each alternative then carries its own id and its own interfaces, so the pooling and the first-alternative match both disappear.

### Environment:
`shared` (`@finos/calm-shared`), `src/spectral/functions/pattern/interface-id-exists-on-node.ts`. Reproduced on `main` at `0034591f`.

### Additional Context:
Found while scoping #2859. It is a defect in the `prefixItems` model and needs no `items` catalog to reproduce, so it is filed and fixed separately from that work.

This fix can reject a pattern that `calm validate` accepts today. No pattern in this repository is affected.

### LINKED ISSUE #3069: Unreferenced nodes declared as prefixItems alternatives are never reported
## Bug Report

`pattern-nodes-must-be-referenced` warns when a pattern declares a node that no relationship and no decision names. Such a node can never appear in an architecture, so it is either unfinished work or a leftover.

Its selector reads bare `prefixItems` entries only:

```
$.properties.nodes.prefixItems[*].properties.unique-id.const
```

A node declared inside `prefixItems[i].oneOf` or `prefixItems[i].anyOf` has its `unique-id` one level deeper, so the selector never reaches it. The rule cannot warn about an alternative, whatever its state.

### Steps to Reproduce:
1. Write a pattern with `webapp` and `database` connected, plus one `prefixItems` entry holding the alternatives `redis` and `memcached`.
2. Do not reference either alternative from any relationship or decision.
3. Run `calm validate`.

### Expected Result:
A warning for `redis` and one for `memcached`, as the rule already gives for an unreferenced bare node.

### Actual Result:
Nothing is reported.

| Declaration | Warns today |
|---|---|
| a bare `prefixItems` entry | yes |
| an alternative in `prefixItems[i].oneOf` or `.anyOf` | no |

### Note
A decision references its alternatives through `choices[].nodes`, which sits under `relationship-type`, and the rule's function reads every string under `relationship-type`. So an alternative that a decision offers is already counted as referenced. Widening the selector therefore warns only about an alternative that nothing offers and nothing connects.

### Environment:
`shared` (`@finos/calm-shared`), `src/spectral/rules-pattern.ts`. Reproduced on `main` at `0034591f`.

### Additional Context:
Found while scoping #2859. It is a defect in the `prefixItems` model and needs no `items` catalog to reproduce, so it is filed and fixed separately from that work.

The fix is `warn` severity and can add a warning to a pattern that is clean today. No pattern in this repository is affected.

YoofiTT96:
## Description
A `prefixItems` entry is either a node or a `oneOf`/`anyOf` holding alternatives. Every rule that walks `$.properties.nodes.prefixItems[*]` has to cope with both shapes, and each coped differently. Four coped wrongly or not at all.

Part of #2932 rework

`declaration-paths.ts` now owns where a pattern declares a node. It builds the queries that find declarations and reads the resulting pointers back, both driven by one keyword list, so a rule cannot disagree with another about where a declaration lives. The shape-handling disappears from the rules instead of being repeated four ways, and `interface-id-exists-on-node` is a net deletion.

None of this concerns #2859. These are pre-existing gaps in the `prefixItems` model, reachable by a pattern that uses no new construct. They are grouped because they are one theme with one release note, and because they carry every behaviour change in the wider items-support work — leaving that work free to be purely additive.

Closes #3058
Closes #3062
Closes #3068
Closes #3069

### Behaviour changes — breaking
`main` never looked inside `oneOf` or `anyOf`, so every shape below was accepted. Each is now checked. No published API changes.

| Shape | Now | Before |
|---|---|---|
| Two alternatives of one entry share a node `unique-id`<br>`oneOf: [ node "db", node "db" ]` | error | `calm generate` emitted **both** nodes. Verified: with a decision selecting `db`, architecture validation reports the duplicate |
| A relationship names an interface belonging to a sibling alternative<br>`webapp -> postgres, interfaces: ["my-port"]` where `my-port` is on `mysql` | error | interfaces were pooled across an entry, so the chosen node need never have declared it. Architecture validation caught it after generation |
| An entry declares both keywords<br>`{ oneOf: [...], anyOf: [...] }` | error | an element must satisfy both, so the entry is unsatisfiable when the id sets are disjoint and declares unselectable alternatives when they overlap |
| Nothing references an alternative<br>`oneOf: [ node "orphan" ]` | warning | invisible to the rule. Fails a build only under `--strict` |

One further rejection is a deliberate contract, not a correction. A `unique-id` names one kind of thing, so a node id may not equal an interface id even across alternatives that never appear together. Both architectures such a pattern produces are valid, so this rejects something that was not broken. It is recorded in `PATTERN-DECISIONS.md` rather than justified by #3058.

No pattern in this repository changes result. Every tracked `.json` declaring `properties.nodes.prefixItems` or `properties.relationships.prefixItems` — 28 files, 13 of which use alternatives — was run against the real ruleset on `main` and on this branch. The findings are byte-identical.

### Worth knowing
Two alternatives may expose the same interface id, because only one of them is ever built. A decision never names an interface on its own, so nothing has to tell them apart:

```
oneOf: [ postgres exposing "db-port", mysql exposing "db-port" ]      accepted
```

Do not give an entry its own `properties` as well as alternatives. `calm generate` keeps the selected alternative and discards the entry's own `properties`, so whatever it declares is silently lost. Validation reports a duplicate id when the two halves share one, which catches the common case without naming the fault. A rule that names it directly is worth a follow-up.

The two-keyword check is a filtered `given` with the built-in `falsy`, not a custom function. Spectral's `xor` cannot express it, because an entry may legitimately declare neither keyword.

When `unique-ids-must-be-unique-in-pattern` finds an id twice, it points at the second one. That is only useful if the second one is the later in the file.

It is not, because the rule finds ids by running one query per declaration site and joining the results. Every fixed entry arrives before any alternative, whatever order they appear in. So a pattern declaring `dup` as an alternative at index 0 and again as a fixed entry at index 1 was reported at index 0 — the earlier of the two.

The matches are now sorted by position before the rule looks for duplicates.

`PATTERN-DECISIONS.md` is a new behaviour contract at the repository root. It records what each tool guarantees, with no implementation detail and no test references, so it does not go stale when a function is renamed.

### Review round
Four of the five review findings led to a change, and the fifth changed in a different way than its diagnosis suggested.

Ordering no longer compares pointer text at all. Text put an alternative ahead of the entry that holds it, because `oneOf` precedes `properties`, so a node id declared in both was blamed on the entry — the half an architecture always builds. `byBuildOrder` compares the parsed indices of the containing declaration instead, which also removes the digit padding that capped array indices at 999,999. There was no test pinning which side gets blamed for that shape; there is now.

`JSONPathMatch` is exported from `helper-functions.ts` rather than redeclared, and the hand-rolled `groupBy` is lodash's.

On the performance finding: these paths are rooted at `properties.nodes`, so they are lookups, not full-document traversals — adding 5,000 relationships to a document leaves the timing unchanged. Against `main` the trade runs the other way from how the comment reads, since `main` runs one query over the pattern and three per node inspected where this runs three over the pattern and one per node. But the eager part was real: all three queries ran before the search started, so a node declared as a plain `prefixItems` entry paid for two it never needed. The lookup now returns at the first path that matches, and `declaredId` reads its path with lodash instead of a JSONPath query. Over 600 endpoints on a 300-node pattern that is 23ms against `main`'s 112ms.

## Type of Change
- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update

A pattern that validates clean today can fail after this change, so a `BREAKING CHANGE:` footer is on the commit. Without it `fix(shared)` would ship these rejections as a patch release.

## Affected Components
- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

`cli/` is the JUnit fixture only, which pins the live rule list. No CLI source changes.

## Commit Message Format ✅
`fix(shared): resolve prefixItems alternatives in pattern validation`, carrying a `BREAKING CHANGE:` footer that lists the rejections above.

The review round adds `fix(shared): order declarations by index, not by pointer text`, `refactor(shared): reuse the shared match type and lodash groupBy`, and `perf(shared): stop searching declaration sites at the first match`.

## Testing
- [x] All existing tests pass

Four spec files are new: `rules-pattern.spec.ts`, `node-id-exists.spec.ts`, `declaration-paths.spec.ts` and `is-defined-in-oneof-or-anyof.spec.ts`. No ruleset-level test existed, because `validate.spec.ts` mocks Spectral out, so selectors and registration were untested. The last two rules had no unit test at all, on `main` or here.

Every new guarantee was mutation-tested rather than trusted. Reintroducing the interface pooling, dropping the document-order sort, treating an entry's `properties` as an alternative, and unregistering the keyword rule each turn the matching tests red and leave the rest green.

Coverage on the changed files is 100% statements and branches, except `interface-id-exists-on-node` at 95.45/92.85 on a pre-existing branch. Across `src/spectral` it is 94.64/93.04. Full root `npm test` is green.

After the review round: `shared` passes 105 files and 1208 tests, `cli` 29 files and 659 tests, and root `npm test` is green. Restoring the raw-pointer sort turns the new ordering test red along with the existing numeric one, and nothing else.

## Checklist


shared/src/spectral/functions/pattern/ids-are-unique.ts
```diff
@@ -1,25 +1,70 @@
 import { JSONPath } from 'jsonpath-plus';
+import { partition } from 'lodash';
 import { IFunctionResult, RulesetFunctionContext } from '@stoplight/spectral-core';
 import { detectDuplicates } from '../helper-functions';
+import { containingDeclaration, containingEntry, declaredIdPaths, declaredInterfaceIdPaths, isAlternative } from './declaration-paths';
+
+interface Match {
+    value: unknown;
+    pointer: string;
+}
```
markscott-ms: duplicates JSONPathMatch already defined in ../helper-functions.ts.

shared/src/spectral/functions/pattern/ids-are-unique.ts
```diff
@@ -1,25 +1,70 @@
...
+
+/**
+ * The rule blames the second declaration it sees, but three queries per id kind arrive
+ * grouped by query rather than by position. Padding keeps prefixItems/2 before /10.
+ */
+function inDocumentOrder(matches: Match[]): Match[] {
```
markscott-ms: Sorts JSONPath match pointers as strings after zero-padding digit runs, so detectDuplicates can walk them in "document order" and blame whichever one it sees second as the duplicate.

The bug: for a single `prefixItems` entry that declares `unique-id` both directly and inside its own `oneOf`/`anyOf`, the two pointers are:

- fixed: /properties/nodes/prefixItems/000000/properties/unique-id/const
- alternative: /properties/nodes/prefixItems/000000/oneOf/000000/properties/unique-id/const

After the shared prefix, one continues `properties/...` and the other `oneOf/....` Lexicographically 'o' < 'p', so the `oneOf` pointer sorts before the fixed pointer — the opposite of what the file's own comment claims ("blames the second declaration it sees"). `detectDuplicates` then sees the alternative first, records its id as "seen", and blames the fixed declaration when it hits the same id — backwards from the intended behavior of pointing at the newer/alternative declaration.

This only misfires for same-entry fixed-vs-own-alternative duplicates (the exact anti-pattern the new `PATTERN-DECISIONS.md` calls out as invalid) — cross-entry ordering is unaffected since `prefixItems/000000` vs `prefixItems/000001` sorts correctly regardless. There's no test pinning which side gets blamed in this case.

shared/src/spectral/functions/pattern/ids-are-unique.ts
```diff
@@ -1,25 +1,70 @@
...
+    const position = (pointer: string) => pointer.replace(/\d+/g, index => index.padStart(6, '0'));
+    return [...matches].sort((left, right) => position(left.pointer) < position(right.pointer) ? -1 : 1);
+}
+
+function groupBy(matches: Match[], key: (pointer: string) => string): Match[][] {
```
markscott-ms: The file hand-rolls a Map-based groupBy (lines 21-30) even though lodash's `groupBy` is already imported for partition.

`Object.values(groupBy(matches, m => key(m.pointer)))` may do the same thing

shared/src/spectral/functions/pattern/interface-id-exists-on-node.ts
```diff
@@ -23,13 +24,9 @@ export function interfaceIdExistsOnNode(input: ConnectsRelationship | null | und
...
     const nodeId = input.node;
-    const nodes: object[] = JSONPath({ path: '$.properties.nodes.prefixItems[*]', json: context.document.data as object });
-    const node = nodes.find((node) => {
-        const uniqueId: string[] = JSONPath({ path: '$.properties.unique-id.const', json: node });
-        uniqueId.push(...JSONPath({ path: '$.oneOf[*].properties.unique-id.const', json: node }));
-        uniqueId.push(...JSONPath({ path: '$.anyOf[*].properties.unique-id.const', json: node }));
-        return uniqueId && uniqueId[0] === nodeId;
-    });
+    const nodes: object[] = declarationPaths('nodes').flatMap(path =>
+        JSONPath({ path, json: context.document.data as object }));
```
markscott-ms: `declarationPaths('nodes').flatMap(...)` re-runs three JSONPath queries over the whole node tree, and per the Spectral rule wiring this function fires once per `connects` relationship endpoint (source and destination), with no caching across calls in one document lint. For patterns with many relationships and nodes this multiplies into hundreds of redundant full-tree scans versus the old, narrower per-call lookup.

shared/src/spectral/functions/pattern/ids-are-unique.ts
```diff
@@ -1,25 +1,70 @@
...
+    value: unknown;
+    pointer: string;
+}
+
+/**
+ * The rule blames the second declaration it sees, but three queries per id kind arrive
+ * grouped by query rather than by position. Padding keeps prefixItems/2 before /10.
+ */
+function inDocumentOrder(matches: Match[]): Match[] {
+    const position = (pointer: string) => pointer.replace(/\d+/g, index => index.padStart(6, '0'));
```
markscott-ms: The `padStart(6, '0')` digit-padding trick in inDocumentOrder reinvents a numeric-aware sort that `shared/src/commands/validate/validation-helpers.ts` already does correctly via `localeCompare(..., { numeric: true })`.

The hand-rolled version also silently breaks past 999,999 array indices (though not concerned about this!)

markscott-ms (changes requested):

shared/src/spectral/functions/pattern/interface-id-exists-on-node.ts
YoofiTT96: Every path here is rooted at properties.nodes, so these are lookups rather than full-tree traversals — adding 5,000 relationships to the document leaves the timing unchanged. Main makes one query per pattern and three per node; this makes three per pattern and one per node. Main needs three per node because it receives raw entries and cannot tell whether an entry is a node or a wrapper around alternatives. declarationPaths resolves that before the search, so every object it hands back is a declaration and declaredId reads it in one. 

The general idea is correct in that there is avoidable work.  All three queries run before the search starts, so a node declared in prefixItems, which is the ordinary case, pays for two queries it never needed. I will return at the first path that yields a match instead.

shared/src/spectral/functions/pattern/ids-are-unique.ts
YoofiTT96: Fixed replaced with lodash implementation

shared/src/spectral/functions/pattern/ids-are-unique.ts
YoofiTT96: Helper imported now

markscott-ms (approved):
I've read through the pattern decisions and performed a review of the code and am satisfied that this seems reasonable.

PR #3060, opened 2026-09-08
feat(shared): read items in pattern validation

YoofiTT96:
## Description
A pattern can now declare nodes and relationships under `items` as well as `prefixItems`, so an architecture may *add* members rather than only choose between them. `calm validate` reads the new sites; generation, visualisation and the pattern differ follow separately.

Stacks on #3059, which exists because building this work is what exposed the `prefixItems` gaps. GitHub will not let a cross-fork PR target a branch in my fork, so the base is `main` and the diff carries #3059's commits. **The commits to review are the last six.**

Part of #2859.

`items` members can coexist; `prefixItems` alternatives cannot. A `prefixItems` entry is one position, so one alternative wins. `items` covers every position after it, so any number of its members can be built.

That is what decides the uniqueness rules: two `items` nodes must not share a `unique-id` or an interface id, because both can appear. Two alternatives of one entry may. `exclusiveGroup()` is where this lives.

### Gotcha
`calm generate` does not read `items` yet. An author can declare an `items` node, have `calm validate` accept the pattern, and get an architecture without that node, with nothing said. It closes when the generation PR lands, and `PATTERN-DECISIONS.md` records it meanwhile.

### Rule changes
`pattern-decision-must-be-declared-in-prefix-items` is new. An `items` member may be left out, so a decision declared there can vanish and an answer never gets to decline it.

`pattern-items-must-fit-within-max-items` is new. `maxItems` counts the whole array and the `prefixItems` entries fill it from the front, so a pattern that leaves no room declares a member nothing can ever build.

`pattern-prefix-items-must-declare-one-keyword` becomes `pattern-choice-must-declare-one-keyword`, since it now covers `items` blocks too.

### Ordering
Duplicate ids are blamed on the later declaration. Nothing ranked the two sites, so a pair spanning `prefixItems` and `items` was blamed backwards. Ordering is a property of the site — `prefixItems` fills the first positions and `items` fills everything after — and that rank now joins the parsed indices `byBuildOrder` compares. Every existing ordering test compared two declarations at the same kind of site, which is why none caught it; three that span both are added.

### Tidying
`decisionIsDeclaredInPrefixItems` asked whether a relationship carries options with a JSONPath query over a path that has no wildcard. It now reads that path with lodash off a named constant, beside the one that locates a `unique-id`. Raised in review on #3059 against the same pattern elsewhere.

## Type of Change
- [ ] 📚 Documentation update

Additive. No pattern in the repository changes result, because none of them uses `items` yet.

## Affected Components
- [x] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

`cli/` is the JUnit fixture only. It carries two counts, a `testsuites` total and a `Spectral Suite` total, and both move.

## Commit Message Format ✅
`feat(shared): read items in pattern validation`
`fix(shared): order items declarations after prefixItems entries`
`feat(shared): reject maxItems that leaves no room for items`
`refactor(shared): read the decision marker without JSONPath`

## Testing
- [x] All existing tests pass

161 spectral tests, 659 CLI tests including the end-to-end suite, `shared` at 107 files and 1238 tests, root `npm test` green. Every tracked pattern file was run against the real ruleset on `main` and on this branch; the findings are identical.

Rebased onto #3059 after its review round, so the ordering work here builds on the parsed-index comparator that landed there rather than on the string key it replaced.

## Checklist


shared/src/spectral/rules-pattern.ts
markscott-ms, line 158: node-reference validation for pattern decision options only checks `options.prefixItems`, not `options.items.oneOf`/`anyOf`

markscott-ms, line 170: Relationship-reference validation for pattern decision options only checks `options.prefixItems`, not `options.items.oneOf`/`anyOf`

markscott-ms, line 185: `xor(oneOf, anyOf)` check for decision options only targets `options.prefixItems`, not `options.items`

markscott-ms, line 238: Max-one-item cap on decision options only counts `options.prefixItems.length`, ignoring `options.items`

shared/src/spectral/functions/pattern/items-fit-within-max-items.ts
```diff
@@ -0,0 +1,26 @@
...
+    items?: { oneOf?: unknown[]; anyOf?: unknown[] };
+}
+
+/**
+ * `maxItems` counts the whole array, and the `prefixItems` entries fill it from the front.
+ * An `items` member can only be built in what is left over.
+ */
+export function itemsFitWithinMaxItems(input: unknown, _: unknown, context: RulesetFunctionContext): IFunctionResult[] {
+    const array = input as CalmArray;
+    const offersChoice = Boolean(array?.items?.oneOf || array?.items?.anyOf);
```
markscott-ms: offersChoice uses `Boolean(items.oneOf || items.anyOf)`, which is `true` even for an empty array

markscott-ms (changes requested):
# General problems in this PR:
shared/src/spectral/functions/pattern/is-defined-in-oneof-or-anyof.spec.ts
  ●   1 [test-coverage]  No test exercises a decision option legitimately referencing a node/relationship declared under items


# General notes (some for later PR, recognising that this one only targets spectral validation rules)

calm-hub-ui/src/visualizer/components/reactflow/utils/patternTransformer.ts
  ●  32 [correctness]    Pattern visualizer's getPrefixItems only reads prefixItems, never items, so items-declared nodes/relationships vanish from the rendered graph

calm-models/src/diff/pattern-diff.ts
  ●  50 [correctness]    diffPatterns only inspects prefixItems, never items, when normalising patterns for comparison

calm-plugins/vscode/src/webview/panels/PatternPicker.tsx
  ●  66 [correctness]    VSCode PatternPicker's instantiateFromPattern/extractValue only recurse into prefixItems, never items

shared/src/spectral/rules-pattern.ts
YoofiTT96: Will add rule to cover. Should be defined only in prefixItems

YoofiTT96: Will add rule to cover. Should be defined only in prefixItems

shared/src/spectral/functions/pattern/items-fit-within-max-items.ts
YoofiTT96: Noted

shared/src/spectral/rules-pattern.ts
YoofiTT96: Will add rule to define only in prefix Items

YoofiTT96: Same

cli/test_fixtures/validate_output_junit.xml
```diff
@@ -41,7 +41,9 @@
         <testcase name="nodes-referenced-in-pattern-decision-must-be-in-oneof-or-anyof-block"/>
         <testcase name="relationships-referenced-in-pattern-decision-must-be-in-oneof-or-anyof-block"/>
         <testcase name="pattern-option-relationship-must-only-have-oneof-or-anyof-items"/>
-        <testcase name="pattern-prefix-items-must-declare-one-keyword"/>
+        <testcase name="pattern-choice-must-declare-one-keyword"/>
```
markscott-ms, line 44: JUnit fixture is missing a testcase for the new `pattern-decision-options-must-be-declared-in-prefix-items` rule and undercounts total tests, so the CLI's golden-file e2e tests are failing

markscott-ms (changes requested):

cli/test_fixtures/validate_output_junit.xml
YoofiTT96: Updated

markscott-ms (approved):

Issue #2728, opened 2026-06-23
Timeline bar: long moment names block the expand control and clip the moment cards (no truncation)


rocketstack-matt:
## Bug Report

Two timeline-bar problems in CALM Hub UI with a single shared root cause: **moment labels have no width constraint or truncation.** Long moment names (1) block the collapsed bar's expand control and (2) clip the expanded moment cards.

### Steps to Reproduce:

Open an architecture whose timeline has **long moment names** — e.g.  *"Recommendation 1 — Remove Single Points of Failure (High Availability)"*).

**Bug 1 — timeline won't expand:**
 1. Make sure the timeline bar starts **collapsed** (fresh browser profile, or clear the `localStorage` key `calmHub.timelineExpanded`).
 2. Click the expand chevron (▲ "Expand timeline") at the right of the collapsed timeline strip.

**Bug 2 — moment cards render poorly:**
 1. Expand the timeline panel.
 2. Look at the horizontal row of moment cards.

### Expected Result:
- **Bug 1:** The panel expands to show the moment cards — as it does for an architecture with short moment names (e.g. `#/timeline-demo/architectures/6/3.0.0`, names ≤ 21 chars).
- **Bug 2:** Each moment card stays a readable, bounded size; long names truncate with an ellipsis, full name on hover (the `title` tooltip already exists).

### Actual Result:

**Bug 1:** Nothing happens — the panel cannot be expanded. The only workaround is to navigate to another architecture with short names, expand it there (the expanded state is persisted globally in `localStorage`), then navigate back.

*Root cause (verified):* in the collapsed `Sparkline`, each moment's label/date text is rendered under its dot with no width constraint. Long labels overflow horizontally, and because each moment wrapper is absolutely positioned, the overflowing label text paints **above** the statically-positioned "Expand timeline" button and intercepts its clicks. Verified with Playwright against `main`:
```
document.elementFromPoint(<centre of the Expand button>)
  → <div> "Recommendation 3 — Graceful Degradation & DR"   // the moment label, not the button
A real click on button[aria-label="Expand timeline"] times out (the element is covered).
```
Screenshot — collapsed strip: labels overlap into an unreadable run, and the rightmost label runs over the ▲ expand control (top-right):

[...]

PR #3061, opened 2026-09-08
Slice 1/5: feat(calm-hub-ui): loading states and link-based navigation

jpgough-ms:
## Description

- Extracted from #3001 ("OIDC + SCM (Git) backend support") — that PR bundled these UI changes in with the GitHub storage backend work, with no connection to auth or GitHub
- **Layered PR 1 of 5** splitting #3001 into reviewable slices — see the stacking order and full plan in the tracking comment on #3001
- **This PR targets `main` directly** — it has no dependency on the other slices
- Loading states for the explore rail and mobile nav while counts resolve
- Namespace/type in the section header now render as links
- Sparkline overflow fix (#2728)
- Commit authorship preserved: all commits are attributed to @byrash (original PR author), committed by @jpgough-ms as part of this split

## Type of Change
- [x] 🎨 Code style/formatting changes

## Affected Components

## Testing
- [x] All existing tests pass (lint clean, no test regressions)

## Checklist
- [x] I have added tests for my changes (if applicable) — n/a, existing tests carried over unchanged

---
Split from #3001.

### DIFF
--- a/calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
+++ b/calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
@@ -151,12 +151,13 @@ export function Sparkline({
 
             {/* Track row. A single-version resource has nothing to scrub, so the
                 track is suppressed (the version pill already states what's shown).
-                overflow-hidden clips long labels at the track edge (#2728). */}
+                Individual labels are bounded by maxWidth + ellipsis (#2728); the
+                track itself uses overflow-visible so the last label isn't clipped. */}
             {!singleVersion && (
                 <div
                     data-testid="timeline-sparkline-track"
                     className="relative"
-                    style={{ height: 72, overflow: 'hidden' }}
+                    style={{ height: 72 }}
                 >
                     {/* Inner track wrapper inset 10px each side so dot percentages map directly */}
                     <div className="absolute" style={{ left: 10, right: 10, top: 0, bottom: 0 }}>
--- a/calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
+++ b/calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { Fragment, useEffect, useRef, useState } from 'react';
 import { IoChevronUpOutline } from 'react-icons/io5';
 import { colors } from '../../../../theme/colors.js';
 import { TimelineHeader } from './TimelineHeader.js';
@@ -24,7 +24,7 @@ interface ContextMenuState {
     y: number;
 }
 
-/** Label box width; edge labels are anchored so this can never leave the card (#2728). */
+/** Label box width; labels are clamped to the track so this can never leave the card (#2728). */
 const LABEL_WIDTH = 120;
 
 /**
@@ -65,11 +65,7 @@ export function Sparkline({
     // Progress overlay: left edge to the viewed dot in single mode; between
     // FROM and TO in compare mode.
     const progressLeft = comparing && fromIdx >= 0 && toIdx >= 0 ? pct(Math.min(fromIdx, toIdx)) : 0;
-    const progressRight = comparing && fromIdx >= 0 && toIdx >= 0
-        ? pct(Math.max(fromIdx, toIdx))
-        : viewedIdx >= 0
-            ? pct(viewedIdx)
-            : 0;
+    const progressRight = comparing && fromIdx >= 0 && toIdx >= 0 ? pct(Math.max(fromIdx, toIdx)) : viewedIdx >= 0 ? pct(viewedIdx) : 0;
 
     useEffect(() => {
         if (!menu) return;
@@ -154,21 +150,14 @@ export function Sparkline({
 
             {/* Track row. A single-version resource has nothing to scrub, so the
                 track is suppressed (the version pill already states what's shown).
-                Labels are bounded by LABEL_WIDTH + ellipsis, and the first/last are
-                anchored to their column so they stay inside DiagramSection's
-                overflow-hidden card rather than being sliced (#2728). */}
+                Labels are bounded by LABEL_WIDTH + ellipsis, and clamped to stay
+                inside the track — and so inside DiagramSection's overflow-hidden
+                card — for any dot, not just the true first/last (#2728). */}
             {!singleVersion && (
-                <div
-                    data-testid="timeline-sparkline-track"
-                    className="relative"
-                    style={{ height: 72 }}
-                >
+                <div data-testid="timeline-sparkline-track" className="relative" style={{ height: 72 }}>
                     {/* Inner track wrapper inset 10px each side so dot percentages map directly */}
                     <div className="absolute" style={{ left: 10, right: 10, top: 0, bottom: 0 }}>
-                        <div
-                            className="absolute"
-                            style={{ left: 0, right: 0, top: 15, height: 2, backgroundColor: colors.ink[200] }}
-                        />
+                        <div className="absolute" style={{ left: 0, right: 0, top: 15, height: 2, backgroundColor: colors.ink[200] }} />
                         {progressRight > progressLeft && (
                             <div
                                 data-testid="timeline-progress"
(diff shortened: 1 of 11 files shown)


calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
```diff
@@ -287,7 +289,8 @@ export function MobileNavMenu({ namespaceCounts, domainCounts, onClose }: Mobile
         }
     })();
 
-    const isEmpty = !loading && rows.length === 0;
+    const showLoading = loading || (countsLoading && (view.level === 'root' || view.level === 'namespaces' || view.level === 'domains'));
```
jpgough-ms: The root-level rows ('Namespaces', 'Control Domains', line 240-241) are static and don't depend on either count — but `showLoading` gates `view.level === 'root'` behind `countsLoading` too, and `rows.map` only renders when `!showLoading` (line 322). So opening the mobile nav (which defaults to root) while counts are still loading hides both entry points behind a spinner, when they could render immediately as before this change.

calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
```diff
@@ -151,12 +151,13 @@ export function Sparkline({
...
                 track is suppressed (the version pill already states what's shown).
-                overflow-hidden clips long labels at the track edge (#2728). */}
+                Individual labels are bounded by maxWidth + ellipsis (#2728); the
+                track itself uses overflow-visible so the last label isn't clipped. */}
             {!singleVersion && (
                 <div
                     data-testid="timeline-sparkline-track"
                     className="relative"
-                    style={{ height: 72, overflow: 'hidden' }}
+                    style={{ height: 72 }}
```
jpgough-ms: This relaxes the track's own `overflow: hidden`, but `DiagramSection.tsx` (not touched by this PR) wraps the whole timeline panel in an ancestor `overflow-hidden` — desktop at line 703, mobile at line 738. The last moment's label is centered on a dot at 100% of the track, so if its 120px maxWidth box extends past the track edge, it's still clipped by that ancestor. Worth confirming this actually fixes #2728 in the real DOM — the unit test only renders `Sparkline` in isolation, so it wouldn't catch the ancestor clipping.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -82,28 +84,40 @@ export function ExploreRail({ namespaceCounts, domainCounts, onCollapse }: Explo
...
-                    {filteredNamespaces.map((nc) => (
-                        <RailItem
-                            key={nc.namespace}
-                            label={nc.namespace}
-                            count={nc.total}
-                            active={nc.namespace === activeNamespace}
-                            to={`/namespace/${nc.namespace}`}
-                        />
-                    ))}
+                    {loading ? (
```
jpgough-ms: A single combined `loading` prop gates both the NAMESPACES section here and CONTROL DOMAINS at line 106, but `Hub.tsx` passes `loading={!namespaceCountsLoaded || !domainCountsLoaded}` — one OR'd boolean instead of the two flags it already tracks separately. If namespace counts resolve quickly and domain counts are slow, the namespaces list keeps spinning even though its own data is ready. Passing the two flags through separately would let each section resolve independently.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.test.tsx
```diff
@@ -91,4 +92,21 @@ describe('ExploreRail', () => {
         expect(onCollapse).toHaveBeenCalled();
         await screen.findByRole('link', { name: /finos/ });
     });
+
+    it('shows loading spinners instead of items when loading is true', () => {
+        renderRail('/', undefined, true);
```
jpgough-ms: `screen.getAllByClassName` isn't part of testing-library's `screen` API, so it's always `undefined` and this ternary always takes the else branch. Harmless today since the fallback happens to work, but the intended branch is dead code that a future DOM refactor could silently stop covering.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
jpgough-ms: Confirmed, fixed in f334daa6 — root-level rows no longer depend on countsLoading.

calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
jpgough-ms: Confirmed, fixed in f334daa6 — edge labels now anchor to their column instead of centering, so they can't reach the ancestor's clip boundary.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
jpgough-ms: Confirmed, fixed in f334daa6 — split into namespacesLoading/domainsLoading.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.test.tsx
jpgough-ms: Confirmed, fixed in f334daa6 — removed the dead branch and switched to role-based queries.

jpgough-ms:
One item noticed while fixing the loading-state findings, not addressed here: `namespaceCountsFailed` (Hub.tsx) never reaches ExploreRail or MobileNavMenu, so a failed counts fetch is indistinguishable from "no data" in the UI. Bigger than a loading-state fix and touches Hub's fetch-failure handling more broadly — flagging as a follow-up rather than folding it into this slice.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -82,28 +104,40 @@ export function ExploreRail({ namespaceCounts, domainCounts, onCollapse }: Explo
...
-                            key={nc.namespace}
-                            label={nc.namespace}
-                            count={nc.total}
-                            active={nc.namespace === activeNamespace}
-                            to={`/namespace/${nc.namespace}`}
-                        />
-                    ))}
+                    {namespacesLoading ? (
+                        <RailSpinner label="Loading namespaces" />
+                    ) : filteredNamespaces.length === 0 ? (
```
jpgough-ms, line 118: **Correctness: failed counts fetch renders as "Nothing here", not "unknown"**

- Hub.tsx already tracks this distinction explicitly (comment at Hub.tsx:67-69): "a failed counts fetch means counts are unknown, not zero, so consumers must render them as unknown rather than a misleading 0"
- Trace: fetch rejects → Hub.tsx `setNamespaceCounts([])` + `setNamespaceCountsFailed(true)` → `setNamespaceCountsLoaded(true)` → ExploreRail receives `namespacesLoading=false`, `namespaceCounts=[]`
- Here that renders `<RailEmpty>Nothing here</RailEmpty>`, i.e. it affirmatively asserts the hub is empty on a fetch failure — worse than the old (silently empty) behaviour, and it ignores the invariant Hub already computes
- Same shape applies to the domains branch below (line ~128), though Hub has no `domainCountsFailed" flag to thread through yet
- Fix: pass `namespaceCountsFailed`/an equivalent failed flag down and render an error/unknown state instead of the plain empty state

calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
```diff
@@ -188,11 +193,21 @@ export function Sparkline({
...
+                            // The first/last dot sit 30px from the card edge, so a centred
+                            // label would hang outside DiagramSection's rounded,
+                            // overflow-hidden card and be sliced (#2728). Anchor the edge
+                            // labels to their column so the text grows inwards instead.
+                            const labelBox =
+                                i === 0
+                                    ? { left: 0, textAlign: 'left' as const }
+                                    : i === total - 1
+                                      ? { right: 0, textAlign: 'right' as const }
+                                      : { left: '50%', transform: 'translateX(-50%)', textAlign: 'center' as const };
```
jpgough-ms: **Correctness (edge case): only the true first/last dot is anchored — near-edge dots can still overflow the card**

- Comment at line 27 claims LABEL_WIDTH-bounded labels "can never leave the card," but only `i === 0` and `i === total - 1` get the edge-anchored `labelBox`; every other dot — including `i === 1` and `i === total - 2` — stays centred with ±60px around its dot
- Dot 1 sits at `30 + trackWidth/(total-1)` px from the card edge (30px = the inset + the first dot's offset). Once `trackWidth/(total-1) < 30`, dot 1's centred label bleeds past the same `overflow-hidden` boundary on DiagramSection.tsx's card (line 703) that this PR is fixing for dots 0/total-1 — e.g. ~13+ versions at a 400px track, ~25+ at 800px
- This reproduces #2728's original symptom (a sliced label) for near-edge dots, just shifted from the second dot inward
- The regression test (TimelineBar.test.tsx) only exercises 3 moments, so the one "centered" case tested is also the only non-edge dot — it can't catch this
- Consider anchoring based on proximity to the edge (e.g. `i * spacing < LABEL_WIDTH/2`) rather than exact index equality

calm-hub-ui/src/hub/components/section-header/SectionHeader.tsx
```diff
@@ -45,11 +46,12 @@ export function SectionHeader({ icon, namespace, id, version, typeSegment, right
                 <h2 className="text-base sm:text-xl font-semibold flex flex-wrap items-center gap-x-2 gap-y-1 min-w-0">
                     {icon}
                     {breadcrumbs && <BreadcrumbTrail breadcrumbs={breadcrumbs} onBreadcrumbClick={onBreadcrumbClick} />}
-                    {namespace}
+                    <Link to={`/namespace/${encodeURIComponent(namespace)}`} className="text-accent hover:underline">{namespace}</Link>
```
jpgough-ms, line 49: **Testing gap: new navigation Links have no test coverage**

- Namespace and typeLabel go from static text to real `<Link>`s with computed `href`s (`/namespace/${encodeURIComponent(namespace)}` and `.../namespace/...?type=${encodeURIComponent(typeSegment)}`)
- No test in SectionHeader.test.tsx asserts a `link` role, its `href`, or that clicking navigates — existing tests only check `heading` text content, which passes whether or not the text is wrapped in a working link
- Per repo AGENTS.md: "All new code must have tests (unit and/or integration)" — this is new interactive behaviour (encoding + query-param construction) that could silently break (e.g. wrong query param name/casing vs. `useActiveType`'s `mapTypeInUIToTypeInUrl`) without any test failing
- Suggest adding: `expect(screen.getByRole('link', { name: 'my-namespace' })).toHaveAttribute('href', '/namespace/my-namespace')` and an equivalent for the type link with `?type=...`

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
jpgough-ms: Confirmed, fixed in ae07e8c2 — threaded namespaceCountsFailed through (and added the equivalent domainCountsFailed, which didn't exist), both surfaces now show a distinct message on failure.

calm-hub-ui/src/hub/components/diagram-section/timeline/Sparkline.tsx
jpgough-ms: Confirmed, fixed in ae07e8c2 — replaced the per-index special case with a CSS clamp() applied to every dot, so no label can overflow the track regardless of dot count or width.

calm-hub-ui/src/hub/components/section-header/SectionHeader.tsx
jpgough-ms: Confirmed, fixed in ae07e8c2 — added link role/href coverage, including an encoding case.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -12,10 +12,30 @@ interface ExploreRailProps {
...
+    domainsLoading?: boolean;
+    /** True if the namespace counts fetch failed — distinct from "loaded and empty". */
+    namespacesFailed?: boolean;
+    /** True if the domain counts fetch failed — distinct from "loaded and empty". */
+    domainsFailed?: boolean;
     /** Collapse the rail (keeps the existing sidebar collapse affordance). */
     onCollapse?: () => void;
 }
 
+function RailSpinner({ label }: { label: string }) {
```
jpgough-ms, line 28: RailSpinner's markup (role=status + this exact className) is duplicated verbatim in MobileNavMenu.tsx, added in this same PR. Not a defect, but a future style change now has to be made in two places and can drift.

jpgough-ms:
One more finding from a third review pass (couldn't anchor inline — the diff position isn't resolvable across the intermediate commits):

**`calm-hub-ui/src/hub/Hub.tsx`** — `domainControlCount`/`controlDomainCount` only gated on `domainCountsLoaded`, not the `domainCountsFailed` added in the previous round, so a failed domain-counts fetch rendered a confirmed "0 controls" via DomainPage instead of unknown. The comment directly above claimed this mirrors `activeNamespaceCounts`, which does check the failed flag — this didn't. Fixed in 84827d3d, with a Hub-level test simulating the failure via a spy on the mocked CountsService.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
jpgough-ms: Confirmed, fixed in 84827d3d — extracted a shared LoadingSpinner used by both.

calm-hub-ui/src/hub/components/tree-navigation/MobileNavMenu.tsx
```diff
@@ -287,7 +304,19 @@ export function MobileNavMenu({ namespaceCounts, domainCounts, onClose }: Mobile
...
+    const showLoading =
+        leafLoading ||
+        (view.level === 'namespaces' && namespacesLoading) ||
+        (view.level === 'domains' && domainsLoading);
+    const isEmpty = !showLoading && rows.length === 0;
+    // Distinguish "the fetch failed" from "there's genuinely nothing here" — a
+    // failed counts fetch is unknown, not zero (mirrors Hub's own namespaceCountsFailed).
+    const failed =
+        (view.level === 'namespaces' && namespacesFailed) || (view.level === 'domains' && domainsFailed);
+    const emptyMessage = failed ? "Couldn't load — try again" : 'Nothing here';
```
jpgough-ms: A fourth review pass caught: this promises a retry ("try again") that doesn't exist — Hub fetches counts once on mount with no refetch trigger. Fixed in 117452c5 with level-specific wording matching ExploreRail's equivalent copy.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -82,28 +111,44 @@ export function ExploreRail({ namespaceCounts, domainCounts, onCollapse }: Explo
...
-                            count={nc.total}
-                            active={nc.namespace === activeNamespace}
-                            to={`/namespace/${nc.namespace}`}
-                        />
-                    ))}
+                    {namespacesLoading ? (
+                        <RailSpinner label="Loading namespaces" />
+                    ) : namespacesFailed ? (
+                        <RailEmpty>Couldn&apos;t load namespaces</RailEmpty>
+                    ) : filteredNamespaces.length === 0 ? (
```
jpgough-ms, line 118: A fourth review pass caught: this fires identically whether the filter matched nothing or there are genuinely no namespaces. Fixed in 117452c5 with a distinguishing message.

jpgough-ms:
Two more items from a fourth review pass:

1. **Not addressed — pre-existing, out of scope:** the unknown-vs-zero distinction this slice adds for the namespace/domain counts doesn't reach MobileNavMenu's leaf-level fetches (resource types within a namespace, controls within a domain). Those `.catch(() => finish([]))`/`.catch(() => setLeafItems([]))` blocks predate this PR unchanged — adding `*Failed` tracking for four more service calls is a separate, larger piece of work unrelated to the counts-loading fix this slice makes. Flagging as a follow-up rather than folding it in here.
2. **Not addressed — currently unreachable:** Sparkline's label `clamp()` inverts if the inner track ever drops under 120px (all labels would pin to the left edge). Review verified this isn't reachable at any current desktop (≥1024px) or mobile viewport in this codebase. Left as a known theoretical edge rather than adding a speculative width guard for a layout constraint that doesn't exist yet.

jpgough-ms:
A fifth review pass found two more real, fixed in 1dba0778:

1. Removing the track's `overflow: hidden` (for the #2728 clamp fix) also dropped vertical containment on the `validFrom` date row, which had no bound of its own — an unusually long value could bleed past the track. Gave it the same single-line ellipsis treatment the label row already has.
2. MobileNavMenu's drill-down spinner used a bare "Loading" regardless of section, unlike ExploreRail's specific labels. Now matches.

Five findings from this same pass were architecture/DRY preferences, not defects — declining rather than churning the diff further:
- Generalizing namespace/domain loading+failed state into a shared async-resource abstraction (Hub.tsx)
- A shared hook/component for the loading→failed→empty→list branch, currently hand-written once in each of ExploreRail and MobileNavMenu
- Calling `encodeURIComponent(namespace)` three times in SectionHeader instead of memoizing it once
- RailEmpty (ExploreRail) vs MobileNavMenu's inline empty-state `<li>` being separately maintained
- ExploreRail.tsx crossing this repo's ~150-line component-file guideline (158 lines) by defining RailSpinner/RailEmpty inline rather than as separate files — the reviewer itself flagged this as low-confidence/a judgment call

All are legitimate refactors, but they generalize working code rather than fix something broken, and each component's actual render differs enough (div vs li wrapper, different spacing, different sections) that a shared abstraction adds its own indirection cost. Open to revisiting if a maintainer wants them as a separate follow-up.

rocketstack-matt (approved):
Reviewed independently (own critical + idiom pass, full local build/lint/test run, and a live visual check of the #2728 sparkline fix against a seeded long-name timeline). No defects found beyond what your own five self-review rounds already caught and fixed — all 10 threads check out against the code. Nit: the PR description's Type of Change / Testing checkboxes are stale (still say "refactoring, no functional changes" / tests n/a) relative to what actually shipped across the review rounds — worth a quick update before merge, non-blocking.

PR #3063, opened 2026-09-08
Slice 2/5: build(calm-hub): add observability and GitHub-backend dependencies

jpgough-ms:
## Description

- Extracted from #3001 — the second of 5 layered slices, see the tracking comment on #3001 for the full stack
- **Targets `main`** — #3061 (slice 1 of 5) merged, and GitHub retargeted this PR automatically
- Adds caffeine, jgit `7.2.0`, `quarkus-scheduler`, and the OpenTelemetry/Micrometer stack as dependencies
- All additions are inert until a later slice wires a consumer — no runtime behaviour change
- OpenTelemetry is always compiled in (`quarkus.otel.enabled=true`) and disabled at runtime by default via the genuinely runtime-mutable `quarkus.otel.sdk.disabled` — set `CALM_OTEL_DISABLED=false` on the deployed container to turn it on, no rebuild required

## Review history

Went through 4 rounds of review-and-fix before being marked ready:
1. Comment claimed the test profile disabled both OTEL and Micrometer, but only OTEL was `%test.`-prefixed — Micrometer's core registry stayed active under `mvn test`
2. That same `%test.` prefix only applies when a test's profile name is literally `test` — this module has 4 other custom `QuarkusTestProfile` names, so Micrometer stayed active under `mvn -P integration verify` too
3. **`quarkus.otel.enabled` is build-time-fixed** — verified via `javap` against the Quarkus 3.34.7 jars and empirically (packaged one jar, ran it twice with only the env var changed: no toggle effect either way). The originally-documented `CALM_OTEL_ENABLED=true` production toggle silently never worked in a build-once-deploy-many pipeline. Fixed by switching to `quarkus.otel.sdk.disabled`, which is genuinely `RUN_TIME`-scoped — re-verified empirically that the same built jar now responds live to the env var. Micrometer has no equivalent runtime switch (confirmed via `javap`), so its binder toggles are honestly documented as build-time-only.
4. Clean — no findings

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)
- [x] Dependencies

## Testing
- [x] All existing tests pass — full unit suite (2868 tests) and full Docker-backed integration suite (`mvn -P integration verify`, 540 tests), both 0 failures
- [x] Empirically verified the runtime OTel toggle in both JVM mode (uber-jar) and GraalVM native compilation (Docker image against real MongoDB) — same jar/binary, only the env var changed between runs

## Checklist
- [x] I have added tests for my changes (if applicable) — n/a, no new application code

---
Split from #3001.

### DIFF
--- a/calm-hub/src/main/resources/application.properties
+++ b/calm-hub/src/main/resources/application.properties
@@ -167,6 +167,19 @@ quarkus.micrometer.enabled=true
 quarkus.micrometer.binder.http-server.enabled=${CALM_OTEL_METRICS_ENABLED:false}
 quarkus.micrometer.binder.http-client.enabled=${CALM_OTEL_METRICS_ENABLED:false}
 
-# Test profile: disable OTEL/Micrometer to avoid noise
+# Disable OTEL/Micrometer to avoid noise across every test profile in this module.
+# Both quarkus.otel.enabled and quarkus.micrometer.enabled are build/run-time-fixed
+# per active profile NAME (the same pitfall the jacoco comment above documents) —
+# a %test. prefix only applies when a test's QuarkusTestProfile.getConfigProfile()
+# is literally "test", so each custom profile name used in this module needs its
+# own override or these stay enabled (default true/CALM_OTEL_ENABLED) under it.
 %test.quarkus.otel.enabled=false
 %test.quarkus.micrometer.enabled=false
+%integration-test.quarkus.otel.enabled=false
+%integration-test.quarkus.micrometer.enabled=false
+%nitrite-integration-test.quarkus.otel.enabled=false
+%nitrite-integration-test.quarkus.micrometer.enabled=false
+%secure.quarkus.otel.enabled=false
+%secure.quarkus.micrometer.enabled=false
+%proxy-auth.quarkus.otel.enabled=false
+%proxy-auth.quarkus.micrometer.enabled=false
--- a/calm-hub/src/main/resources/application.properties
+++ b/calm-hub/src/main/resources/application.properties
@@ -156,3 +156,16 @@ calm.mcp.enabled=false
 %dev.quarkus.mcp.server.traffic-logging=true
 quarkus.log.category."org.finos.calm".level=DEBUG
 quarkus.log.category."org.mongodb.driver".level=OFF
+
+# OpenTelemetry — disabled by default, enable via CALM_OTEL_ENABLED=true in production
+quarkus.otel.service.name=${OTEL_SERVICE_NAME:calm-hub}
+quarkus.otel.enabled=${CALM_OTEL_ENABLED:false}
+quarkus.otel.exporter.otlp.endpoint=${OTEL_EXPORTER_OTLP_ENDPOINT:http://localhost:4317}
+quarkus.otel.traces.sampler=parentbased_traceidratio
+quarkus.otel.traces.sampler.arg=${CALM_OTEL_SAMPLING_RATIO:1.0}
+quarkus.micrometer.enabled=true
+quarkus.micrometer.binder.http-server.enabled=${CALM_OTEL_METRICS_ENABLED:false}
+quarkus.micrometer.binder.http-client.enabled=${CALM_OTEL_METRICS_ENABLED:false}
+
+# Test profile: disable OTEL/Micrometer to avoid noise
+%test.quarkus.otel.enabled=false
--- a/calm-hub/src/main/resources/application.properties
+++ b/calm-hub/src/main/resources/application.properties
@@ -169,3 +169,4 @@ quarkus.micrometer.binder.http-client.enabled=${CALM_OTEL_METRICS_ENABLED:false}
 
 # Test profile: disable OTEL/Micrometer to avoid noise
 %test.quarkus.otel.enabled=false
+%test.quarkus.micrometer.enabled=false
--- a/calm-hub/src/main/resources/application.properties
+++ b/calm-hub/src/main/resources/application.properties
@@ -157,29 +157,35 @@ calm.mcp.enabled=false
 quarkus.log.category."org.finos.calm".level=DEBUG
 quarkus.log.category."org.mongodb.driver".level=OFF
 
-# OpenTelemetry — disabled by default, enable via CALM_OTEL_ENABLED=true in production
+# OpenTelemetry — the extension is always compiled in (quarkus.otel.enabled is a
+# build-time flag; a build-once-deploy-many pipeline can't flip it per environment)
+# and disabled at runtime by default via the genuinely runtime-mutable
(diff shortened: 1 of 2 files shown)


calm-hub/src/main/resources/application.properties
```diff
@@ -155,4 +155,17 @@ calm.mcp.enabled=false
...
+quarkus.otel.service.name=${OTEL_SERVICE_NAME:calm-hub}
+quarkus.otel.enabled=${CALM_OTEL_ENABLED:false}
+quarkus.otel.exporter.otlp.endpoint=${OTEL_EXPORTER_OTLP_ENDPOINT:http://localhost:4317}
+quarkus.otel.traces.sampler=parentbased_traceidratio
+quarkus.otel.traces.sampler.arg=${CALM_OTEL_SAMPLING_RATIO:1.0}
+quarkus.micrometer.enabled=true
+quarkus.micrometer.binder.http-server.enabled=${CALM_OTEL_METRICS_ENABLED:false}
+quarkus.micrometer.binder.http-client.enabled=${CALM_OTEL_METRICS_ENABLED:false}
+
+# Test profile: disable OTEL/Micrometer to avoid noise
```
jpgough-ms: Comment claims the test profile disables both OTEL and Micrometer, but only `%test.quarkus.otel.enabled=false` is set — `quarkus.micrometer.enabled=true` two lines up is unprefixed and stays active under `mvn test`, contradicting the stated intent.

jpgough-ms: Confirmed, fixed in d83a8fdc — added %test.quarkus.micrometer.enabled=false.

calm-hub/src/main/resources/application.properties
```diff
@@ -155,4 +155,18 @@ calm.mcp.enabled=false
...
+%test.quarkus.otel.enabled=false
+%test.quarkus.micrometer.enabled=false
```
jpgough-ms: Second review pass caught: `quarkus.otel.enabled`/`quarkus.micrometer.enabled` are both build/run-time-fixed per active profile NAME (verified via javap against the Quarkus 3.34.7 jars) — the same pitfall this file's jacoco comment above already documents. A `%test.` prefix only applies when `QuarkusTestProfile.getConfigProfile()` is literally "test"; this module has four other custom profile names (integration-test, nitrite-integration-test, secure, proxy-auth) under which both properties stayed enabled.

jpgough-ms: Confirmed, fixed in 7713e7d4 — prefixed both properties for all four custom profile names. Verified with the full integration suite (mvn -P integration verify), 540 tests, 0 failures.

calm-hub/src/main/resources/application.properties
```diff
@@ -155,4 +155,31 @@ calm.mcp.enabled=false
 # Log every JSON-RPC message in dev mode for easier debugging of MCP clients.
 %dev.quarkus.mcp.server.traffic-logging=true
 quarkus.log.category."org.finos.calm".level=DEBUG
-quarkus.log.category."org.mongodb.driver".level=OFF
\ No newline at end of file
+quarkus.log.category."org.mongodb.driver".level=OFF
+
+# OpenTelemetry — disabled by default, enable via CALM_OTEL_ENABLED=true in production
+quarkus.otel.service.name=${OTEL_SERVICE_NAME:calm-hub}
+quarkus.otel.enabled=${CALM_OTEL_ENABLED:false}
```
jpgough-ms: **Confirmed:** `quarkus.otel.enabled` (and the OTel/Micrometer properties it sits alongside) is a **build-time-fixed** Quarkus property. Setting `CALM_OTEL_ENABLED=true` as a runtime/container env var on an artifact built with the default (`false`) has **no effect** — OTel stays off, silently, with no warning logged.

Verified empirically in this worktree:
- Packaged with `CALM_OTEL_ENABLED=false` (the default), then ran the jar with `CALM_OTEL_ENABLED=true` — zero OTel export attempts, no activity at all.
- Packaged with `CALM_OTEL_ENABLED=true` baked in at build time, then ran with `CALM_OTEL_ENABLED=false` at runtime — OTel kept actively trying to export (connection-refused warnings to the configured endpoint), proving the runtime override is ignored either direction.

This contradicts the comment on line 160 ("disabled by default, enable via `CALM_OTEL_ENABLED=true` in production"), which describes a deploy-time toggle that the config as written cannot provide — in a typical pipeline (build once in CI, set env vars at deploy), production would never actually get OTel even with the env var set at deploy time. If the intent is a runtime on/off switch, this needs the runtime-mutable counterpart instead (e.g. `quarkus.otel.sdk.disabled`, and the equivalent runtime property for the micrometer binders — worth checking whether one exists for those too), with the current `quarkus.otel.enabled`/`quarkus.micrometer.enabled` left `=true` at build time so the extension code path is compiled in.

jpgough-ms: Confirmed independently via javap against the Quarkus 3.34.7 jars (OTelBuildConfig.enabled() and MicrometerConfig.enabled() are both BUILD_AND_RUN_TIME_FIXED; OTelRuntimeConfig.sdkDisabled() is genuinely RUN_TIME) and empirically: packaged one uber-jar, ran it twice with only CALM_OTEL_DISABLED changed — unset produced zero OTel activity, =false produced live OTLP export attempts (connection-refused, proving the SDK was active). Fixed in 33d6c9bb: quarkus.otel.enabled is now always true (compile in), gated at runtime via quarkus.otel.sdk.disabled (env var renamed CALM_OTEL_DISABLED, inverted polarity, since the old name never worked). Micrometer has no equivalent runtime switch (confirmed via javap: binder .enabled properties nest under the same build-time-fixed root, and the RUN_TIME HttpServerConfig/HttpClientConfig classes have no enabled field) — the comment now says so honestly. Full unit (2868) and integration (540) suites still pass.

jpgough-ms:
**Native-compiled smoke test, run per request** — confirming the runtime toggle actually holds under GraalVM native compilation, not just JVM mode:

- `../mvnw package -Dnative -Dquarkus.native.container-build=true` — **BUILD SUCCESS** in 2:49, no reflection/JNI registration errors from the new deps (caffeine, jgit, quarkus-scheduler, opentelemetry, micrometer). 121.57MB binary.
- Built `Dockerfile.native`, ran the **same image** against a real MongoDB container twice, changing only the env var:
  - Default: zero OTel export activity
  - `CALM_OTEL_DISABLED=false`: live OTLP export attempts (connection-refused, since nothing was listening — proof the SDK was genuinely active)
- Confirmed the native binary actually serves traffic: `GET /q/swagger-ui` → 302, no-auth mode granting the request correctly, `Installed features` lists both `opentelemetry` and `micrometer`.

Same result as the JVM-mode verification in 33d6c9bb — the fix holds under native compilation.

rocketstack-matt (approved):

PR #3064, opened 2026-09-08
Slice 3/5: feat(calm-hub): add a GitHub API response cache

jpgough-ms:
## Description

- Extracted from #3001 — the third of 5 layered slices, see the tracking comment on #3001 for the full stack
- Base auto-retargeted to `main` after #3063 (slice 2) merged
- Adds `GitHubApiResponseCache` (`org.finos.calm.store.github.util`), caching GitHub API responses on behalf of `GitHubVersionService` (slice 5): version lists (5 min TTL) and file content at an immutable commit SHA (365 day TTL)
- Unused until slice 5 wires the consumer
- Originally landed as a generic, calm-hub-wide `CalmCacheService`. Rescoped to be GitHub-only after a design review flagged that a generic-shaped cache with no cross-instance coordination is unsafe to reuse for Mongo/Nitrite-backed data in multi-instance deployments — see #3073 for the full writeup

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass

## Checklist


jpgough-ms:
Checked one item carried over from #3001's review: a prior finding said `calm.cache.ttl.*` was read by nothing. Not applicable to this cache service as it landed here — `CaffeineCacheService.put(key, value, ttl)` takes TTL as a per-call `Duration` argument, not a config property, so there's no `calm.cache.ttl.*` to be dead. No action needed.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubApiResponseCache.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private static <V> void write(Cache<String, V> cache, String key, V value) {
+        if (value == null) {
+            return;
+        }
+        cache.put(key, value);
+    }
+
+    private static String versionsKey(String repoFullName, String filePath) {
```
rocketstack-matt, line 122: Cache keys are built by string concatenation with `:` as delimiter. `repoFullName` can't contain `:` (GitHub forbids it in owner/repo names), so the only reachable collision is a `:` inside `filePath` — not reachable today since CALM document paths don't have those. Still, since this class has no consumer yet, worth closing off for free: a `record VersionsKey(String repo, String path)` (and similarly for content) as the cache key type removes the delimiter question entirely instead of relying on paths never containing `:`. (non-blocking nit)

rocketstack-matt (approved):

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubApiResponseCache.java
jpgough-ms: Merged as-is — good nit, but agreed it's non-blocking. Tracked to pick up on the next slice: replace the `:`-delimited string keys with typed `VersionsKey`/`ContentKey` records so the collision question goes away structurally rather than relying on GitHub's repo-naming rules.

PR #3065, opened 2026-09-08
Slice 4/5: feat(calm-hub): OIDC-driven auth config and VS Code plugin login

jpgough-ms:
## Description

- Extracted from #3001 — the fourth of 5 layered slices, see the tracking comment on #3001 for the full stack
- **This PR targets #3064 (slice 3 of 5)**, not `main` — merge order matters, review it after #3064
- Serves OIDC config to the SPA from the server (`/api/calm/auth/config`) instead of build-time constants
- Adds the VS Code plugin's browser-based OIDC login flow (`PluginAuthResource` + `OidcPluginAuthClient`)
- Lets namespace/domain read access be granted via any `UserAccessStore` namespace grant, not only the existing `UserAccessValidator` path — needed for backends where that validator isn't resolvable
- **Open review items carried forward from the original PR's review** (not fixed in this split, flagged for review here):
  - `authService.tsx` sends the OIDC `id_token` as the API bearer credential for all IdPs — an Entra-ID-specific workaround applied globally
  - `PluginAuthResource` is a public, unauthenticated endpoint holding pending login sessions in an in-memory `ConcurrentHashMap` — no visible expiry/eviction; worth reviewing for session handling and unbounded growth

## Type of Change

## Affected Components
- [x] CALM Hub (`calm-hub/`)

## Testing
- [x] All existing tests pass (133 Java tests, 23 UI tests, lint clean)

## Checklist

---
Split from #3001. Depends on #3064.


calm-hub-ui/src/authService.tsx
```diff
@@ -61,25 +66,26 @@ export async function logout(): Promise<void> {
...
+    if (!userManager) {
         return '';
     }
-    const user = await userManager?.getUser();
+    const user = await userManager.getUser();
     if (user && !user.expired) {
-        return user.access_token;
+        // Entra ID: access_token audience is MS Graph, not our API.
+        // Send the id_token which has our client_id as audience.
+        return user.id_token || user.access_token;
```
jpgough-ms, line 82: Carried forward from #3001's review (still open there): the ID token is sent as the API bearer credential for every IdP, not just Entra ID. An ID token is an authentication receipt for the client, not an authorization credential for a resource server — this is the anti-pattern the OAuth 2.0 Security BCP specifically calls out. Worth scoping this to Entra specifically (or switching to `access_token` once the server-side audience validation supports it) rather than applying the workaround globally.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
```diff
@@ -0,0 +1,231 @@
...
+ * the IdP redirect.
+ */
+@Path("/api/calm/auth")
+@Produces(MediaType.APPLICATION_JSON)
+public class PluginAuthResource {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PluginAuthResource.class);
+    private static final String DEFAULT_REDIRECT_PATH = "/callback";
+
+    private final Map<String, PendingSession> pendingSessions = new ConcurrentHashMap<>();
```
jpgough-ms, line 92: `pendingSessions` is an in-memory `ConcurrentHashMap` on a public, unauthenticated endpoint (see the class javadoc — both `/plugin-login` and `/plugin-callback` are intentionally public). I don't see an eviction policy or TTL on entries here — a login that's started but never completed (browser closed, user abandons the flow) leaves its entry in the map indefinitely. Worth checking whether entries expire, and whether an attacker can grow this map by hitting `/plugin-login` repeatedly without completing the flow.

jpgough-ms: Fixed. `pendingSessions` now has a 10-minute TTL (opportunistic sweep-on-insert, so a login that's never completed no longer lives forever) and a 10,000-entry cap that returns 503 once exceeded. Landed alongside a broader security rework of this endpoint — see the PR diff for the full change.

calm-hub-ui/src/authService.tsx
jpgough-ms: Correct, and confirmed — not scoped to Entra. Tracked as #3077 rather than fixed here: switching the bearer credential to the access token touches every configured IdP's token issuance/scope config, not just this slice's code, so it's out of scope for this PR. Verified nothing new in this slice compounds it — PluginAuthResource's callback page has the same id_token preference, noted in the issue too.

calm-hub/src/main/java/org/finos/calm/security/CalmHubPermissionChecker.java
```diff
@@ -32,6 +34,9 @@ public class CalmHubPermissionChecker {
 
     public CalmHubPermissionChecker(UserAccessStore userAccessStore) {
         this.userAccessStore = userAccessStore;
+    }
+
+    void onStartup(@Observes StartupEvent event) {
```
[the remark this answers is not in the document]
jpgough-ms: Rejected, not fixed — the parameter has to stay for functional reasons, not style. It's what makes this an @Observes CDI startup callback (invoked after field injection completes) instead of constructor logic — the bug the surrounding commit fixes is exactly that the old constructor-based check read authEnabled before @ConfigProperty injection ran. Added a one-line comment on the method explaining this so it doesn't get 'cleaned up' again.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
```diff
@@ -0,0 +1,231 @@
...
+                + "else{document.body.innerHTML='<h2>No token received</h2>';}"
+                + "}catch(e){document.body.innerHTML='<h2>Error</h2><p>'+e.message+'</p>';}"
+                + "})();"
+                + "</script></body></html>";
+
+        return Response.ok(html).type("text/html").build();
+    }
+
+    // Visible for testing
+    Map<String, PendingSession> getPendingSessions() {
```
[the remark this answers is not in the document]
jpgough-ms: Re-checked now that the TTL/cookie rework has landed: not applying Map.copyOf here. getPendingSessions() is package-private and explicitly "Visible for testing" — it has no production caller, only the test suite, which relies on mutating the live map directly to plant/inspect PendingSession fixtures (the class has no other seam to insert one without driving the full plugin-login HTTP flow through a mocked OIDC client). Wrapping it in an immutable copy would break that established test pattern for no production benefit, since nothing outside the test suite ever sees this accessor.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
```diff
@@ -0,0 +1,442 @@
...
+                .type("text/html")
+                .header("Content-Security-Policy", "script-src 'nonce-" + cspNonce + "'; "
+                        + "connect-src " + tokenOrigin + "; base-uri 'none'; form-action 'none'")
+                .header("Cache-Control", "no-store")
+                .header("Referrer-Policy", "no-referrer")
+                .build();
+    }
+
+    // Visible for testing
+    Map<String, PendingSession> getPendingSessions() {
```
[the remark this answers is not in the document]
jpgough-ms: Same disposition as the earlier round on this method (see the reply on the prior comment at this line): getPendingSessions() is package-private, "Visible for testing" only, with no production caller — the test suite relies on mutating the live map directly to plant/inspect PendingSession fixtures. Map.copyOf would break that established pattern for no production benefit.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
```diff
@@ -0,0 +1,442 @@
...
+        NewCookie sessionCookie = new NewCookie.Builder(SESSION_COOKIE_NAME)
+                .value(correlator)
+                .path("/api/calm/auth")
+                .maxAge(SESSION_TTL_SECONDS)
+                .httpOnly(true)
+                .secure(hubBaseUrl.startsWith("https://"))
+                .sameSite(NewCookie.SameSite.LAX)
+                .build();
+
+        return Response.temporaryRedirect(URI.create(authorizeUrl.toString()))
```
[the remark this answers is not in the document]
jpgough-ms: Fixed. The redirect's destination host was never attacker-controlled here — it comes from the discovery document fetched against the configured, admin-set OIDC authority, and this parameter's value is URL-encoded before being appended as a query parameter — but the analyzer is right that it shouldn't have to take that on faith. Added an allowlist regex (REPLAY_GUARD_PATTERN, 1-128 chars from [A-Za-z0-9._-]) validating the nonce parameter before it reaches the redirect URL at all, rejecting anything outside it with 400. That gives the static analysis a validated value instead of a raw request parameter reaching Response.temporaryRedirect.

calm-hub/src/main/resources/application-oidc.properties
```diff
@@ -0,0 +1,58 @@
...
+calm.oidc.http.connect-timeout=${CALM_OIDC_HTTP_CONNECT_TIMEOUT:10}
+calm.oidc.http.request-timeout=${CALM_OIDC_HTTP_REQUEST_TIMEOUT:30}
+
+# Base URL for constructing callback URIs (override in production)
+calm.hub.base-url=${CALM_HUB_BASE_URL:http://localhost:8080}
+
+# Public endpoints specific to this profile (accessible before login or as full-page
+# navigations). /api/calm/auth/config is also permitted profile-independently in the
+# base application.properties; repeated here for readability of what this profile
+# exposes publicly — the base entry alone would be sufficient.
```
rocketstack-matt: This comment is incorrect. Per Quarkus's config-reference guide (https://quarkus.io/guides/config-reference): 'If a profile does not define a value for a specific attribute, the default (no profile) value is used' — the converse also holds: when a profile DOES define the same key, its value fully replaces the base one; scalar/list values are not merged or concatenated. Under the oidc profile, `public.paths` here IS the complete definition — nothing falls back to application.properties's version. If this line is ever removed as 'redundant,' plugin-login/plugin-callback (and config) lose their public permit under oidc. Worth fixing the comment, and ideally a @QuarkusTest asserting 200 on these three paths under the oidc profile specifically, since nothing currently tests this cross-profile behavior.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
```diff
@@ -0,0 +1,470 @@
...
+        }
+
+        LOG.debug("Redirecting plugin auth to OIDC authorize endpoint for port {}", port);
+
+        NewCookie sessionCookie = new NewCookie.Builder(SESSION_COOKIE_NAME)
+                .value(correlator)
+                .path("/api/calm/auth")
+                .maxAge(SESSION_TTL_SECONDS)
+                .httpOnly(true)
+                .secure(hubBaseUrl.startsWith("https://"))
```
rocketstack-matt: This derives Secure from calm.hub.base-url's scheme, not the actual request. OWASP's Secure Cookie Attribute guidance (https://owasp.org/www-community/controls/SecureCookieAttribute) notes the same gap for ASP.NET's equivalent config-driven approach: 'in case TLS is offloaded to a load balancer, the requireSSL solution wouldn't work' — same issue here. Behind the TLS-terminating proxy/ingress this profile's properties file says to expect, an internal CALM_HUB_BASE_URL=http://... would omit Secure on a cookie served over a browser HTTPS connection. Since `quarkus.http.insecure-requests=disabled` already fails plain HTTP closed outside %dev, consider just defaulting this to `true` outside dev instead of deriving it from the base-url string.

rocketstack-matt (dismissed):

calm-hub/src/main/resources/application-oidc.properties
jpgough-ms: Fixed — you're right, the base file's entry does not apply here at all. Quarkus config sources don't merge same-named properties across files; a profile-specific value wholly replaces the base value for that key. Rewrote the comment to say this list is the complete, authoritative definition under `oidc`.

While verifying this I also found that `quarkus.http.insecure-requests=disabled` a few lines below (added in the prior review round) has no paired `quarkus.http.ssl.certificate.*`, so per Quarkus's own docs the app refuses to even start outside `%dev`. Fixed that too — set to `enabled`, since this profile has no TLS-terminating mode of its own and is meant to sit behind a TLS-terminating proxy/ingress, which forwards plain HTTP anyway.

Added `OidcPublicPathsIntegration` (a `@QuarkusTest` under the real `oidc` config profile, following the same pattern as `ProxyAuthIntegration`/`IntegrationTestSecureProfile`) asserting the plugin-auth paths are actually public under this profile, plus a negative control on an unrouted `/api/calm/*` path. Verified the test discriminates by temporarily reverting the fix locally and confirming it fails.

calm-hub/src/main/java/org/finos/calm/resources/PluginAuthResource.java
jpgough-ms: Fixed — replaced `.secure(hubBaseUrl.startsWith("https://"))` with a dedicated `calm.hub.plugin-auth.cookie-secure` config property, `true` by default and relaxed to `false` only under `%dev` (mirroring the existing `%dev.`-relaxation pattern already in this file for `insecure-requests`). Matches the OWASP guidance you cited — the cookie is now Secure by default regardless of what `calm.hub.base-url`'s scheme happens to read internally.

Extended the existing session-cookie unit test to assert `Secure`, and added a test for the `%dev` relaxation.

rocketstack-matt (approved):

PR #3066, opened 2026-09-08
Slice 5/5: feat(calm-hub): GitHub-backed read-only storage for CALM resources

jpgough-ms:
## Description

- Extracted from #3001 — fifth and final layered slice; see the tracking comment on #3001 for the full stack
- **Targets #3065 (slice 4 of 5), not `main`** — review after #3065
- Core: GitHub storage mode (`calm.database.mode=github`) — namespaces map to cloned repos, resources read from an in-memory registry built off the clone, writes return 501 pending account-linking (staged for a later iteration — see PR author's comment on #3001, 2026-08-29)
- Includes producer wiring for all resource types, SHA-based version reads, markdown rendering for raw repo documents
- The 19 store producers aren't split further — the hunk that renames mode-string literals to `DatabaseMode` constants also adds the GitHub branch, so a constants-only slice would reference non-existent classes
- Commit authorship preserved: attributed to @byrash, committed by @jpgough-ms. One follow-up commit (RBAC to Controls, 53f743dd) folded in — its content was already in the final PR head pulled into this slice
- Dropped: a stale JaCoCo exclusion for `security/GitHubOAuthClient` — removed from #3001 in an earlier force-push, class no longer exists

### Structural rework since the original split (Office Hours, 2026-09-10, #3052)

- **Removed "Building Block."** New CALM concept, introduced without agreement — Office Hours: model it as `Standard` instead. `building-blocks/` now aliases to `STANDARD` in both JSON and markdown classifiers; existing repo layouts and the VSCode plugin's glob keep working. **Behaviour change:** `.md` files under `building-blocks/` were previously dropped entirely (no markdown classifier case existed) — now indexed as standards, merged into the same listing as `standards/`
- **Removed `GUIDELINE`** — same problem as Building Block. Nothing served `guidelines/` files, so no regression
- **Reverted `CONTROL` from the namespace-scoped front controller.** Controls are domain-scoped (`ControlResource` is `@Path("/api/calm/domains")`, every `ControlStore` method takes `domain` first); the front controller is namespace-scoped and was passing a namespace into the `domain` parameter — broken in every storage mode:
  - GitHub mode returned **HTTP 500**: `DomainNotFoundException` wasn't in the front controller's catch list. Hidden by test fixtures using the same string for namespace and domain. No test exercised `/calm/namespaces/{ns}/controls/...` before this fix
  - Mongo/nitrite can't represent it: `createMapping` requires a namespace; a control is keyed by `(domain, controlId)`, no namespace
  - Domain-scoped `/api/calm/domains/{domain}/controls/...` (fully implemented) is untouched. Proposing a tracking issue for namespace-addressable controls as its own design — cost: domain-keyed mapping rows, a schema migration in both backends, a backfill
- **Removed `latest` from the shared, backend-agnostic version contract.** Reverses a prior "no concept of latest" decision. `GET .../versions/latest` now 400s in every backend. GitHub mode: an unresolvable/non-SHA version 404s via the resource's `*VersionNotFoundException`, never a silent HEAD read. Version lists contain only real SHAs, or the clone's current HEAD SHA when the API has nothing — assembled once in a shared base store, not duplicated per resource
- **Dissolved `store/github/util` entirely** (#3052 — ~15 runtime CDI beans misfiled as "util," ~20 package-private fields/methods existing only as test seams). Moved into `store.github.{config,sync,registry,api,access}` by responsibility; every store now full constructor injection (final fields, no field pokes); house-standard JavaDoc added throughout. `CalmResourceType` renamed to `RegistryResourceType`, kept deliberately separate from `domain.ResourceType` (different questions: file classification vs. front-controller addressability; `ResourceType` is serialized/name-coupled elsewhere); `UNKNOWN` dropped for `Optional<RegistryResourceType>`
- **Extracted `AbstractGitHubStore` / `AbstractReadOnlyGitHubStore`** (closes the structural half of #3079) — `verifyNamespace`, id lookup, version listing, and the SHA-vs-local-file read block were duplicated across 9 stores and already drifting (`GitHubPatternStore` had inlined two error strings its siblings had hoisted into constants)
- **`CUSTOM_ID_REGEX` now allows uppercase** (`^[a-z]...` → `^[a-zA-Z]...`) — pre-existing change from the original split, flagged here as the public API change it is: custom IDs are case-sensitive in every storage mode
- **Resolved from the original split's open findings:** `store/github/util` duplication and missing shared base class; the 10 near-identical GitHub stores with no shared base
- **Still open** (tracked as issues below): no readiness gate during initial clone; `uniqueId().hashCode() & 0x7FFFFFFF` + `.findFirst()` has no collision detection; `PatternLayoutStoreProducer` has no GitHub branch; `GitHubForkPRService.createPullRequest` always throws; no docs/ADR for the storage backend

## Type of Change
- [x] 🔧 Refactoring (no functional changes)

## Affected Components
- [x] CALM Hub (`calm-hub/`)
- [x] Dependencies

## Testing
- [x] All existing tests pass — calm-hub: 3,380 unit tests + 554 Docker integration tests (5 new, covering this rework's behaviour changes end to end against a real local clone), 0 failures, JaCoCo 90%-per-class gate met; UI suite unchanged by this rework

## Checklist

---
Split from #3001. Depends on #3065.


calm-hub/src/main/java/org/finos/calm/store/github/sync/GitHubStartupInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+            String name = parts[0].trim();
+            String repo = parts[1].trim();
+            String branch = parts.length > 2 ? parts[2].trim() : "main";
+            java.util.Set<String> accessGroups = parts.length > 3 ? parseGroups(parts[3]) : java.util.Set.of();
+            cloneManager.registerNamespace(name, repo, branch, accessGroups);
+            LOG.info("Registered namespace [{}] → repo [{}] branch [{}] accessGroups={}",
+                    name, repo, branch, accessGroups);
+        }
+
+        executor.runAsync(this::cloneAndRebuild);
```
jpgough-ms, line 95: Class javadoc says "health endpoints are available during clone," but there's no `@Readiness`/`HealthCheck` anywhere in calm-hub and no smallrye-health dependency in the pom — so that's describing an endpoint that doesn't exist. Meanwhile `InMemoryRegistryService` starts empty and only gets rebuilt once `cloneAndRebuild` finishes (line 76), so any request for a validly-configured namespace during the clone window hits `verifyNamespace` against an empty namespace list and gets a 404 "namespace does not exist" rather than a 503 "still starting up." Worth adding a real readiness gate here.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public void deleteArchitecture(String namespace, int architectureId) throws NamespaceNotFoundException, ArchitectureNotFoundException {
+        throw new GitHubWriteNotSupportedException(WRITE_UNSUPPORTED);
+    }
+
+    private RegistryEntry findEntryById(String namespace, int id) throws ArchitectureNotFoundException {
+        List<RegistryEntry> entries = registryService.listByType(namespace, CalmResourceType.ARCHITECTURE);
+        Optional<RegistryEntry> found = entries.stream()
+                .filter(e -> (e.uniqueId().hashCode() & 0x7FFFFFFF) == id)
```
jpgough-ms: `uniqueId().hashCode() & 0x7FFFFFFF` used as the public numeric ID, looked up here via `.filter(...).findFirst()`. This is the same pattern repeated across ~21 sites in the GitHub stores (Architecture, BuildingBlock, Control, Interface, Standard, Flow, Pattern, Search, ResourceMapping). A 31-bit hash collision between two resources in the same namespace would silently return the wrong document — `findFirst()` has no way to detect or report the collision. Worth a shared ID scheme with actual collision handling rather than each store rolling its own.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubForkPRService.java
```diff
@@ -0,0 +1,54 @@
...
+            throws PendingWriteException {
+        // Full implementation will:
+        // 1. GET /repos/{upstream}/forks filtered by user -> find or create fork
+        // 2. POST /repos/{user}/{repo}/merge-upstream -> sync fork
+        // 3. POST /repos/{user}/{repo}/git/refs -> create branch
+        // 4. PUT /repos/{user}/{repo}/contents/{path} -> commit file
+        // 5. POST /repos/{upstream}/pulls -> open cross-repo PR
+
+        // Placeholder — will be implemented when GitHub OAuth token exchange is complete
+        throw new GitHubWriteNotSupportedException(
```
jpgough-ms: Confirmed placeholder — this throws unconditionally and has no production callers anywhere in the tree. Consistent with the write path being staged for later (per the PR author's comment on #3001, 2026-08-29), but worth stating explicitly in this PR's description too, since "GitHub as backend" reads as read-write until you find this.

jpgough-ms:
One more finding that doesn't have a line to attach to, since the file isn't touched by this PR's diff:

**`calm-hub/src/main/java/org/finos/calm/store/producer/PatternLayoutStoreProducer.java`** has no `github` branch — only `standalone` (nitrite) and a mongo fallback (lines 20-40). In `calm.database.mode=github`, `GET/PUT /patterns/{id}/layout` resolves to `MongoPatternLayoutStore` with no MongoDB configured, which will fail at request time rather than returning a clean "not supported" response the way the other GitHub stores do. The architecture equivalent, `LayoutStoreProducer` (touched by this PR), does have a github branch — this one looks like it was missed when the GitHub branches were added across the producers.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubAdrStore.java
```diff
@@ -0,0 +1,92 @@
...
+    public AdrMeta updateAdrStatus(AdrMeta adrMeta, Status status) throws AdrNotFoundException, NamespaceNotFoundException, AdrRevisionNotFoundException, AdrPersistenceException, AdrParseException, AdrRevisionExistsException {
+        throw new GitHubWriteNotSupportedException(WRITE_UNSUPPORTED);
+    }
+
+    @Override
+    public void deleteAdr(String namespace, int adrId) throws NamespaceNotFoundException, AdrNotFoundException {
+        throw new GitHubWriteNotSupportedException(WRITE_UNSUPPORTED);
+    }
+
+    private void verifyNamespace(String namespace) throws NamespaceNotFoundException {
```
jpgough-ms: This exact `verifyNamespace` body is duplicated across 10 of the GitHub stores (Adr, Architecture, BuildingBlock, Control, Decorator, Flow, Interface, Pattern, Standard, Timeline), along with the `WRITE_UNSUPPORTED` message constant (12 files) and the hash-based id scheme flagged elsewhere in this review. No shared base class or trait — fixes have to be hand-applied per file, and it's already visibly drifted (`GitHubPatternStore` inlines two version-related error strings its siblings hoisted into shared constants). Worth extracting a common abstract base for the read-only GitHub stores before this pattern gets copied further.

jpgough-ms:
No docs or ADR changes accompany a new storage backend and its config surface (`calm.github.*` in `application-oidc.properties` — clone directory, sync interval, service token, namespace registration format `name|repo|branch[|accessGroups]`). Worth an ADR before this merges, given the design decisions embedded here (in-memory registry rebuilt on a schedule, hash-based ids, read-only-for-now scope).

calm-hub/src/main/java/org/finos/calm/store/github/sync/GitHubStartupInitializer.java
jpgough-ms: Fixed the javadoc — it previously claimed a `@Readiness`/health-check endpoint existed here; there is no smallrye-health dependency in this module, so that was describing something that doesn't exist. It now documents the actual behavior honestly (404 during the clone window, `GitHubCloneManager.getState()` as the real poll signal) rather than papering over it.

The actual readiness gate (503 while cloning, ahead of `verifyNamespace`) isn't implemented yet — tracked in https://github.com/finos/architecture-as-code/issues/3093, which also folds in a second ordering gap found while building this PR's end-to-end integration test: `getState()` flips to READY before the registry is actually rebuilt, so a gate on clone state alone has the same hole. Leaving this thread open pending that issue.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
jpgough-ms: Agreed, and not something to fix piecemeal at this one site — filed https://github.com/finos/architecture-as-code/issues/3080 to cover all ~21 sites with a shared ID scheme that can actually detect a collision, rather than the current per-store `.filter(...).findFirst()`. Leaving this thread open pending that issue.

calm-hub/src/main/java/org/finos/calm/store/github/util/GitHubForkPRService.java
jpgough-ms: Removed outright rather than kept as a documented placeholder — GitHubForkPRService, and PendingWriteException/its mapper, had zero callers anywhere in the tree once every write path settled on GitHubWriteNotSupportedException. This is the last slice in the stack, so nothing downstream needs it back the way an earlier slice needed to keep OidcRoleResolver around.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubAdrStore.java
jpgough-ms: Agreed — filed https://github.com/finos/architecture-as-code/issues/3079 to extract a shared base for the read-only GitHub stores (verifyNamespace, findEntryById, the WRITE_UNSUPPORTED constant, the SHA-vs-local-file read block, and the ID scheme flagged in a sibling thread). Too large a change to fold into this review pass, but worth doing before the pattern gets copied into any further store. Leaving this thread open pending that issue.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
```diff
@@ -0,0 +1,141 @@
...
+    @Inject
+    public GitHubArchitectureStore(InMemoryRegistryService registryService) {
+        this.registryService = registryService;
+    }
+
+    @Override
+    public List<NamespaceResourceSummary> getArchitecturesForNamespace(String namespace, PageRequest page) throws NamespaceNotFoundException {
+        verifyNamespace(namespace);
+        List<RegistryEntry> entries = registryService.listByType(namespace, CalmResourceType.ARCHITECTURE);
+        String repo = cloneManager != null ? cloneManager.getRepoForNamespace(namespace) : null;
```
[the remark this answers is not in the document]
jpgough-ms: Fixed — the unused local is gone; getArchitecturesForNamespace now only computes what it uses.

calm-hub/src/test/java/org/finos/calm/store/github/TestGitHubControlStoreShould.java
```diff
@@ -0,0 +1,366 @@
...
+        when(registryService.getSnapshot()).thenReturn(snapshot);
+        when(registryService.listByType("finos", CalmResourceType.CONTROL)).thenReturn(List.of(accessible));
+        when(accessFilter.getAccessibleNamespaces()).thenReturn(Set.of("finos"));
+
+        assertThrows(ControlNotFoundException.class,
+                () -> store.getRequirementVersions(DOMAIN, HASH_ID));
+    }
+
+    @Test
+    void return_versions_for_control_in_accessible_namespace_with_mixed_access(@TempDir Path tempDir) throws Exception {
```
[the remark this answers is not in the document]
jpgough-ms: Fixed — removed the unused @TempDir parameter.

calm-hub/src/main/java/org/finos/calm/store/github/sync/GitHubStartupInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+    GitHubMetrics metrics;
+
+    @Inject
+    ManagedExecutor executor;
+
+    @Inject
+    @ConfigProperty(name = "calm.github.namespaces")
+    Optional<List<String>> namespaceConfigs;
+
+    void onStart(@Observes StartupEvent ev) {
```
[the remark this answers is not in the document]
jpgough-ms: Declining this one, same reasoning as the identical finding on the previous slice in this stack: the `ev` parameter is what makes this an `@Observes StartupEvent` callback — it's required by the CDI contract, not dead code, even though the method body never reads it.

calm-hub/src/main/java/org/finos/calm/store/github/util/InMemoryRegistryService.java
```diff
@@ -0,0 +1,194 @@
...
+            Instant lastModified = Files.getLastModifiedTime(filePath).toInstant();
+
+            return new RegistryEntry(uniqueId, relativePath, type, name, lastModified);
+        } catch (IOException e) {
+            LOG.debug("Failed to parse file: {}", filePath, e);
+            return null;
+        }
+    }
+
+    private RegistryEntry parseMarkdownFile(Path root, Path filePath, Path relativePath) throws IOException {
```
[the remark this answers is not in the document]
jpgough-ms: Fixed — removed the unused root parameter from parseMarkdownFile and updated the one call site.

jpgough-ms:
Two things flagged for the author's awareness, not fixed as part of this review pass — call is theirs, not mine:

1. **12 `calm-hub-ui` files bundled into commit `861961cc`.** Unrelated to a storage-backend slice on their face — worth checking whether they belong in a separate PR, or whether there's a dependency this slice actually has on them that isn't obvious from the file list alone.
2. **`CUSTOM_ID_REGEX` widened from lowercase-only to `[a-zA-Z]`** in the same commit. A behavior change to an existing public API contract riding along in this diff — worth confirming whether that's intentional.

Full review otherwise covered: a symlink-following arbitrary-file-read vulnerability (fixed), the `@LookupIfProperty` scheduler/startup mode-guard gap (fixed), DOMAIN_READ derived from the registry per the chosen design (implemented + end-to-end tested), a cache branch-keying bug and a `Link`-header host-check gap (both fixed), dead code removed, `ClasspathCoreSchemaStore`'s missing production resources populated, restored javadoc dropped as diff noise, and a round of test-gap closures. Tracking issues filed for what's out of scope for this pass: #3079–#3093.

jpgough-ms:
Update: consolidated the tracking issues down from 15 to 8 — several were symptoms of the same root cause and read better as one ticket than three. Final list, everything out of scope for this review pass:

- #3079 — extract a shared base for the read-only GitHub stores (now also covers the hashCode-based ID collision risk)
- #3081 — audit trail silently dropped in GitHub storage mode
- #3086 — GitHub clone-directory defaults to `/tmp`, `isValidRepo` has no origin-URL check
- #3087 — `quarkus.oidc.tls.verification=none` (confirmed pre-existing, unrelated to this slice)
- #3088 — doc/naming accuracy cleanup (misleading "parallel cloning" javadoc + `GitHubWriteNotSupportedException` naming)
- #3089 — `getFileAtVersion` silently serves HEAD content mislabeled as a historic SHA on a non-200 response
- #3092 — GitHub-mode access/namespace resolution needs a single authoritative path (two independent access-control mechanisms, domain-matching inconsistency, `extractDomain`'s no-subdirectory case, `getGrantsForUser`'s unused argument, dotted-namespace ancestor-chain gap, proxy-auth + github silent denial)
- #3093 — GitHub clone lifecycle has no coordination with concurrent readers (readiness gate + the `reset --hard` race)

CI is green — both `Build Calm Hub` jobs (unit + the full `-P integration` suite, including the new end-to-end DOMAIN_READ test) passed. Only the pre-existing `CanonicalVersion.java` polynomial-redos CodeQL alert fails, confirmed unrelated to this slice (same finding surfaced and confirmed unrelated during slice 4's review).

jpgough-ms:
General comments:

Things we must fix on this PR

* There is a utils package which is usually a design smell and here there are many, these need to be refactored into meaningful classes. The side effect of this is that we end up with a lot of package scoped methods, loss of encapsulation and difficulty in testing. Based on the discussion in #3052 I'd like to propose this refactor goes ahead along with also refactoring all unit tests
* We need to discuss the introduction of Building Block, it looks like a new concept wrapped into the PR
* We need to talk about the introduction of latest, what does that mean across CALM hub, in other design conversation we have been specific about no having the concept of latest

Other items that would be good to tidy up or ensure we have in an issue:

* `CalmResourceType` looks to be duplicated from model into the GitHub package
* A few issues have been raised from the automatic analysis, but one major item for fix afterwards is that lack of concurrency support on operations such as `cloneAll`
* All the Java classes are without JavaDoc, so it's hard to review what they do. As we restructure from a utils based mechanism to something more injectable and well structured, we should consider this too.

jpgough-ms:
In the discussion on office hours we've agreed that `building-blocks` should be modelled with the existing calm type Standard

byrash:
> In the discussion on office hours we've agreed that `building-blocks` should be modelled with the existing calm type Standard

Hello @jpgough-ms , @markscott-ms  & @rocketstack-matt ,  

Upon further reflection we are leaning to using /packs to create a company specific pack that would contain our controls or any one else customized nodes as an extensional model for future enterprises.  The node vs. standard caused us challenges because the standard is a schema.  Would us shifting from building-blocks to packs will suffice?

Thank You.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubAdrStore.java
jpgough-ms: Fixed — extracted AbstractGitHubStore / AbstractReadOnlyGitHubStore, closing the structural half of #3079. verifyNamespace, id lookup, version listing, and the SHA-vs-local-file read block are now defined once and shared across all 9 read-only stores rather than duplicated. Narrowing #3079 to the remaining ID-collision work (see the reply on the numeric-ID thread) rather than closing it outright.

calm-hub/src/main/java/org/finos/calm/store/github/GitHubArchitectureStore.java
jpgough-ms: This PR now extracts the shared read-path base (AbstractReadOnlyGitHubStore), which makes findEntry single-site instead of duplicated per store — so this stays reachable as one small change rather than 9 separate ones. Scoping #3079 to cover the collision-safe ID work alone (reject vs. log is a behaviour decision, not a mechanical extraction) rather than closing it here. Leaving this thread open, pointed at #3079.

jpgough-ms:
> > In the discussion on office hours we've agreed that `building-blocks` should be modelled with the existing calm type Standard
> 
> Hello @jpgough-ms , @markscott-ms & @rocketstack-matt ,
> 
> Upon further reflection we are leaning to using /packs to create a company specific pack that would contain our controls or any one else customized nodes as an extensional model for future enterprises. The node vs. standard caused us challenges because the standard is a schema. Would us shifting from building-blocks to packs will suffice?
> 
> Thank You.

@byrash,

I think that we need to take this into an issue and then into a discussion on office hours. To make it clear, any new introduction of core concepts, needs to be more than footnote in a PR. 

For new ideas to be accepted into open source projects, you need to convince the community that it makes sense to add this, what the use case is, and why it is not supported by anything in CALM at this time. At this point packs, building blocks are new words without being tied into the CALM domain/terminology. The former was covered by standards. 

The other concept in CALM Template to consider is `bundles` not supported by calm-hub at this time, but have been used in description/discussion for code and template generation for a while. I'd want to try compare the two things before introduction.

I'm going to continue to try get the PR in a mergeable shape, but there's a wide array of outstanding issues to get the core of this working. 

cc/ @markscott-ms , @rocketstack-matt

rocketstack-matt:
@byrash perhaps you can explain why standards being a schema is a problem? After all the whole core of CALM is a schema.

To @jpgough-ms's point, we need to make sure that what you need to support your organization makes sense for the whole community and is done in a way that doesn't duplicate functionality that may already exist, where there is something similar, but not quite sufficient we should prefer to extend rather than replace.

Thanks.

byrash:
> @byrash perhaps you can explain why standards being a schema is a problem? After all the whole core of CALM is a schema.
> 
> To @jpgough-ms's point, we need to make sure that what you need to support your organization makes sense for the whole community and is done in a way that doesn't duplicate functionality that may already exist, where there is something similar, but not quite sufficient we should prefer to extend rather than replace.
> 
> Thanks.

Thanks @jpgough-ms and @rocketstack-matt — agreed this deserves a proper issue and office hours discussion. We'll open one. Providing our context here as well for continiuity,

To @rocketstack-matt's question — we've reviewed the Standards tutorial and understand that Standards extend core CALM definitions with organizational requirements (e.g., requiring costCenter, owner on all nodes via allOf composition). We also see that Pattern + Standard layered governance covers "what nodes must exist" + "what properties they must have."

Standards won't work in our scenario because we really need custom nodes that don't exist today — apologies we couldn't share more, though. Looking through the CALM Hub Interfaces (https://calm.finos.org/calm-hub/calm-hub-developer-guide/#the-store-interfaces), our main goal was to distribute a private, concrete set of nodes to all of our internal consumers via CALM Hub. This allows us to avoid continuous VS Code plugin updates with new private nodes.

After some review of the documentation,we feel the best way forward is to introduce a NodeStore alongside Controls, ADRs, etc. in CALM Hub. This will give us centralized management of custom nodes.

byrash:
Hello @jpgough-ms @rocketstack-matt , Create this feature request for NodeStore, Happy to have a discussion in our next  WG session for consensus.

https://github.com/finos/architecture-as-code/issues/3102

Thank You.

rocketstack-matt:
@byrash re: "Standards won't work in our scenario because we really need custom nodes that don't exist today" — I don't think that's accurate, and I want to walk through why before we go further down the NodeStore route.

`node-type` is already free-form in the core schema, not a closed enum: [`core.json#L184-L203`](https://github.com/finos/architecture-as-code/blob/main/calm/release/1.2/meta/core.json#L184-L203) defines it as `anyOf: [enum(9 well-known values), string]`. You can write `"node-type": "feature-store"` today — nothing in CALM needs to change for that.

What you're actually describing — a *shared, enforced* definition of a custom node type across your internal consumers — is what a Standard gives you. A Standard is a schema because a node type is a contract (required properties, shape), not because it replaces the node. The node instance still lives in the architecture document.

Two examples:
- `feature-store` node: Standard = `allOf [core node, {node-type: const "feature-store", retention-days: required, owning-team: required}]`. Instance: a normal node with `"node-type": "feature-store"` plus those two properties.
- `kafka-topic` node: same pattern, gating on `partition-count` / `retention-policy`.

A Pattern then `$ref`s the Standard via `items` (not `prefixItems`), so `calm validate` enforces the shape wherever that node type appears, without dictating what else exists in the architecture. This is the same mechanism already used for decorator standards — see [`deployment.decorator.standard.json`](https://github.com/finos/architecture-as-code/blob/main/calm/draft/2026-03/standards/deployment/deployment.decorator.standard.json), which gates on `"type": {"const": "deployment"}`.

On distribution: CALM Hub already has this — [`StandardStore`](https://github.com/finos/architecture-as-code/blob/main/calm-hub/src/main/java/org/finos/calm/store/StandardStore.java) plus its resource/store implementations. Publishing a new custom node type is publishing a new Standard document to the Hub — no VS Code plugin redeploy, since it's a hosted schema, not code baked into the client. That's the exact distribution mechanism you asked for.

One gap: Standards define types and Patterns mandate their use — neither is a browsable catalogue of pre-built node snippets a user picks from. If what your team actually needs is that catalogue rather than a schema contract, that's closer to what `template-bundles/` already does, and worth comparing before we introduce a new NodeStore concept. Happy to take that as its own issue if the schema-contract approach above doesn't cover your case.

byrash:
> @byrash re: "Standards won't work in our scenario because we really need custom nodes that don't exist today" — I don't think that's accurate, and I want to walk through why before we go further down the NodeStore route.
> 
> `node-type` is already free-form in the core schema, not a closed enum: [`core.json#L184-L203`](https://github.com/finos/architecture-as-code/blob/main/calm/release/1.2/meta/core.json#L184-L203) defines it as `anyOf: [enum(9 well-known values), string]`. You can write `"node-type": "feature-store"` today — nothing in CALM needs to change for that.
> 
> What you're actually describing — a _shared, enforced_ definition of a custom node type across your internal consumers — is what a Standard gives you. A Standard is a schema because a node type is a contract (required properties, shape), not because it replaces the node. The node instance still lives in the architecture document.
> 
> Two examples:
> 
> * `feature-store` node: Standard = `allOf [core node, {node-type: const "feature-store", retention-days: required, owning-team: required}]`. Instance: a normal node with `"node-type": "feature-store"` plus those two properties.
> * `kafka-topic` node: same pattern, gating on `partition-count` / `retention-policy`.
> 
> A Pattern then `$ref`s the Standard via `items` (not `prefixItems`), so `calm validate` enforces the shape wherever that node type appears, without dictating what else exists in the architecture. This is the same mechanism already used for decorator standards — see [`deployment.decorator.standard.json`](https://github.com/finos/architecture-as-code/blob/main/calm/draft/2026-03/standards/deployment/deployment.decorator.standard.json), which gates on `"type": {"const": "deployment"}`.
> 
> On distribution: CALM Hub already has this — [`StandardStore`](https://github.com/finos/architecture-as-code/blob/main/calm-hub/src/main/java/org/finos/calm/store/StandardStore.java) plus its resource/store implementations. Publishing a new custom node type is publishing a new Standard document to the Hub — no VS Code plugin redeploy, since it's a hosted schema, not code baked into the client. That's the exact distribution mechanism you asked for.
> 
> One gap: Standards define types and Patterns mandate their use — neither is a browsable catalogue of pre-built node snippets a user picks from. If what your team actually needs is that catalogue rather than a schema contract, that's closer to what `template-bundles/` already does, and worth comparing before we introduce a new NodeStore concept. Happy to take that as its own issue if the schema-contract approach above doesn't cover your case.

Thanks @rocketstack-matt  for the feedback. The gap that you highlighted is what we wanted to address, we need a pre built catalog of node snippets which includes correct & curated controls with their requirements mapped for Architects to reuse and confirm to our governance process. Thank You. 
PS: we can continue our research on template bundles and can talk about it in our next WG if that meets the needs & address the gaps.

markscott-ms:
This PR was discussed in the OH. The terminology of `Building Block` and the general concept was discussed on the call with multiple organisations deeming it useful..

Some post-OH notes...

- Architecture Building Blocks (ABBs) — abstract, capability/requirement-level definitions (closer to what CALM standard does: defines the constraints, like "integer dimensions, these colours").
- Solution Building Blocks (SBBs) — concrete, selectable, potentially vendor/product-specific implementations that satisfy an ABB, but not yet placed into a specific architecture.

A catalogue of partially-defined nodes maps closely to SBBs — pieces you can pick off the shelf, concrete enough to be selected, but not yet instantiated into a specific design. So "Building Block Catalogue" (or "Solution Building Block Catalogue" if we want to lean on the TOGAF term precisely) reads as standard architecture vocabulary rather than an invented name.

So... `Building Block` is ok with me.

jpgough-ms:
I'm going to move this forward as `Standard`, since that's what's on the PR. We have the other issues in place (#3102 and related), so I expect this to keep evolving over the next few weeks as we dig into more of what our contributors are trying to do.

**Fixes from a final code review**

- Namespace-listing summaries (architectures, flows, standards, patterns, interfaces, timelines, search) no longer show the internal `uniqueId` as a description.
- Search no longer scans unreadable namespaces in its reverse namespace lookup, and now pins one registry snapshot per request.
- `pullAll()` now falls back to a fresh clone when a namespace's clone dir stays invalid, matching `cloneAll()`.
- Version listing no longer treats an all-digit semver version as a SHA.
- Single-ADR GET requests return not-found instead of a write-unsupported error.
- `getRequirementVersions` now fails closed on the same registry race `getRequirementForVersion` already handles.
- Restored the flow/AIGF edge indicator dropped by `FloatingEdge`'s badge-to-label change.

All Java and UI tests, lint, and the JaCoCo coverage gate pass on the updated branch.

byrash:
@jpgough-ms — thank you for the work getting this to a mergeable state and for the code quality fixes.

  I would like to request that we avoid removing the building-block store and its API surface in this PR. Collapsing building blocks into Standards without prior alignment changes a contract that downstream consumers depend on.

  **Why building blocks are not Standards:**

  | | Standard | Building Block (proposed: Node Interface) |
  |---|---|---|
  | **Purpose** | Governance rule or guideline (e.g. "all APIs must use mTLS") | Reusable, composable node definition (e.g. "API Gateway", "Kafka Cluster") |
  | **Consumed by** | Compliance checks, documentation | Architecture composition — drag onto a canvas, wire together, generate |
  | **Structure** | Prose or schema constraints | Concrete node with interfaces, connection points, and metadata |
  | **Lifecycle** | Authored once, referenced in audits | Published to a central catalog, pulled by multiple consumers, versioned independently |

  The core value building blocks bring is a **reusable enterprise node catalog**. An architecture team publishes approved, composable components once in CALM Hub, and every downstream consumer — editors, CLI tools, code generators —
  resolves them centrally. When the catalog evolves, consumers pick up changes without a tooling release. Standards cannot fill this role because they do not carry the node structure, interfaces, or connection points needed for
  composition.

  For reference, Karl shared recordings from the last office hours where we explained the need for this capability in detail.

  **Request:**

  1. **We are open to renaming.** "Building blocks" → "node interfaces" as proposed in #3102 is acceptable — the name is not the concern, the distinct store and API surface is.
  2. **Can we set up a call with @markscott-ms and @rocketstack-matt before the next office hours to align on the path forward?** This would help us move faster.

Issue #3031, opened 2026-08-26
Workspace command suite issues


willosborne:
## Bug Report

Using one issue for a whole list of bugs/missing QoL issues that can be fixed in one go.

- No standards support in `add` or `new` commands
- No way to use workspaces from a folder that isn't a git repo without a fake .git - need to default looking in current directory if no .git found.
- Add puts the full absolute path to the file into the workspace.  This is practically useless. You should either resolve a relative path from the workspace file, or have every path be relative to the workspace root (preferred).
- `rm` makes you specify the english name of the thing to delete. This should probably take the $id, but users will want to use an interactive flow - so add inquirer support
- Templates for `new` are using a very old CALM core schema.

willosborne:
2 and 3 are solved by #3074 as they are most urgent

PR #3074, opened 2026-09-09
fix(cli): allow workspaces in non-git projects and store relative paths in add

willosborne:
## Description

Fixes two of the five issues from #3031 ("Workspace command suite issues"):

- `calm workspace` commands (`init`, `list`, `show`, `switch`, `clean`, `push`, `bump`) required a `.git` directory and hard-failed otherwise. A new `findProjectRoot()` resolver falls back to the given directory (or cwd) when there's no git repo, instead of requiring a fake `.git`.
- `workspace add` stored the full absolute path to a referenced file in the manifest. It now stores a path relative to the bundle directory, matching the convention already used for `--copy`'d files.

The remaining three items in #3031 (standards support in `add`/`new`, stale `new` templates, and fully interactive `rm`) are not part of this PR — tracked separately.

## Type of Change

## Affected Components
- [x] CLI (`cli/`)

## Testing
- [x] All existing tests pass

## Checklist

### DIFF
--- a/cli/src/workspace-resolver.ts
+++ b/cli/src/workspace-resolver.ts
@@ -28,14 +28,47 @@ export function findGitRoot(startPath?: string): string | null {
 }
 
 /**
- * Resolve the root directory workspace state should live under: the git root if one
- * exists, otherwise startPath (or the cwd) itself. This lets workspaces be used from
- * folders that aren't git repositories, without walking further up the filesystem
- * when there's no git boundary to stop at.
+ * Walk upward from startPath looking for an existing `.calm-workspace` directory,
+ * stopping at the nearest one found. The search never goes past gitBoundary (inclusive)
+ * — without that limit, a workspace could be "found" in a wholly unrelated ancestor
+ * project once the walk crosses out of the current git repo. When gitBoundary is null
+ * (no git repo above startPath at all), the walk goes all the way to the filesystem root,
+ * same as findGitRoot.
+ */
+function findExistingWorkspaceRoot(startPath: string, gitBoundary: string | null): string | null {
+    let currentPath = startPath;
+    while (true) {
+        const workspacePath = join(currentPath, '.calm-workspace');
+        if (existsSync(workspacePath) && statSync(workspacePath).isDirectory()) {
+            return currentPath;
+        }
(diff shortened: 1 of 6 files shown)


rocketstack-matt (commented):
A question around how you expect this edge case to work . . .

cli/src/workspace-resolver.ts
```diff
@@ -27,10 +27,20 @@ export function findGitRoot(startPath?: string): string | null {
     return null;
 }
 
+/**
+ * Resolve the root directory workspace state should live under: the git root if one
+ * exists, otherwise startPath (or the cwd) itself. This lets workspaces be used from
+ * folders that aren't git repositories, without walking further up the filesystem
+ * when there's no git boundary to stop at.
+ */
+export function findProjectRoot(startPath?: string): string {
```
rocketstack-matt, line 68: Edge case: `findProjectRoot` re-resolves on every invocation, so a workspace created in a non-git directory becomes silently unreachable if a `.git` later appears in a parent directory — `list`/`show`/`switch`/`add` will start looking under the new (higher) root and report "No workspaces found" even though the workspace still exists on disk at the original path.

Repro: `workspace init` in `/a/b` (no git anywhere) → works; `git init` at `/a` → `workspace list` from `/a/b` now reports nothing found.

Not a blocker for this PR's scope, but worth a note or a defensive check (e.g. prefer a project root that already has a `.calm-workspace` over the git root) as a follow-up. (non-blocking nit)

willosborne: good shout. let me fix that

willosborne: Fixed in 960f61f0 — findProjectRoot now prefers the nearest existing `.calm-workspace` over the git root, bounded to the current git repo so it doesn't accidentally pick up an unrelated ancestor project's workspace. Added tests covering your repro plus the nested-subdirectory variant.

willosborne: fixed

rocketstack-matt (approved):
Verified: tests (137/137), lint, and build all pass on 960f61f0. The earlier edge case (workspace unreachable once a .git appears higher up) is fixed with good regression coverage.

PR #3103, opened 2026-09-11
feat(shared): instantiate items in calm generate

YoofiTT96:
## Description
`calm generate` now honours an answer that selects an `items` member. Before this, the member was silently dropped — the architect asked for it and the architecture came back without it, with no error.

Part of #2859.

### PR sequence
Three stacked PRs, each reviewable alone, in this order:

| | PR | Scope |
|---|---|---|
| 1 | #3059 | `prefixItems` alternatives — the pre-existing gaps that had to be fixed first |
| 2 | #3060 | `items` in pattern validation |
| 3 | this | `items` in generation |

GitHub will not let a cross-fork PR target a branch in my fork, so all three sit on `main` and each diff carries the ones below it. **The commits to review here are the last two.**

#3060 must not merge without this PR. On its own it teaches `calm validate` to accept an `items` node that `calm generate` then drops, and `feat(shared)` cuts a release on merge.

Narrowing **promotes**. A selected `items` member is appended to `prefixItems` and `items` is deleted, so the instantiator only ever reads positions and never learns what `items` is. The alternative — teaching the instantiator a second shape — spreads the change across two stages for no gain.

### Gotcha
A pattern that pins `maxItems` to the number of `prefixItems` entries can never build an `items` member, and the architecture `calm generate` produces then fails its own pattern on `maxItems`. That is an authoring error, and #3060 adds `pattern-items-must-fit-within-max-items` to refuse it up front rather than have generation quietly rewrite the author's bound.

### Tidying
`options.ts` and the decision validation rule asked the same question of the same shape — does this relationship carry `relationship-type.properties.options` — one with optional chaining and one with a JSONPath query. Both now call `declaresOptions`, which #3060 puts beside the helper that reads a `unique-id`. That adds the first `commands/ → spectral/` import in `shared`; `declaration-paths.ts` is pattern-shape knowledge that happens to live under `spectral/`, and the shared reader in `calm-models` is where both will eventually point.

### Also fixed
Three faults reachable without `items` at all: an array with no `prefixItems` produced `{}` where an array belongs, an empty choice list threw on `flatMap`, and a pattern with no `relationships` property threw on `prefixItems`.

## Type of Change
- [ ] 📚 Documentation update

Additive. No pattern in the repository changes result, because none of them uses `items` yet.

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [x] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Commit Message Format ✅
`feat(shared): instantiate items in calm generate`
`refactor(shared): share the decision predicate with validation`

## Testing
- [x] All existing tests pass

Ten new tests across `options.spec.ts` and `instantiate.spec.ts`. Each new guarantee was mutation-tested: disabling the promotion, forcing the choice check true, and restoring the old array condition each turn the matching tests red and leave the rest green.

`options.ts` and `instantiate.ts` are at 100% statements. `shared` passes 107 files and 1246 tests, and root `npm test` is green.

Rebased onto #3060 after the review round on #3059. The shared predicate above replaced two call sites in `options.ts`, not the one the review named — the second was caught by sixteen failing tests, and forcing the predicate true still turns the decision rule's negative case red.

## Checklist


shared/src/commands/generate/components/options.ts
```diff
@@ -119,16 +116,34 @@ function flattenOneOfAndAnyOf(item: Item, selectionPredicate: (item: SchemaNode)
...
+    }
 
     const selectionPredicate = (x: SchemaNode) => ids.includes(x['properties']['unique-id']['const']);
-    pattern['properties'][calmType]['prefixItems'] = calmItems
+    const chosen = (declarations['prefixItems'] ?? [])
         .flatMap((item: Item) => flattenOneOfAndAnyOf(item, selectionPredicate));
+
+    if (isChoice(declarations['items'])) {
+        chosen.push(...flattenOneOfAndAnyOf(declarations['items'], selectionPredicate));
+        delete declarations['items'];
```
markscott-ms: **correctness**: `selectDeclarations` unconditionally deletes a pattern's `items` schema whenever it isn't matched by the current choice set, and this function is reused during architecture validation (not just generation), silently loosening strictness.

`applyArchitectureOptionsToPattern` (shared/src/commands/validate/validation-helpers.ts:67-73), invoked from the live `calm validate` architecture-with-pattern path (shared/src/commands/validate/rules/json-schema-rule.ts:50), calls `selectChoices(pattern, choices, debug)` whenever the architecture has at least one relationship carrying `relationship-type.options` — regardless of whether that decision has anything to do with an unrelated `nodes.items`/`relationships.items` block elsewhere in the pattern. Inside `selectDeclarations`, `ids` is built from *all* choices in the architecture; if none of those ids match the items alternatives, `delete declarations['items']` still runs and strips the items schema from the pattern used to compile the AJV validator. An array with `prefixItems` but no `items` places no constraint on elements beyond the prefix, so additional nodes/relationships the architecture adds are no longer checked against the pattern's `items` schema — `calm validate` passes architectures it should reject. Before this PR, `items` was never touched by `selectChoices`.

shared/src/commands/generate/components/instantiate.ts
```diff
@@ -99,9 +99,11 @@ async function instantiateFromProperties(
     for (const [key, def] of Object.entries(properties)) {
         const resolvedDef = await resolveSchema(def as JsonSchema, schemaDir);
 
-        if (resolvedDef.type === 'array' && resolvedDef.prefixItems) {
+        if (resolvedDef.const !== undefined) {
```
markscott-ms: **correctness**: Moving the `const !== undefined` check ahead of the `type === 'array'` check reverses precedence for a resolved schema that carries both, silently dropping prefixItems-built content in favour of a stale const.

`resolveSchema` (lines 18-47) merges a `$ref`-resolved definition with the local `def` via `{...resolved, ...def}`; if the referenced schema defines a top-level `const` while the local pattern definition adds `prefixItems` (or vice versa), the merged `resolvedDef` ends up with both fields. The old code always built the array from `prefixItems` when `type==='array' && prefixItems`, ignoring any stray `const`. The new code checks `const !== undefined` first and returns that value directly, skipping the `prefixItems` branch entirely — `output[key]` becomes the const value instead of the instantiated array, and any real prefixItems entries are silently discarded. Nothing in this diff establishes const and array+prefixItems are mutually exclusive on a resolved schema, and no test exercises this combination.

shared/src/commands/generate/components/options.ts
```diff
@@ -119,16 +116,34 @@ function flattenOneOfAndAnyOf(item: Item, selectionPredicate: (item: SchemaNode)
...
+    if (!declarations) {
+        return;
+    }
 
     const selectionPredicate = (x: SchemaNode) => ids.includes(x['properties']['unique-id']['const']);
-    pattern['properties'][calmType]['prefixItems'] = calmItems
+    const chosen = (declarations['prefixItems'] ?? [])
         .flatMap((item: Item) => flattenOneOfAndAnyOf(item, selectionPredicate));
+
+    if (isChoice(declarations['items'])) {
```
markscott-ms: **correctness**: The only way a pattern-declared `items` member's unique id ever appears in a `CalmChoice` is if a pattern author manually copies it into an unrelated relationship-options choice's `nodes`/`relationships` const array — `extractOptions` never scans `items` blocks itself, so the new promotion feature can silently never fire.

A pattern author declares `nodes: { prefixItems: [...], items: { oneOf: [cacheNode, queueNode] } }` expecting `calm generate` to let the user pick cache/queue. `extractOptions` (options.ts:74-92, unchanged by this PR) only derives `CalmOption`s from relationships whose `relationship-type.properties.options` is set; it never looks at `nodes.items`/`relationships.items`. `promptUserForOptions` therefore never prompts about cache/queue, so no `CalmChoice` ever contains their ids, `selectDeclarations`'s `selectionPredicate` never matches them, and they are silently dropped via the unconditional delete — with no error and no indication to the architect that the feature didn't engage. PATTERN-DECISIONS.md's new text doesn't document this precondition.

shared/src/commands/generate/components/options.ts
```diff
@@ -119,16 +116,34 @@ function flattenOneOfAndAnyOf(item: Item, selectionPredicate: (item: SchemaNode)
         .filter((x: SchemaNode) => selectionPredicate(x));
 }
 
-function flattenCalmItems(pattern: SchemaNode, calmType: 'nodes' | 'relationships', ids: string[]): void {
-    const calmItems = pattern['properties'][calmType]['prefixItems'];
+function isChoice(items: Item | undefined): items is Item {
```
markscott-ms: **maintainability**: `isChoice` re-derives the same oneOf/anyOf test that `flattenOneOfAndAnyOf` already performs a few lines below (inverted), giving the file two independent definitions of "is this an alternatives block". If a third alternative keyword is ever added or one deprecated, it's easy to update one and forget the other, silently desyncing items-promotion eligibility from prefixItems-flattening eligibility with no compiler or test signal.

shared/src/commands/generate/components/options.ts
```diff
@@ -1,4 +1,5 @@
 import { initLogger } from '../../../logger';
+import { declaresOptions } from '../../../spectral/functions/pattern/declaration-paths';
```
markscott-ms: **maintainability**: `options.ts` hand-rolls the oneOf/anyOf keyword pair inline (in `isChoice` and `flattenOneOfAndAnyOf`) instead of reusing `ALTERNATIVE_KEYWORDS` from `declaration-paths.ts`, the very module this diff newly imports `declaresOptions` from. The accepted-keyword set now lives in two unlinked places, so a future change has two spots to update.

shared/src/commands/generate/components/options.ts
```diff
@@ -119,16 +116,34 @@ function flattenOneOfAndAnyOf(item: Item, selectionPredicate: (item: SchemaNode)
...
+        return;
+    }
 
     const selectionPredicate = (x: SchemaNode) => ids.includes(x['properties']['unique-id']['const']);
-    pattern['properties'][calmType]['prefixItems'] = calmItems
+    const chosen = (declarations['prefixItems'] ?? [])
         .flatMap((item: Item) => flattenOneOfAndAnyOf(item, selectionPredicate));
+
+    if (isChoice(declarations['items'])) {
+        chosen.push(...flattenOneOfAndAnyOf(declarations['items'], selectionPredicate));
```
markscott-ms: **style**: `selectDeclarations` builds `chosen` functionally via `.flatMap` and then switches to an imperative `chosen.push(...)` to add items alternatives, inconsistent with every other transform in the file. Matching the existing expression style used by `flattenOptionsRelationships` would be cleaner (cosmetic only).

markscott-ms (commented):
Automated code review (effort: high).

One additional finding that falls outside the diff's changed lines, so it's noted here instead of inline:

**consistency** (`shared/src/commands/generate/components/instantiate.ts`, `instantiateObject`): The "array with no prefixItems should yield `[]` not `{}`" fix was applied only to `instantiateFromProperties` (top-level pattern properties); the structurally parallel array-handling loop in `instantiateObject`, one call frame deeper, still gates on `valueDef.type === 'array' && valueDef.prefixItems` with no `?? []` fallback and was left untouched. Not a regression from this diff, but the two near-duplicate array-instantiation loops have now drifted further apart in behaviour, making the file easier to under-fix again next time.

markscott-ms (changes requested):

markscott-ms (approved):

Issue #2933, opened 2026-08-03
Nest decision boxes and containers in pattern visualisation (follow-up to #2859)


YoofiTT96:
## Summary

Follow-up to #2859 / #2932. When visualising a pattern, a node belongs to one box only: either a container (`deployed-in`/`composed-of`) or a decision box, never a decision box nested inside a container. This flat, single-parent model cannot draw a choice whose candidates live inside a container.

## Background

`patternTransformer.ts` gives each node one parent. Decision boxes are always created at the top level and never given a container parent, and container nodes are parented only from the container map. Neither of the two clean shapes below can be drawn.

## The two shapes to support

- **Container holds a decision.** The candidates are leaf nodes sharing one container. Nest the decision box inside that container: container → decision box → candidates.
- **Decision between containers.** The candidates are themselves containers. Nest the containers inside the decision box: decision box → container → contents. This is the cleaner authoring model for "choose which subsystem".

## Current interim behaviour, shipped in #2932

Container precedence is flat. A node that is both a decision candidate and a container child renders inside its container, and a decision box left empty by that is suppressed. So when every candidate sits in the same container, the "choose one" cue is lost; when only some are containerised, the alternatives are split between the container and a floating box. This issue replaces that with proper nesting.

A decision whose candidates are *themselves* containers now keeps its box and its prompt (#2932), but the box is drawn empty and beside them rather than around them. Nesting is what makes it read correctly.

## Also in scope: two decisions sharing a candidate

#2932 keyed decision groups per decision, so two decisions drawing from one catalog now render as two boxes. What it could not solve is a candidate named by **two** decisions: a node has one parent, so the first decision keeps it and the second renders with whatever is left. A decision whose candidates are *all* claimed by an earlier one renders no box at all.

Both are pinned by tests in `patternTransformer.test.ts` as the current limit, not as desired behaviour. Whatever model this issue lands has to answer it — a candidate that belongs to two decisions is exactly the case a single-parent tree cannot express.

## Design sketch

- Compute, per decision group, one "where does this group belong" answer from three inputs already available in `createReactFlowNodes`: each node's container parent (`parentMap`), its decision-group membership (`decisionGroupId`), and whether it is itself a container (`containerNodeIds`). Assign parents top-down: container, then decision box where applicable, then leaves.
[...]

PR #3110, opened 2026-09-14
feat(calm-hub-ui): render items catalogues in pattern diagrams

YoofiTT96:
## Description
A pattern may declare nodes and relationships under `items` as well as `prefixItems`. The renderer read only `prefixItems`, so a catalogue node never appeared on the diagram and a decision naming one drew no box at all. The pattern in #2859 rendered two nodes and one edge, with the catalogue and the decision missing entirely.

Now the catalogue draws as a decision box carrying the author's prompt and each answer, with the catalogue nodes inside it.

Part of #2859.

### PR sequence
Four stacked PRs, each reviewable alone, in this order:

| | PR | Scope |
|---|---|---|
| 1 | #3059 | `prefixItems` alternatives — the pre-existing gaps that had to be fixed first |
| 2 | #3060 | `items` in pattern validation |
| 3 | #3103 | `items` in generation |
| 4 | this | `items` in the pattern diagram |

GitHub will not let a cross-fork PR target a branch in my fork, so all four sit on `main` and each diff carries the ones below it. **The commit to review here is the last one.**

`items` is read as one more choice block, nothing else. Its members get a decision group exactly as `prefixItems` alternatives do, so the existing parenting and the existing options lookup handle them with no further change. That keeps this PR to the reading side of the transformer and leaves the parent-resolution model alone.

A decision is read from `prefixItems` only. #3060 rejects a decision declared in `items`, so the renderer does not tolerate what validation refuses.

### Gotcha
This makes `items` visible; it does not change how decision boxes relate to containers. A catalogue node that also sits in a `deployed-in` container still parents to the decision box, leaving the container drawn empty beside it. That shape was unreachable before, because the catalogue rendered nothing at all, so this PR makes an existing limitation newly visible. #2933 owns the fix and has been updated with the current behaviour.

### Tidying
The `allOf` lookup and the `oneOf`/`anyOf` read were each written out four times. Both are now one function, which is most of the line count in the diff.

## Type of Change
- [ ] 📚 Documentation update

Additive. A pattern that uses no `items` renders exactly as before.

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Commit Message Format ✅
`feat(calm-hub-ui): render items catalogues in pattern diagrams`

## Testing
- [x] All existing tests pass

Seven new tests in `patternTransformer.test.ts`, covering a catalogue node rendering, its grouping into one box, the prompt attaching, an `anyOf` catalogue's label, a `prefixItems` decision staying separate from the catalogue box, a relationship under `items`, and a plain `items: { $ref }` correctly ignored.

Mutation-tested: stop reading the catalogue and five of the seven fail while the rest stay green.

`calm-hub-ui` passes 118 files and 1450 tests. Lint and build are clean, and the full root `npm test` is green.

Rendering is not viewport-dependent — the transformer returns the same nodes and edges at any width — so no mobile-specific case applies here.

## Checklist


calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
markscott-ms: isPatternData's docstring says pattern detection only checks properties.nodes.prefixItems, but the function now also matches on nodes.items.

markscott-ms (changes requested):

calm-hub-ui/src/visualizer/components/drawer/Drawer.tsx
YoofiTT96: Fixed

markscott-ms (approved):

PR #3119, opened 2026-09-16
feat(calm-hub-ui): nest namespaces in the Explore rail

### LINKED ISSUE #2787: calm-hub-ui: represent the namespace hierarchy in the Explore rail (collapsible dot-prefix tree)
### Target Project:
`calm-hub-ui/` — the CALM Hub **Explore rail** (`src/hub/components/explore-rail/ExploreRail.tsx`, `RailItem.tsx`, `RailSectionLabel.tsx`, `CountBadge.tsx`).

### Description of Feature:
Represent the **namespace hierarchy** in the Explore rail as a collapsible dot-prefix tree, so nested namespaces (`finos.calm`, `finos.traderx`) sit *beneath* their parent (`finos`) rather than as flat siblings.

This addresses feedback from **@markscott-ms** on the redesign epic #2754:

> "how do the hierarchy of namespaces get represented? Still nested? … [they] are no longer nested … `finos.calm` and `finos.traderx` are not 'beneath' `finos`, which may become unwieldy in a large enterprise."

The redesign (#2761–#2766) replaced the old Explore tree with a flat one-level rail. That reads cleanly at demo scale but loses the "these belong together" grouping the tree afforded, and does not scale to an enterprise with hundreds of namespaces several dot-levels deep.

A design exploration (produced in Claude Design — *CALM Namespace Hierarchy Browser*) compared three directions and recommends **Direction A: a collapsible dot-prefix tree, with the filter flattening to matches (the hybrid)**. The current flat state is captured in `current-state-flat-rail.png` on the `design/namespace-hierarchy` branch of the rocketstack-matt fork.

**Recommended design — the live rail** (`finos` expanded with `calm` / `traderx` nested beneath it, active row, indent guides, own-count pills):

Recommended nested rail — finos expanded with calm and traderx nested beneath, indent guides and count pills

**Every state** — a collapsed branch showing the ghost `+8` nested-count chip; filtering `tra` (full names shown so `traderx` vs `finos.traderx` stay distinct, `finos` auto-expanded as ancestor, match highlighted); and the collapsed rail (roots as initials):

States: collapsed branch with +N chip, filtering with highlighted matches and auto-expanded ancestor, and the collapsed rail as initials

### User Stories:
- As a **platform architect in a large org**, I want namespaces grouped by their dotted prefix so I can scan and collapse whole areas of the catalogue instead of scrolling a flat list of hundreds.
- As a **CALM Hub user**, I want to open a parent namespace's *own* page **and** expand its children independently, because a prefix like `finos` both owns artefacts and parents others.
- As a **user searching**, I want to type a fragment and see every matching namespace at any depth surface immediately, with its ancestors expanded, without losing which full namespace each match is.

### Current Limitations:
`ExploreRail` renders a flat list — `filteredNamespaces.map(nc => <RailItem label={nc.namespace} …/>)` — with `label` set to the full dotted name. There is no tree construction, no expand/collapse, and no parent/child relationship; hierarchy information latent in the dotted names is discarded.

### Proposed Implementation:

**Recommended design — Direction A (collapsible dot-prefix tree + search-flatten hybrid).** All required design elements:

**Tree derived purely from the dotted name.** Split each namespace on `.`; `a.b.c` nests under `a.b` under `a`. No parent field exists or is needed. Sort children alphabetically (`localeCompare`) at each level.

**Two hit targets per row — the key to parent-is-also-a-leaf:**
- **Chevron** — toggles children only; a separate target that never navigates.
- **Label** — navigates to `/namespace/:ns` (the prefix's *own* page). Stays a router `<Link>`, so the active row remains URL-derived and **one row = one namespace** (highlight can't drift).

**Counts — two visually distinct pills, never ambiguous:**
- **Own count** — *solid* pill (accent when active, grey otherwise); identical meaning to today (artefacts in *this* namespace); shown on every real row.
- **Nested count** — *ghost* (outlined, muted) `+N` pill; **collapsed-only**; sums artefacts hidden in descendants (e.g. collapsed `finos` shows `+8` = 5 + 3). Disappears on expand, since they're then visible.

**Legibility at the narrow rail width (~236px):**
- **Prefix elision** — collapsed/normal view shows a child's last segment only (`calm` under `finos`); full dotted name on `title` hover.
- **Indent guides** — a hairline vertical rule per depth level.

**Filtering** — typing in "Filter namespaces": matches surface at any depth, **ancestors auto-expanded**, **full names shown** (so `traderx` vs `finos.traderx` stay distinct), and the matched substring is highlighted.

**Synthetic parent (no own page)** — if `platform.payments.ledger` exists but `platform` does not, derive `platform`/`payments` as group-only rows: muted italic, **no own pill, label not clickable — only the chevron works**.

**Collapsed rail** — roots become single-letter initials; the active namespace keeps its accent bar; hover opens a fly-out of that subtree.

**Mobile (rail → drawer)** — one code path, breakpoint switch. Two enlarged tap zones both ≥44px: the chevron becomes a ~40px column with a hairline divider, the label fills the rest. **Tap chevron → expand/collapse in place, drawer stays open** (browsing structure); **tap label → navigate to `/namespace/:ns` and dismiss the drawer** (chosen a destination). Expansion state persists across open/close.

**Tunable props (A/B without a rebuild):** `accent` colour, `elidePrefixes` (bool), `showNestedChips` (bool).

**Recommended rail — expanded (`finos` open, active):**
```
NAMESPACES
▾ finos                       [10]
  │ calm                       5
  │ traderx                    3
  workshop                    13
  traderx                      6
  ai-governance-v2             2
  timeline-demo                6
```
**Collapsed branch (`finos` closed) — own count stays, ghost chip flags the hidden subtree:**
```
▸ finos                  (+8) [10]
  workshop                    13
  …
```
**Filtering "tra" — matches at any depth, full names, ancestor auto-expanded:**
```
▾ finos                       [10]
  │ finos.·tra·derx            3
  ·tra·derx                    6
```
**Row anatomy (collapsed parent):** `▸`  `finos`  `(+8 ghost)`  `[10 solid]` — chevron · label(→ /namespace/finos) · nested-count · own-count.

- **Technical design considerations:** build the tree in `ExploreRail` (or a small `useNamespaceTree(namespaceCounts)` hook) from `namespaceCounts`; flatten to a render list honouring expansion + filter; keep expansion in local state (persisted across the mobile drawer open/close). No new data model — hierarchy is a pure function of the dotted names already returned by `/api/calm/namespaces/counts`.
- **API changes:** none. Nested counts are derived client-side by summing descendants.
- **Data model changes:** none.
- **Dependencies:** `CONTROL DOMAINS` section stays flat and unchanged.

### Alternatives Considered:
- **B — Indented grouping, no collapse** (always-visible children with elided prefix). Zero clicks, but doesn't scale (hundreds of rows still scroll) and a header that is *also* a namespace muddies "which row is selected." *Rejected as the primary; A degrades toward it.*
- **C — Drill-down + breadcrumb** (roots only; click a prefix to descend). Scales best for very wide/deep trees but adds navigation steps and hides cross-group context; overkill at today's single-dot-level scale. *Kept as a "focus" affordance A can borrow for pathologically deep subtrees.*
- **Do nothing (keep flat).** Rejected — the concern @markscott-ms raised (enterprise scale) only grows.

### Testing Strategy:
- **Unit (vitest):** tree construction from dotted names (incl. synthetic parents, `traderx` vs `finos.traderx` disambiguation); expand/collapse toggles only the chevron's subtree; label navigates to `/namespace/:ns` and is the single active row; own vs nested count pills (nested only when collapsed, sums descendants, hidden when expanded); prefix elision + full-name `title`; filter surfaces matches at any depth with ancestors expanded and full names; synthetic-parent row is not clickable.
- **Interaction/a11y:** chevron and label are distinct, keyboard-reachable targets; mobile tap zones ≥44px; drawer expansion persistence.
- Regression: `CONTROL DOMAINS` section unchanged.

Update any Explore-rail screenshots in the CALM Hub docs to show the nested tree.

### Implementation Checklist:
- [ ] Documentation updated

### Additional Context:
- Follow-up to the redesign epic #2754; addresses @markscott-ms's nested-namespace comment there.
- Design exploration (three directions, recommendation, live rail across all states, mobile flow, count/parent-as-leaf anatomy) produced in Claude Design — *CALM Namespace Hierarchy Browser*.
- Current flat-rail screenshot + brief on the `design/namespace-hierarchy` branch (rocketstack-matt fork): `current-state-flat-rail.png`, `namespace-hierarchy-brief.md`.
- Out of scope: hiding zero-count artefact-type tabs on the namespace page (a separate point in the same #2754 thread) — the current "auto-select first non-empty type" behaviour is intentional and kept.

YoofiTT96:
## Description
A CALM Hub namespace has a dotted name. The hierarchy is therefore already in the name. The Explore rail did not use it. It drew `finos.calm` and `finos.traderx` as siblings of `finos`, not below it. This is easy to read at demo scale. It does not scale to an enterprise with hundreds of namespaces at several levels. @markscott-ms raised this point on the redesign epic #2754.

The rail now nests rows on the dot prefix, and the filter flattens the tree to the matches. Collapsing the rail keeps that structure as a strip of root initials with a fly-out. The mobile drawer shows the same tree.

Closes #2787.

### Visual Sample
image<figcaption>Nested View</figcaption>

Screenshot 2026-09-16 at 3 32 50 PM<figcaption>Filtered Search</figcaption>

image<figcaption>Collapsed View with pop out. Namespaces with same first character appear duplicated however the pop out provides a form of disambiguation
</figcaption>

image<figcaption>Mobile View</figcaption>

### Decisions worth knowing

**A row has two hit targets.** The chevron opens and closes the subtree. The label opens the page of that namespace. A prefix such as `finos` owns artefacts and also parents other namespaces, so a user must be able to do each action separately. This is why a namespace row is a new component. The existing `RailItem` is a single link, and it stays untouched for the `CONTROL DOMAINS` rows.

**Two counts, two weights.** The solid pill is the own count of the row, unchanged from today. The ghost `+N` pill is the sum the descendants hold, and a row shows it only while closed. The two add. `finos` owning 10 with children of 5 and 3 reads `+8` and `[10]`, so the branch holds 18.

**The rail stores the closed rows, not the open ones.** Every row is open by default, so each namespace visible before this change is still visible after it. Storing the open rows would show only roots on a first load, which reads as lost data.

**Filtering never writes to that stored set.** A filter opens ancestors to reveal a match, but it does so in the render only. Clearing the filter restores exactly what the user chose.

**No `role="tree"`.** That ARIA pattern wants one focusable element per row under a roving tabindex, which is incompatible with the two hit targets above. The rows are disclosure buttons and links in the natural tab order instead. This codebase has no `role="tree"` anywhere.

### Reuse
The trie builder is recovered from `TreeNavigation.tsx`, which commit `61047513` deleted when the flat rail replaced the old tree, along with its unit tests. Two changes are deliberate: it no longer compresses paths, so each segment gets a row, and it now sorts siblings rather than relying on insertion order.

The client tree agrees with the rule the server already uses to decide what a child namespace is.

### Gotchas
The namespace href now runs through `encodeURIComponent`. That changes nothing for any name the server permits. It is protection against a malformed payload, not a correction of a fault.

The rail is 236px wide. A filter shows full paths so `finos.traderx` stays distinct from `traderx`, and a match four levels deep will still truncate.

The collapsed strip previously floated as a rounded card inside grey padding, while the expanded rail sits flush. It now matches the expanded rail, so collapsing narrows the rail rather than swapping one shape for another.

### Mobile, and a convention this changes
Mobile does not render this rail. `Hub.tsx` draws a separate drill-down component, `MobileNavMenu`, where each tap replaces the list with the next flat level. Its namespaces level now renders the same tree, with a 40px chevron column and a label as separate targets, each at least 44px tall. A closed row shows the same ghost `+N` pill the desktop rail does. Expansion persists through the same stored set the desktop rail uses.

**One deliberate departure from the issue.** #2787 says the mobile label should navigate to the namespace page and dismiss the drawer. That tap is the only route into the types and resources levels, so following it would leave both unreachable and remove mobile's only path from the drawer to a specific architecture. The label therefore keeps its present behaviour and opens the types list. The chevron carries the new expand action.

**Mobile has no namespace filter, deliberately.** The drawer already carries one text input, `ExplorerSearch`, which queries the hub and replaces the list rather than narrowing it. A second input filtering the list beneath it would put two text fields with different jobs at the top of a 390px drawer. Drilling into a branch is the mobile equivalent of filtering, so the tree is built with filtering switched off there.

That leaves one asymmetry worth knowing. Both surfaces share a single stored set of closed rows, so a branch closed on the desktop rail is closed in the drawer too. On the desktop a filter reveals it regardless, because filtering ignores that set. On mobile the chevron is the only way back. The default is fully open, so this reaches only a user who closed the branch on purpose.

**This changes a documented convention, so it needs a reviewer's eye.** `calm-hub-ui/AGENTS.md` said the mobile explorer is one flat list per level and never a tree. That rule predates #2787 by sixteen days, and #2787 asks for a tree in the component the rule names. The rule is now narrowed rather than dropped: it still governs the types, resources, domains and controls levels, and names the namespaces level as the exception.

## Type of Change
- [ ] 📚 Documentation update

The API does not change. The data model does not change. The hierarchy is a pure function of names the counts endpoint already returns.

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Commit Message Format ✅
`feat(calm-hub-ui): nest namespaces in the Explore rail`
`feat(calm-hub-ui): show namespace roots in the collapsed rail`

## Testing
- [x] All existing tests pass

This PR adds 44 tests. Every existing test passes without a change, including all twelve `ExploreRail.test.tsx` cases and the `Hub.test.tsx` expectations for the collapse button. That was a constraint of the design, not a hoped-for result.

One decision carries the most weight, so I checked it by mutation. If the flatten step reads the stored set of closed rows while a filter is active, exactly three tests fail — one in the pure module, one in the hook, one in the integration suite.

`calm-hub-ui` passes 135 files and 1610 tests. Lint reports 0 errors and 9 warnings, all pre-existing and none in a new file. The build is clean.

Checked by hand against a standalone hub holding `finos`, `finos.calm`, `finos.traderx`, `traderx` and `test-ns`, in both the expanded and the collapsed state.

On viewport coverage: this rail is not drawn on mobile at all, so there is no mobile render of it to test. The suite for the mobile drawer is untouched and passes.

## Checklist


calm-hub-ui/src/hub/components/explore-rail/NamespaceRailItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    return (
+        <div className={`flex items-center gap-1 pr-1.5 py-1 rounded-[7px] text-[13px] ${active ? 'font-semibold' : ''}`} style={rowStyle}>
+            <div className="flex items-center shrink-0" aria-hidden="true">
+                {Array.from({ length: Math.min(depth, MAX_INDENT_GUIDES) }).map((_, i) => (
+                    <span key={i} style={{ display: 'inline-block', width: 14, height: 20, borderLeft: `1px solid ${colors.redesign.border}` }} />
+                ))}
+            </div>
+
+            {hasChildren ? (
```
aamanrebello, line 73: The chevron button renders whenever `hasChildren` is true, without checking `filtering`. In filter mode, `flattenNamespaceTree` always returns `collapsed: false` and ignores the collapsed set, so clicking the chevron here has no visible effect — but it still calls `onToggleCollapsed`, which flips the entry in the persisted collapse set. After the user clears the filter, a namespace can appear collapsed or expanded that they never consciously toggled. Consider disabling the toggle (or hiding the button) while `filtering` is true.

calm-hub-ui/src/hub/components/explore-rail/useNamespaceTree.ts
```diff
@@ -0,0 +1,90 @@
...
+import {
+    ancestorPathsOf,
+    buildNamespaceTree,
+    filterNamespaceTree,
+    flattenNamespaceTree,
+    type NamespaceRow,
+} from './namespace-tree.js';
+
+/** Persist which namespaces are collapsed so a refresh keeps the rail as it was. */
+const COLLAPSED_STORAGE_KEY = 'calmHub.railCollapsedNamespaces';
```
aamanrebello, line 12: `COLLAPSED_STORAGE_KEY` is one global key, but this hook is used by both `ExploreRail` (desktop) and `MobileNavMenu` (mobile), both defaulting `storage` to `localStorage`. Collapsing a namespace on one surface silently changes the starting state on the other, even though they show different node sets. Consider namespacing the key per surface (e.g. a `storageKey` param).

calm-hub-ui/src/hub/components/explore-rail/CollapsedRail.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <div
+            className="relative"
+            onMouseEnter={onOpen}
+            onMouseLeave={onClose}
+            onFocus={onOpen}
+            onBlur={(e) => {
+                if (!isWithin(e.currentTarget, e.relatedTarget)) onClose();
+            }}
+            onKeyDown={(e) => {
+                if (e.key === 'Escape') onClose();
```
aamanrebello, line 74: Escape closes the fly-out unconditionally, which unmounts the `{isOpen && (...)}` block. If a keyboard user has tabbed onto one of the fly-out's `Link` rows, that element disappears and the browser drops focus to `<body>` — nothing here returns focus to the trigger button. Worth keeping a ref to the trigger and calling `.focus()` on it after closing (the same pattern `onBlur` already uses for detecting focus movement).

calm-hub-ui/src/hub/components/explore-rail/CollapsedRail.tsx
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function RootInitial({ root, isActive, isOpen, activeNamespace, onOpen, onClose }: RootInitialProps) {
+    const rows = useMemo(() => flattenNamespaceTree([root], { collapsed: new Set(), filtering: false, visible: new Set() }), [root]);
+
+    return (
+        <div
+            className="relative"
+            onMouseEnter={onOpen}
+            onMouseLeave={onClose}
```
aamanrebello, line 68: `onMouseLeave` closes the fly-out unconditionally, unlike `onBlur` just below it, which checks `isWithin` before closing. A hybrid mouse+keyboard user who opens the fly-out via Tab (`onFocus`) and then nudges the mouse across/off the container will have it closed even though focus is still logically inside. Might be worth mirroring the `isWithin` check here, or dropping `onMouseLeave` in favor of `onBlur`.

calm-hub-ui/src/hub/components/explore-rail/CollapsedRail.tsx
```diff
@@ -0,0 +1,153 @@
...
+                }}
+            >
+                {root.segment.charAt(0).toUpperCase()}
+            </button>
+
+            {isOpen && (
+                <div
+                    className="absolute top-0 left-full ml-1 flex flex-col gap-0.5 p-1.5 rounded-[12px] z-50"
+                    style={{
+                        width: 220,
```
aamanrebello, line 99: This panel has a fixed `width: 220` and no `maxHeight`/`overflowY`, and `rows` is built with `collapsed: new Set()` — so it always renders the full subtree regardless of what the user collapsed on the main rail. For a root with a large or deep subtree this can overflow the viewport with no way to scroll. A capped `maxHeight` plus `overflowY: 'auto'` would be a cheap safety net.

calm-hub-ui/src/hub/components/explore-rail/namespace-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Builds the dot-prefix trie, one row per segment — no path compression.
+ * `platform.payments.ledger` renders as three rows: `platform` and
+ * `platform.payments` are group-only (`total: null`), `ledger` carries the count.
+ */
+export function buildNamespaceTree(namespaceCounts: NamespaceCounts[]): NamespaceTreeNode[] {
+    const root: TrieNode = { total: null, children: new Map() };
+    for (const nc of namespaceCounts) {
+        const segments = (nc.namespace ?? '').split('.').filter(Boolean);
```
aamanrebello, line 37: `.filter(Boolean)` drops empty path segments, so e.g. `"billing..invoices"` normalizes to the same trie path as a legitimate `"billing.invoices"`. If both exist, whichever is processed second silently overwrites `current.total` for the first, and the rendered row links via the normalized path, not the original string. Not sure if malformed namespace strings like this can actually reach this component — if they can't (e.g. validated upstream), feel free to disregard.

calm-hub-ui/src/hub/components/explore-rail/ExploreRail.tsx
```diff
@@ -115,18 +117,21 @@ export function ExploreRail({
                         <RailSpinner label="Loading namespaces" />
                     ) : namespacesFailed ? (
                         <RailEmpty>Couldn&apos;t load namespaces</RailEmpty>
-                    ) : filteredNamespaces.length === 0 ? (
-                        <RailEmpty>
-                            {namespaceCounts.length === 0 ? 'Nothing here' : 'No namespaces match your filter'}
-                        </RailEmpty>
+                    ) : rows.length === 0 ? (
+                        <RailEmpty>{needle === '' ? 'Nothing here' : 'No namespaces match your filter'}</RailEmpty>
```
aamanrebello, line 121: This used to check `namespaceCounts.length === 0` for the "nothing fetched" message; now it checks `needle === ''`. These aren't equivalent — if `namespaceCounts` is non-empty but `buildNamespaceTree`/`flattenNamespaceTree` ends up producing zero rows for some other reason (e.g. the malformed-namespace case noted on `namespace-tree.ts:37`), this now reports "Nothing here" as if the fetch failed, even though it didn't. Might be worth deriving this from whether the tree actually produced rows rather than from the filter text.

calm-hub-ui/src/hub/components/explore-rail/NestedCountBadge.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * everywhere it is asserted.
+ */
+export function NestedCountBadge({ count }: NestedCountBadgeProps) {
+    return (
+        <span
+            data-testid="nested-count-badge"
+            className="font-mono-jb text-[11px] leading-none px-1.5 py-0.5 rounded-md min-w-[20px] text-center inline-block"
+            style={{ backgroundColor: colors.redesign.badgeBgFaint, color: colors.redesign.disabled }}
+        >
+            +{count}
```
aamanrebello, line 21: No guard for `count === 0` — if a collapsed namespace's descendants are all zero-total, this renders a "+0" pill, implying hidden content that isn't there. A `descendantTotal > 0` check at the call site (or a guard here) would avoid it.

calm-hub-ui/src/hub/components/explore-rail/namespace-tree.ts
```diff
@@ -0,0 +1,154 @@
...
+ * One row of the dot-prefix namespace tree. `total` is `null` for a synthetic
+ * grouping-only node — a path segment that is not itself a namespace, only an
+ * ancestor of one (e.g. `platform` when only `platform.payments.ledger` exists).
+ */
+export interface NamespaceTreeNode {
+    /** Full dot-separated path, e.g. `finos.calm`. */
+    path: string;
+    /** Last path segment — what a row shows outside filter mode. */
+    segment: string;
+    total: number | null;
```
aamanrebello, line 13: `node.total !== null` (the "group-only vs. real namespace" check) is independently re-derived in `NamespaceRailItem.tsx:55`, `CollapsedRail.tsx:20`, and `MobileNamespaceRow.tsx:25`. Not a bug today, but if this sentinel ever changes, a fix applied to one and missed in another would silently make a namespace clickable in some views and not others. Might be worth an `isNamespace(node)` helper exported from here — could easily be a follow-up rather than part of this PR.

calm-hub-ui/src/hub/components/tree-navigation/MobileNamespaceRow.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * (collapse/expand in place) and the label (drill into the namespace's
+ * types) are separate 44px-tall tap targets, split by a hairline divider —
+ * a `<button>` inside a `<button>` is invalid, and the two do different things.
+ */
+export function MobileNamespaceRow({ row, active, onToggleCollapsed, onOpen }: MobileNamespaceRowProps) {
+    const { node, hasChildren, collapsed, depth, descendantTotal } = row;
+    const isNamespace = node.total !== null;
+
+    return (
+        <div className="flex items-stretch" style={{ paddingLeft: depth * INDENT_PER_DEPTH }}>
```
aamanrebello, line 28: Indentation-per-depth is implemented three different, uncoordinated ways: `NamespaceRailItem.tsx` caps at `MAX_INDENT_GUIDES = 4` (14px each), `CollapsedRail.tsx`'s `FlyoutRow` uses `8 + depth * 14` uncapped, and this uses `depth * 16` uncapped. A namespace nested 10 levels deep would push 160px of padding on a mobile row, with no cap like the desktop rail has. Depends on how deep namespaces realistically nest in practice — if it's rare this might not matter, but a shared cap (mirroring the desktop pattern) would be cheap insurance.

aamanrebello (commented):
This review was run with AI-assisted tooling (Claude Code) against this branch's diff. Sorry for the comment volume from an AI pass — please feel free to respond on any comment you think does not apply, or where the tradeoff was intentional.

calm-hub-ui/src/hub/components/explore-rail/NamespaceRailItem.tsx
YoofiTT96: Resolved

calm-hub-ui/src/hub/components/explore-rail/useNamespaceTree.ts
YoofiTT96: Both surfaces render the same tree from the same counts, so shared collapse state is better for consistency

calm-hub-ui/src/hub/components/explore-rail/CollapsedRail.tsx
YoofiTT96: Resolved

calm-hub-ui/src/hub/components/explore-rail/CollapsedRail.tsx
YoofiTT96: Fair. Resolved

calm-hub-ui/src/hub/components/tree-navigation/MobileNamespaceRow.tsx
YoofiTT96: Will add a shared implementation with padding and overflow cap at depth > 5

Issue #3023, opened 2026-08-26
Add snapshots feature to CalmHub and tooling to support scratch work


willosborne:
### Target Project:
CalmHub, CLI

### Description of Feature:
When working on pattterns, architectures etc, to be published to CalmHub,  it's useful to be able to publish them to visualise + work with them in the UI.
However, these patterns usually need multiple iterations as experimental work. 

This issue proposes the introduction of a 'snapshots' feature, similar to Maven's appraoch to this. 

When publishing a document, you can add `-SNAPSHOT` to the end of its version to mark it as a SNAPSHOT. Only one snapshot can be associated with a given semantic version.

Snapshots are **mutable** - pushing again to the snapshot will **overwrite it** in the database.
Then when work is finished, the user can publish the full version without the -SNAPSHOT suffix; this will delete the snapshot.

### User Stories:
- As an architect, I want a snapshot of a document, so I can quickly iterate on patterns and architectures.

### Current Limitations:
- Currently there are only immutable versions.
- There is no concept of promotion between namespaces or CalmHub instances.

### Proposed Implementation:
When creating a resource via the `POST` or `PUT /calm/**` user-facing endpoints, put `-SNAPSHOT` into the version. (Note: POST creates, you can only update a snapshot with PUT.)

For example:

```bash
POST /calm/namespaces/workshop/architectures/test/versions/1.0.0-SNAPSHOT
201 Created

# then to update:
PUT /calm/namespaces/workshop/architectures/test/versions/1.0.0-SNAPSHOT
204 No Content

# … many changes later, create the final version without the suffix:
POST /calm/namespaces/workshop/architectures/test/versions/1.0.0
201 Created

# this DELETES the snapshot:
GET /calm/namespaces/workshop/architectures/test/versions/1.0.0-SNAPSHOT
404 Not Found
```

#### UI changes

Snapshots will appear in the UI stamped with a snapshot label. They will appear on timelines too, so you can see the diff to the current snapshot.

#### Tooling support 

This would then be added to the `hub` commands via a `--snapshot` flag:

`calm hub push architecture --namespace abcd --mapping test --version 1.0.1 --snapshot`

...and the workspace commands by adding a `--snapshot` flag to `bump` and `push`.

When working on a document, if you want to move to a snapshot version, run `calm workspace bump --snapshot`.
This will bump any modified files as usual, except the NEW version will be the snapshot. 
No files will be pushed.

Then to restore something from a snapshot i.e. to publish the current state:

[...]

PR #3122, opened 2026-09-19
feat(calm-hub): add snapshot versions for iterating before publish

willosborne:
## Description

Adds `-SNAPSHOT` versions to CalmHub, so architects can iterate on a pattern or architecture before they publish it. Closes part of #3023.

A snapshot is a mutable version. Push to it again and it is overwritten. Publish the release version and CalmHub deletes the snapshot.

| Request | Result |
|---|---|
| `POST .../versions/1.0.0-SNAPSHOT` | `201`, or `200` if it is already there |
| `POST .../versions/1.0.0-SNAPSHOT` when `1.0.0` is published | `409` |
| `POST .../versions/1.0.0` | `201`, and the snapshot is deleted |

Scope is **CalmHub only**, and only the five namespace resource types: pattern, architecture, flow, standard and interface. The CLI and UI parts of #3023 need their own issues. The `-SNAPSHOT` suffix is refused on ADR, timeline, layout and domain controls.

The issue also asks to remove auto-bump from `calm workspace push`. That is already the behaviour, so no change was needed.

### Decisions worth review

**Snapshots work on the name-based `/calm/...` API only.** The numeric `/api/calm/...` endpoints call the stores directly and never reach `MappingControllerService`, where the three snapshot rules live. Sharing a store does not share the rules. A numeric endpoint that took a snapshot would create one that shadows a published release, with no check and no cleanup, so those endpoints keep the strict version pattern on `POST` and `PUT`. Their `GET` accepts the suffix, so a snapshot stays readable there.

**Promotion is not atomic, deliberately.** The release is written first, then the snapshot is deleted. There is no transaction across the two. If the delete fails, the release is correct, the request still returns `201`, and the orphan is logged. The reverse order would let a failed release write destroy work in progress.

**The permission model is unchanged.** `WRITE` covers snapshots. No new scope.

### Gotchas that are not obvious

- **Six spellings, one version.** `VERSION_REGEX` makes both separators optional, so `1.0.0`, `1-0-0`, `1.00` and `100` are the same version. Any version that comes from a request and is compared against stored data must be canonicalised first. This was got wrong four times while building this. `docs/superpowers/specs` records it; `ResourceVersion` and `CanonicalVersion` are the only places that should know about the suffix.
- **`Semver.tryParse` order of operations.** It replaces `-` with `.`, which is what makes `1-0-0` parse. The suffix must come off *before* that. After it, `1.0.0-SNAPSHOT` becomes four segments and collapses to `0.0.0`, which sorts lowest — the ADR revision 100 failure that `VersionScheme` documents.

### Known limitations

- **One audit row per request.** `AuditRequestFilter` stages a single context, so promotion cannot record both the release creation and the snapshot deletion. The release keeps its row; the deletion is not audited. Recording both needs multi-row support.
- **Search has no version dimension.** `SearchResult` carries no version, and search matches the resource header, whose name and description come from the most recent write. A snapshot's title therefore reaches search results before the work is published.
- **MCP tools keep the strict pattern** and cannot read a snapshot version.

### Separate from this PR

While working on `MappingControllerService` we found a pre-existing bug, which this PR does **not** change. At line 114 the `PUT` caller passes its arguments in the wrong order for `updateVersionedResourceInStore`, so a `PUT` writes the title as the document body. It comes from `1c95ec5a` and the existing tests miss it because they use `any()` matchers. It only affects instances that set `allow.put.operations=true`. It needs its own issue.

## Type of Change
- [x] 📚 Documentation update

## Affected Components
- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [x] CALM Hub (`calm-hub/`)
- [x] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing
- [x] All existing tests pass

`./mvnw test -pl calm-hub` — 3108 pass, 0 failures. Up from 3090 on `main`.
`../mvnw -P integration verify` — 551 pass, including a new `SnapshotLifecycleIntegration` that runs create, overwrite, publish, and the shadow refusal against MongoDB in TestContainers.

Tests were checked by mutation: production code was deliberately broken to confirm each new test fails. This found several tests that passed against broken code and would not have caught a regression.

## Checklist


willosborne:
Can't make a snapshot if it already exists: 
image
Can make a snapshot: 
image

Appears in UI: 
image

Can edit the document on the snapshot state:
image

image

NB diff works and updates as you edit!

Publishing the real version deletes the snapshot:
image

image

calm-hub/src/main/java/org/finos/calm/store/util/SemanticVersionOrder.java
```diff
@@ -46,6 +47,12 @@ private static int compare(String left, String right) {
         if (comparison != 0) {
             return comparison;
         }
+        // A snapshot precedes the release it belongs to, per semver pre-release ordering.
+        boolean leftSnapshot = ResourceVersion.isSnapshot(leftVersion);
+        boolean rightSnapshot = ResourceVersion.isSnapshot(rightVersion);
+        if (leftSnapshot != rightSnapshot) {
+            return leftSnapshot ? -1 : 1;
```
rocketstack-matt, line 54: With this ordering a `1.1.0-SNAPSHOT` ranks above `1.0.0`, so `FlowResource.getLatestFlow` (`versions.getLast()`, line 126) now serves unpublished work to every READ consumer as "latest" once a snapshot exists past the newest release. Same effect on the implied timeline's `current-moment` in `ArchitectureTimelineService`. Maven distinguishes `LATEST` (includes snapshots) from `RELEASE` here. Is that intended? If not, resolve "latest" to the highest non-snapshot version and fall back to the snapshot only when no release exists.

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
```diff
@@ -421,14 +425,17 @@ public int resolveConfigId(String domain, int controlId, String configName)
...
      */
     private Response createNewResource(String namespace, ResourceType resourceType, String typePath,
                                        String name, String json, CalmDocumentParser.VersionSpec versionSpec) throws URISyntaxException {
         String finalVersion = versionSpec.version() != null ? versionSpec.version() : "1.0.0";
-        if (!"1.0.0".equals(finalVersion)) {
+        // The request's release spelling must be canonicalised before comparison — releaseVersion
+        // alone leaves "100-SNAPSHOT" as "100", which would never equal "1.0.0".
+        if (!"1.0.0".equals(CanonicalVersion.of(ResourceVersion.releaseVersion(finalVersion)))) {
             return Response.status(Response.Status.BAD_REQUEST)
                     .entity("The first version of a resource must be 1.0.0, but " + finalVersion + " was requested")
```
rocketstack-matt, line 440: Message is stale now that `1.0.0-SNAPSHOT` is accepted here. (non-blocking nit)
```suggestion
                    .entity("The first version of a resource must be 1.0.0 or 1.0.0-SNAPSHOT, but " + finalVersion + " was requested")
```

calm-hub/src/main/java/org/finos/calm/resources/MappingControllerResource.java
```diff
@@ -96,7 +96,10 @@ public MappingControllerResource(MappingControllerService service,
                     "For control requirements: {baseUrl}/calm/domains/{domain}/controls/{controlName}/requirement/versions/{version}. " +
                     "For configurations: {baseUrl}/calm/domains/{domain}/controls/{controlName}/configurations/{configName}/versions/{version}. " +
                     "A version is always required. For a brand-new resource the first version must be 1.0.0. " +
```
rocketstack-matt, line 98: Swagger text here and at line 246 still says a brand-new resource must start at 1.0.0; both should mention 1.0.0-SNAPSHOT, since that's the main flow this PR enables. (non-blocking nit)

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
```diff
@@ -14,9 +15,12 @@
 import org.finos.calm.domain.interfaces.CreateInterfaceRequest;
 import org.finos.calm.domain.pattern.CreatePatternRequest;
 import org.finos.calm.domain.standards.CreateStandardRequest;
+import org.finos.calm.domain.ResourceVersion;
```
rocketstack-matt, line 18: Already covered by the wildcard import on line 7. (non-blocking nit)
```suggestion
```

docs/docs/calm-hub/mcp-and-api.md
```diff
@@ -57,6 +57,18 @@ GET  /calm/namespaces/{namespace}/architectures
...
+
+A version ending in `-SNAPSHOT` (e.g. `1.0.0-SNAPSHOT`) is mutable — it can be re-posted with new content. A release version (e.g. `1.0.0`) is immutable once created. Publishing a release deletes the matching snapshot, if one exists. Snapshots are supported on architectures, patterns, flows, standards, and interfaces only.
+
+| Request | Effect | Status |
+|:--------|:-------|:-------|
+| `POST .../versions/1.0.0-SNAPSHOT` (first time) | Creates the snapshot | `201` |
+| `POST .../versions/1.0.0-SNAPSHOT` (again) | Replaces the snapshot's content | `200` |
+| `POST .../versions/1.0.0` | Publishes the release, deletes `1.0.0-SNAPSHOT` | `201` |
+
+A snapshot cannot be created if its release version is already published — CALM Hub returns `409 Conflict`.
```
rocketstack-matt, line 70: Worth stating the numeric-API boundary here, since it's the first thing someone scripting against `/api/calm` will hit. (non-blocking nit)
```suggestion
A snapshot cannot be created if its release version is already published — CALM Hub returns `409 Conflict`.

Snapshots are written through the name-based `/calm/...` API only. The numeric `/api/calm/...` write endpoints reject a `-SNAPSHOT` version; their `GET` endpoints can read one.
```

calm-hub/src/main/java/org/finos/calm/security/AuditRequestFilter.java
```diff
@@ -149,6 +149,21 @@ public static void stage(AuditContext context) {
         STAGED_CONTEXT.set(context);
     }
 
+    /**
+     * Replaces just the {@code action} on the currently staged {@link AuditContext},
+     * leaving every other field as-is. Does nothing when no context is staged. For
+     * callers (e.g. {@code MappingControllerService#addNewVersion}) that only learn the
+     * true action after the resource layer has already staged a provisional one.
+     */
+    public static void restageAction(AuditAction action) {
```
rocketstack-matt, line 158: This is the first service-layer caller into the audit filter's ThreadLocal; until now only resources staged audit context. Not blocking, but an alternative that keeps the layering is for `addNewVersion` to return the action alongside the response, or for the filter to map a 200 on this route to UPDATE. (non-blocking nit)

rocketstack-matt (commented):
One design question on what "latest" should resolve to once snapshots exist, plus nits. Also: the description cites `docs/superpowers/specs` for the six-spellings gotcha, but this PR adds no spec there. Either add it or drop the reference.

willosborne:
CLI-side workspace support for these snapshot versions is stacked on top of this branch in #3132 (`feat/cli-workspace-snapshots`, based on `feat/calm-hub-snapshots`). Once this merges, #3132 will need retargeting to `main`.

calm-hub/src/main/java/org/finos/calm/store/util/SemanticVersionOrder.java
willosborne: Fixed in b76e0a78 — added `SemanticVersionOrder.latestRelease`, which resolves to the highest release and falls back to the highest snapshot only when nothing is published yet. `FlowResource.getLatestFlow` and `ArchitectureTimelineService`'s implied `current-moment` both now use it.

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
willosborne: Fixed in b76e0a78.

calm-hub/src/main/java/org/finos/calm/resources/MappingControllerResource.java
willosborne: Fixed in b76e0a78 (both call sites).

calm-hub/src/main/java/org/finos/calm/services/MappingControllerService.java
willosborne: Fixed in b76e0a78.

docs/docs/calm-hub/mcp-and-api.md
willosborne: Fixed in b76e0a78, using your suggested text.

calm-hub/src/main/java/org/finos/calm/security/AuditRequestFilter.java
willosborne: Deleted claude's auto-reply to this as I don't entirely understand the issue here. @markscott-ms would be interested in your thoughts here.

This is basically a conflict of what is responsible for updating the audit log, right? Previously it was always happening earlier in the flow (REST) but here we have to do it in the service because of the ambiguity about patching snapshot versions.

I'm not sure how to get around this without having a special rule just for snapshots.

willosborne: Claude's justification for not fixing this: 

> The two alternatives Matt suggested (and why I didn't take either)
> 
> 1. Have addNewVersion return the action alongside its Response, and let the resource method stage the correction — keeps the ThreadLocal resource-only, but means changing addNewVersion's return type and every caller.
> 2. Have the filter itself map "this route returned 200 instead of 201" to UPDATE, based on status code alone — removes restageAction entirely, but means teaching the filter route-specific status-code logic it doesn't currently have, in code that's explicitly documented as fragile (the whole ThreadLocal + same-thread-dispatch design is a deliberate, carefully-commented workaround, not something to casually extend).
> 
> Both are real refactors of the audit-trail's core resolution logic, for a benefit that's purely aesthetic (better layering, not a correctness fix). Given audit logging has no test-suite-independent verification in production (per calm-hub's own docs, there's no API to read it back — you'd only notice a logging bug by inspecting the database directly), I judged it not worth the risk for a non-blocking nit, and said so in the reply rather than guessing which of the two Matt would actually prefer.

Not sure whether this is a real structural issue or just something cosmetic that we can ignore

PR #3125, opened 2026-09-20
perf(calm-hub-ui): memoize visualiser node/edge components and cut hover re-allocations

aamanrebello:
## Description

**Human Comment**: Just asked Claude to look at Calm Hub UI and identify opportunities to optimize latency. Also asked it to investigate risks - this optimization seemed very low risk (although you will probably only see any real latency difference on large complicated graphs). In essence, this PR prevents unnecessary re-renders when hovering over nodes or edges in the ReactFlow graph.

**Coding Assistant Comment**: Wraps the visualiser's `CustomNode`, `SystemGroupNode`, and `FloatingEdge` components in `React.memo`, and changes the node hover handlers (`useGraphInteractions.ts`) to only allocate a new node object when its `zIndex` value actually changes.

Before this change, hovering any node caused every node component in the diagram to re-render, because the hover handlers rebuilt the whole node array on every `mouseenter`/`mouseleave` — giving every node a new object reference regardless of whether its z-index actually changed. `React.memo` alone cannot fix this; it needs stable prop references to skip a re-render. This change fixes both together: the hover handlers now only reallocate the node(s) whose z-index truly changed, so the memoized components correctly skip re-rendering everything else.

No behavior change — only render frequency. Hover, click, and drag interactions render identically; diagram output is unchanged.

## Type of Change

- [ ] 📚 Documentation update
- [x] ⚡ Performance improvements

## Affected Components

- [ ] CLI (`cli/`)
- [ ] Schema (`calm/`)
- [ ] CALM AI (`calm-ai/`)
- [ ] CALM Hub (`calm-hub/`)
- [ ] Documentation (`docs/`)
- [ ] Shared (`shared/`)
- [ ] Dependencies
- [ ] CI/CD

## Testing

- [x] All existing tests pass

### Verifying the benefit
Measured actual component re-renders per hover event with a temporary instrumented counter (not part of this diff), comparing `main` against this branch on the same diagram (`finos.fluxnova` / `fluxnova-microservices`: 11 regular nodes + 1 group node):
| Branch | Component re-renders on a single hover |
|---|---|
| `main` (before) | all 11 `CustomNode`s + the group node |
| this branch (after) | 0 |

### Checking for regressions
- Full unit suite: 1562 tests / 130 files pass, no regressions.
- Lint: 0 errors (pre-existing, unrelated warnings only).
- Manual verification against a real seeded backend, on a diagram with a group node and several children:
  - hover shows/elevates/clears correctly on both regular and grouped nodes, and the hover panel escapes the group boundary without clipping
  - dragging, container reflow, and layout persistence are unaffected
  - click-through to the node detail panel and control-jump links still work (confirms `data` callbacks survive memoization)
  - rapid hovering across adjacent nodes leaves no node stuck at an elevated z-index

Two pre-existing, unrelated issues turned up during manual testing and are being filed as separate tickets rather than fixed here, to keep this PR scoped to the performance change:
- The hover popover can be clipped by the diagram pane's overflow boundary near the bottom "Controls" panel — pure CSS containment, reproducible on `main` too.
- On mobile/touch, the hover popover and the click-triggered detail panel race each other, since the popover only closes via an incidental `mouseleave` rather than closing deterministically on click. The race itself exists on `main` too; this change only alters its *visible* symptom, because faster re-rendering changes the timing (main visibly flashes on every tap; this branch only on the first).

## Checklist


calm-hub-ui/src/visualizer/components/reactflow/hooks/useGraphInteractions.ts
```diff
@@ -88,29 +88,28 @@ export function useGraphInteractions({
...
-                    },
-                }))
+                nds.map((n) => {
+                    const zIndex = n.id === node.id && !isGroupType(n.type) ? 1000
+                        : isGroupType(n.type) ? -1
+                        : 1;
+                    // Only allocate a new node object when the value actually
+                    // changes, so memoized node components (CustomNode,
+                    // SystemGroupNode) don't re-render for untouched nodes.
+                    if (n.style?.zIndex === zIndex) return n;
```
rocketstack-matt, line 98: The reference-stability of untouched nodes is the whole optimisation, but nothing pins it: the hook tests don't cover the hover handlers, and the checklist says tests were added but the diff has none. This block in `useGraphInteractions.test.ts` fails on `main` and passes on this branch:

```ts
describe('useGraphInteractions hover z-index', () => {
    const hoverNodes: Node[] = [
        { id: 'a', type: 'custom', position: { x: 0, y: 0 }, data: {}, style: { zIndex: 1 } },
        { id: 'b', type: 'custom', position: { x: 0, y: 0 }, data: {}, style: { zIndex: 1 } },
        { id: 'g', type: 'group', position: { x: 0, y: 0 }, data: {}, style: { zIndex: -1 } },
    ];

    function setupHover() {
        let updated: Node[] = [];
        const setNodes = vi.fn((updater: (nodes: Node[]) => Node[]) => {
            updated = updater(hoverNodes);
        });
        const { result } = renderHook(() =>
            useGraphInteractions({ setNodes, onNodesChangeBase: vi.fn(), groupNodeTypes: ['group'] })
        );
        return { result, getUpdated: () => updated };
    }

    it('elevates only the hovered node and keeps other node references stable', () => {
        const { result, getUpdated } = setupHover();
        result.current.handleNodeMouseEnter({} as React.MouseEvent, hoverNodes[0]);
        const [a, b, g] = getUpdated();
        expect(a.style?.zIndex).toBe(1000);
        expect(a).not.toBe(hoverNodes[0]);
        expect(b).toBe(hoverNodes[1]);
        expect(g).toBe(hoverNodes[2]);
    });

    it('resets only the elevated node on mouse leave', () => {
        const { result, getUpdated } = setupHover();
        result.current.handleNodeMouseEnter({} as React.MouseEvent, hoverNodes[0]);
        const elevated = getUpdated();
        const setNodes = vi.fn((updater: (nodes: Node[]) => Node[]) => updater(elevated));
        const { result: r2 } = renderHook(() =>
            useGraphInteractions({ setNodes, onNodesChangeBase: vi.fn(), groupNodeTypes: ['group'] })
        );
        r2.current.handleNodeMouseLeave();
        const [a, b, g] = setNodes.mock.results[0].value as Node[];
        expect(a.style?.zIndex).toBe(1);
        expect(b).toBe(elevated[1]);
        expect(g).toBe(elevated[2]);
    });
});
```

calm-hub-ui/src/visualizer/components/reactflow/hooks/useGraphInteractions.ts
```diff
@@ -88,29 +88,28 @@ export function useGraphInteractions({
...
-                            : 1,
-                    },
-                }))
+                nds.map((n) => {
+                    const zIndex = n.id === node.id && !isGroupType(n.type) ? 1000
+                        : isGroupType(n.type) ? -1
+                        : 1;
+                    // Only allocate a new node object when the value actually
+                    // changes, so memoized node components (CustomNode,
+                    // SystemGroupNode) don't re-render for untouched nodes.
```
rocketstack-matt, line 97: (non-blocking nit) The mechanism is one layer up: ReactFlow's own `memo(NodeWrapper)` compares `style` by reference, so the identity return is what stops every node re-rendering, and the component `memo()`s are a second layer.
```suggestion
                    // Return the untouched node as-is: ReactFlow's memoised
                    // NodeWrapper compares `style` by reference, so a fresh
                    // object here re-renders every node on each hover.
```

calm-hub-ui/src/visualizer/components/reactflow/SystemGroupNode.tsx
```diff
@@ -37,4 +38,6 @@ export function SystemGroupNode({ data }: NodeProps) {
       </div>
     </div>
   );
-};
+}
+
+export const SystemGroupNode = memo(SystemGroupNodeComponent);
```
rocketstack-matt, line 43: (non-blocking nit) `DecisionGroupNode` is registered alongside these in `PatternGraph.tsx` and `DiffGraph.tsx` but isn't wrapped; worth memoising it too for consistency.

rocketstack-matt (changes requested):
The optimisation is right; the hover reference-stability needs a test before this goes in.

